PR #6, opened 2023-03-16
Add tokenizer utils

nbhargava:
Adds basic utils for fetching the tokenizer used by the Anthropic API and computing the total number of tokens in a string.

### DIFF
--- a/anthropic/tokenizer.py
+++ b/anthropic/tokenizer.py
@@ -1,0 +1,20 @@
+import requests
+from tokenizers import Tokenizer
+
+CLAUDE_TOKENIZER_REMOTE_FILE = "https://public-json-tokenization-0d8763e8-0d7e-441b-a1e2-1c73b8e79dc3.storage.googleapis.com/claude-v1-tokenization.json"
+
+claude_tokenizer = None
+
+def get_tokenizer() -> Tokenizer:
+    global claude_tokenizer
+
+    if not claude_tokenizer:
+        tokenizer_data = requests.get(CLAUDE_TOKENIZER_REMOTE_FILE)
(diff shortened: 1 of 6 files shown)


anthropic/tokenizer.py
```diff
@@ -0,0 +1,20 @@
...
+
+CLAUDE_TOKENIZER_REMOTE_FILE = "https://public-json-tokenization-0d8763e8-0d7e-441b-a1e2-1c73b8e79dc3.storage.googleapis.com/claude-v1-tokenization.json"
+
+claude_tokenizer = None
+
+def get_tokenizer() -> Tokenizer:
+    global claude_tokenizer
+
+    if not claude_tokenizer:
+        tokenizer_data = requests.get(CLAUDE_TOKENIZER_REMOTE_FILE)
```
mikelambert: Can you keep the cache code too? I was worried that every incarnation of someone testing a local dev script needs to download 1.6MB out of the box. Also, if this is code we want stainless to incorporate, I think it's better to start with httpx here.

mikelambert (dismissed):
Thanks!

anthropic/tokenizer.py
nbhargava: Are we okay adding `httpx` as a dependency? (Just realizing that I forgot to explicitly add `tokenizers` as a dependency in this PR.)

mikelambert: Yeah, I'm not worried about the deps. We'll be switching over to httpx eventually. I feel like these deps will mostly end up rounding errors in downstream usage of our code.

mikelambert (approved):

PR #9, opened 2023-03-22
Add some length checks on the prompts to ensure they're within range

mikelambert:


mikelambert:
Going to submit this one, want to get a numeric release bump out that fixes this gap

Issue #13, opened 2023-03-30
Exception: EOF while parsing a value at line 1 column 0


sergeyf:
Hello, 

I'm on Windows and installed with `pip install git+https://github.com/anthropics/anthropic-sdk-python.git`

Seeing this issue when issuing a standard completion prompt. 

This is the command:
```python
import anthropic

anthropic_client = anthropic.Client(api_key=...)

prompt = '\n\nHuman: This is a chat between me and you\n\nAssistant:'

resp = anthropic_client.completion(
    prompt=prompt,
    stop_sequences=[anthropic.HUMAN_PROMPT],
    model="claude-instant-v1",
    max_tokens_to_sample=200,
)
```

Here is the complete error:
```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
Cell In[35], line 1
----> 1 resp = anthropic_client.completion(
      2     prompt=prompt,
      3     stop_sequences=[anthropic.HUMAN_PROMPT],
      4     model=model,
      5     max_tokens_to_sample=max_tokens,
      6 )

File ~\.conda\envs\guideml\lib\site-packages\anthropic\api.py:237, in Client.completion(self, **kwargs)
    236 def completion(self, **kwargs) -> dict:
--> 237     return self._request_as_json(
    238         "post",
    239         "/v1/complete",
    240         params=kwargs,
    241     )

File ~\.conda\envs\guideml\lib\site-packages\anthropic\api.py:196, in Client._request_as_json(self, *args, **kwargs)
    195 def _request_as_json(self, *args, **kwargs) -> dict:
--> 196     result = self._request_raw(*args, **kwargs)
    197     content = result.content.decode("utf-8")
    198     json_body = json.loads(content)

File ~\.conda\envs\guideml\lib\site-packages\anthropic\api.py:117, in Client._request_raw(self, method, path, params, headers, request_timeout)
    109 def _request_raw(
    110     self,
    111     method: str,
   (...)
    115     request_timeout: Optional[Union[float, Tuple[float, float]]] = None,
    116 ) -> requests.Response:
--> 117     request = self._request_params(headers, method, params, path, request_timeout)
    118     result = self._session.request(
    119         request.method,
    120         request.url,
   (...)
    124         timeout=request.timeout,
    125     )
    127     _process_request_error(
    128         method, result.content.decode("utf-8"), result.status_code
    129     )

[...]

Issue #14, opened 2023-04-01
Problem with the billing account that is used to host the tokenizer.


mobarski:
The simplest possible use of count_tokens ends with an error
```python
import anthropic
anthropic.count_tokens('hello')
```

The error:
```
raceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/mobarski/.local/lib/python3.10/site-packages/anthropic/tokenizer.py", line 34, in count_tokens
    tokenizer = get_tokenizer()
  File "/home/mobarski/.local/lib/python3.10/site-packages/anthropic/tokenizer.py", line 29, in get_tokenizer
    claude_tokenizer = Tokenizer.from_str(tokenizer_data)
Exception: expected value at line 1 column 1
```

A quick inspection in `tokenizer.py` shows the URL from where the tokenizer should be downloaded:
```
CLAUDE_TOKENIZER_REMOTE_FILE = "https://public-json-tokenization-0d8763e8-0d7e-441b-a1e2-1c73b8e79dc3.storage.googleapis.com/claude-v1-tokenization.json"
```

And trying to access this link in the browser gives this:
```html
<Error>
  <Code>UserProjectAccountProblem</Code>
  <Message>
    The project to be billed is associated with a closed billing account.
  </Message>
  <Details>
    The billing account for the owning project is disabled in state closed
  </Details>
</Error>
```

mobarski:
The impact of this problem seems to be larger - completion endpoint is also affected!

Workaround:
`pip3 install anthropic==0.2.3`

mikelambert:
Thank you for reporting, and sorry about this bug with the release that went out a few days ago. The bucket should be fixed as of earlier this afternoon (within a couple hours of this being posted). And https://github.com/anthropics/anthropic-sdk-python/pull/16 should make this additional check fail-open, so it shouldn't recur even if something happens to the bucket.

PR #16, opened 2023-04-02
Fail open when the tokenizer fails to load 

mikelambert:
(ie, due to some AWS bucket issue). Fixes https://github.com/anthropics/anthropic-sdk-python/issues/13 and https://github.com/anthropics/anthropic-sdk-python/issues/14 in a better way

### DIFF
--- a/anthropic/api.py
+++ b/anthropic/api.py
@@ -273,10 +273,13 @@ def _validate_request(params: dict) -> None:
 
 def _validate_prompt_length(params: dict) -> None:
     prompt: str = params["prompt"]
-    prompt_tokens = tokenizer.count_tokens(prompt)
-    max_tokens_to_sample: int = params["max_tokens_to_sample"]
-    token_limit = 9 * 1024
-    if prompt_tokens + max_tokens_to_sample > token_limit:
-        raise ApiException(
-            f"Prompt tokens ({prompt_tokens}) + max-sampled tokens ({max_tokens_to_sample}) exceeds max ({token_limit})",
-        )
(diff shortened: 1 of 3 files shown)


anthropic/api.py
```diff
@@ -273,10 +273,13 @@ def _validate_request(params: dict) -> None:
...
+    try:
+        prompt_tokens = tokenizer.count_tokens(prompt)
+        max_tokens_to_sample: int = params["max_tokens_to_sample"]
+        token_limit = 9 * 1024
+        if prompt_tokens + max_tokens_to_sample > token_limit:
+            raise ApiException(
+                f"Prompt tokens ({prompt_tokens}) + max-sampled tokens ({max_tokens_to_sample}) exceeds max ({token_limit})",
+            )
+    except tokenizer.TokenizerException:
+        pass
```
nbhargava: Does it make sense to log a warning or something here, just so that downstream users are aware that something has gone wrong and/or we get a signal?

nbhargava (dismissed):
Want to also bump the version in this PR so we can just go ahead and deploy?

nbhargava:
cc @jenan-anthropic for visibility

anthropic/api.py
mikelambert: good idea, adding

mikelambert:
I also had https://github.com/anthropics/anthropic-sdk-python/pull/17 out and wasn't sure which one was going in first (or whether I'd get both in together), so wasn't sure about bumping version numbers yet.

PR #17, opened 2023-04-02
Better async support

mikelambert:


agola11 (approved):
This lgtm but curious what moving the status check outside of `_process_request_error` fixes?

mikelambert:
Moving the check outside let's us avoid `_arequest_as_stream` waiting on `await result.text()` before it starts streaming, [as in here](https://github.com/anthropics/anthropic-sdk-python/pull/17/files#diff-d9e91b5c7eefb5cfd90688f6926e99976871e9563bf9dcddd3c819eb4aca920fR174).

agola11 (approved):

agola11 (commented):
Ah yeah that makes a lot of sense, was a mistake on my part. thanks for fixing!

PR #27, opened 2023-04-25
fix: api_url concat

### LINKED ISSUE #26: Fix URL concatenation in `anthropic/api.py`
I noticed a small issue with the URL concatenation logic in
https://github.com/anthropics/anthropic-sdk-python/blob/4187c65ae9713b7579fcb15aec43f78ed69b97c4/anthropic/api.py#L71. 
The current code uses urllib.parse.urljoin(self.api_url, path) to concatenate URLs, but this results in incorrect URLs when concatenating URLs like 
`http://localhost:5000/anthropic` and `/v1/complete`, giving 
`http://localhost:5000/v1/complete` instead of the expected 
`http://localhost:5000/anthropic/v1/complete`.
A possible fix could be to directly concatenate the strings, like:
`abs_url = "%s%s" % (self.api_base, url)`
as implemented in the OpenAI API requestor:
https://github.com/openai/openai-python/blob/db3f3526f16ea518bcd16308508e035017f9259e/openai/api_requestor.py#L467

nero19960329:
close #26

### DIFF
--- a/anthropic/api.py
+++ b/anthropic/api.py
@@ -231,21 +231,21 @@ class Client:
         new_kwargs = {"stream": True, **kwargs}
         return self._request_as_stream(
             "post",
-            "/v1/complete",
+            "v1/complete",
             params=new_kwargs,
         )
 
     def completion(self, **kwargs) -> dict:
         return self._request_as_json(
             "post",
-            "/v1/complete",
+            "v1/complete",
             params=kwargs,
         )
 
     async def acompletion(self, **kwargs) -> dict:
         return await self._arequest_as_json(
             "post",
-            "/v1/complete",
+            "v1/complete",
             params=kwargs,
         )
 
@@ -253,7 +253,7 @@ class Client:
         new_kwargs = {"stream": True, **kwargs}
         return self._arequest_as_stream(
             "post",
-            "/v1/complete",
+            "v1/complete",
             params=new_kwargs,
         )
 


mikelambert:
Thanks for the quick fix! I kinda hate raw string concat for this kind of thing, what do you think of a different solution:
You pass up "http://localhost/anthropic/" instead of "http://localhost/anthropic".
And then change this PR from "/v1/complete" to be "v1/complete"

I believe this should work?
```
>>> from urllib.parse import urljoin
>>> urljoin("http://localhost/anthropic/", "/v1/complete")
'http://localhost/v1/complete'
>>> urljoin("http://localhost/anthropic/", "v1/complete")
'http://localhost/anthropic/v1/complete'
>>> urljoin("http://localhost/anthropic", "v1/complete")
'http://localhost/v1/complete'
```

nero19960329:
> Thanks for the quick fix! I kinda hate raw string concat for this kind of thing, what do you think of a different solution: You pass up "http://localhost/anthropic/" instead of "http://localhost/anthropic". And then change this PR from "/v1/complete" to be "v1/complete"
> 
> I believe this should work?
> 
> ```
> >>> from urllib.parse import urljoin
> >>> urljoin("http://localhost/anthropic/", "/v1/complete")
> 'http://localhost/v1/complete'
> >>> urljoin("http://localhost/anthropic/", "v1/complete")
> 'http://localhost/anthropic/v1/complete'
> >>> urljoin("http://localhost/anthropic", "v1/complete")
> 'http://localhost/v1/complete'
> ```

It works! Now I've changed my fixing plan.

ravintpillai:
any idea when this will get merged? Proposed fix looks sensible to me and would actually help us as well! :D

mikelambert (approved):

PR #34, opened 2023-05-15
Add locking to get_tokenizer

### LINKED ISSUE #33: `get_tokenizer` is thread-hostile
Race conditions during file modification between multiple calls to `get_tokenizer` can result in error such as `anthropic.tokenizer.TokenizerException: Failed to load tokenizer: EOF while parsing a value at line 1 column 0`.

`get_tokenizer` is thread-hostile. It should either have a docstring indicating that it's thread-hostile, or a lock should be used to make it thread-safe.

yifanmai:
Fixes #33

### DIFF
--- a/anthropic/tokenizer.py
+++ b/anthropic/tokenizer.py
@@ -1,23 +1,28 @@
 import os
 import tempfile
+import threading
 
 import httpx
 from tokenizers import Tokenizer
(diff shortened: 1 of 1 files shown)


jenan-ant:
See comments in #33.

PR #39, opened 2023-05-25
add getting headers from params

chitalian:


jenan-ant:
The new SDK version should support custom headers.

PR #42, opened 2023-06-03
making typing, imports and constants use more consistent

himalayanZephyr:


jenan-ant:
The new SDK version should have much better typing. Please take a look!

PR #53, opened 2023-07-01
fix(tokenizer): encoding issues for special characters

simingh124:
When tokenizing some non-English texts(such as Chinese) on windows os, there will be an issue causing by the default encoding format on reading the tokenizer.json:
image
image

By specifying " encoding='utf-8' ", this issue can be fixed:
image
image


rattrayalex:
Thanks for the report! We'll get this fixed within the next couple of days.

RobertCraigie (approved):
Looks good, thanks!

PR #77, opened 2023-07-18
Remove pydantic constraint via fallback imports to cover both v1 & v2

tomviner:
Pydantic is constrained to version 1 in pyproject.toml, this means the anthropic library can't be installed with any other library that's upgraded to pydantic version 2.

This is the case, with me trying to develop an [llm-claude](https://github.com/tomviner/llm-claude/) plugin for the [llm library](https://github.com/simonw/llm). I noted the problem in https://github.com/tomviner/llm-claude/issues/1

Luckily there's a workaround: pydantic have, in version 2, [retained a copy of all version 1 functionality](https://docs.pydantic.dev/2.0/migration/#continue-using-pydantic-v1-features) under the pydantic.v1 namespace. This PR attempts to import from this v1 namespace (for when v2 is installed), or upon ImportError (for when v1 is installed) import from pydantic as before.

This allows us to drop the constraint on the pydantic version.

### A note on testing

I couldn't see a documented way to run the tests, to check my changes worked. However after guessing a bit, I got them working against an openapi.yml file I (an AI) generated based on the API docs. I have not included that in this branch, but it did run against v1 and v2 of pydantic here:
- [pydantic changes on top of test runner branch](https://github.com/tomviner/anthropic-sdk-python/compare/main...tomviner:anthropic-sdk-python:pydantic-v2-compat-with-tests?expand=1)

If possible, I'd request some docs showing a better way for contributors to run the tests in future. Thanks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.7"
 httpx = ">= 0.23.0, < 1"
-pydantic = "^1.9.0"
+pydantic = "*"
 typing-extensions = ">= 4.1.1, < 5"
 anyio = ">= 3.5.0, < 4"
 distro = ">= 1.7.0, < 2"
--- a/src/anthropic/_base_client.py
+++ b/src/anthropic/_base_client.py
@@ -28,9 +28,11 @@ from typing_extensions import Literal, get_origin
 import anyio
 import httpx
 import distro
-import pydantic
 from httpx import URL, Limits
-from pydantic import PrivateAttr
+try:
+    from pydantic import v1 as pydantic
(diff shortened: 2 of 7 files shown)


rattrayalex:
Thanks very much for this @tomviner ! Wasn't aware of their `v1` import, that's very handy. 

We're going to take a look at this next week, and think about whether we want to do it this way, a pydantic shim module to import through, or just find a way to be compatible with pydantic v2. In the meantime, I appreciate your patience!

rattrayalex:
Hi @tomviner , wanted to let you know we're working to add full support for pydantic v2, without relying primarily on the compatibility shim. We expect to have that up early next week. Thank you for your generous contribution here!

PR #83, opened 2023-07-25
add context manager and close() method to Anthropic class

erees1:
Possible fix for: https://github.com/anthropics/anthropic-sdk-python/issues/84


rattrayalex:
Thank you for this! We took inspiration and added slightly more in https://github.com/anthropics/anthropic-sdk-python/pull/89

PR #111, opened 2023-08-10
fix(tests): suppress deprecation error of event_loop

natsukium:
The event_loop of pytest-asyncio should be closed explicitly, as shown in the following document.
https://github.com/pytest-dev/pytest-asyncio/tree/v0.18.3#event_loop
As of 0.18.3, which is fixed in pyproject.toml, it closes automatically, but as of 0.21.0, it throws an error. 

It would be good to fix this for future updates.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import asyncio
+from typing import Generator
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for reporting & fixing! This looks right at first blush; we'll take a deeper look at this tomorrow.

RobertCraigie:
Thanks for the fix @natsukium! We've pushed a fix separately that includes your changes and also bumps `pytest-asyncio` to the latest version in #114.

natsukium:
thanks!

PR #139, opened 2023-09-04
Support customizing the "verify" argument in requests.

q-x-iao:
Hi! I'm working on a project that needs to use Anthropic APIs via a proxy. The current API has almost everything we need, but since the proxy uses a custom certificate we also need to change the `verify` argument. This pull requests adds that ability.

I see that `_client.py` is generated by [Stainless](https://www.stainlessapi.com/), but I don't have access to it. Please let me know if I should move my changes to `_client.py` to somewhere else.

Thanks!

### DIFF
--- a/src/anthropic/_base_client.py
+++ b/src/anthropic/_base_client.py
@@ -5,6 +5,7 @@ import time
 import uuid
 import inspect
 import platform
+import ssl
 from types import TracebackType
 from random import random
 from typing import (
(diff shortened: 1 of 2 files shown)


rattrayalex:
Thank you for the feature request & the PR! We'll look into incorporating this. 

In the meantime, if needed, you can work around this by setting a custom `_client` property, like so:

```py
client = Anthropic()
client._client = httpx.Client(verify=False)
```
We also plan to expose better ways to provide a custom client.

dgellow:
@rattrayalex will correct me if I'm wrong, but I believe a solution to pass a custom client has been implemented with https://github.com/anthropics/anthropic-sdk-python/pull/173. The change will be part of the [next release](https://github.com/anthropics/anthropic-sdk-python/pull/152).

Would that close this issue, once released?

rattrayalex:
That's correct! Support for a custom http client has been released and you can now do:
```py
client = Anthropic(http_client=httpx.Client(verify=False))
```

PR #208, opened 2023-10-26
Fix link to anthropic-bedrock-python

uezo:
Add `//` to the beginning of the destination URL.


rattrayalex:
Thanks for catching & fix this! Due to some automations, we'll be releasing this in a separate PR shortly.

PR #408, opened 2024-03-21
Add Type Validation for API Key and Auth Token to Prevent Common Configuration Errors

MikeSoft:
While integrating the Anthropic API client, I encountered an issue where I mistakenly passed the api_key as a list instead of a string. This led to several hours of debugging to identify the root cause of the issue, as the current implementation does not explicitly validate the type of api_key and auth_token upon instantiation of the Anthropic client class.

To prevent such issues for other developers and to enhance the usability of the API client, I propose adding explicit type validation for both api_key and auth_token parameters within the __init__ method of the Anthropic class. This small yet impactful change ensures that developers are immediately alerted with a clear and informative error message if the types of these parameters do not match the expected string type.

The proposed change has been implemented as follows:

### DIFF
--- a/src/anthropic/_client.py
+++ b/src/anthropic/_client.py
@@ -105,10 +105,14 @@ class Anthropic(SyncAPIClient):
         """
         if api_key is None:
             api_key = os.environ.get("ANTHROPIC_API_KEY")
+        if type(api_key) is not str:
+            raise ValueError("api_key must be a string")
         self.api_key = api_key
 
         if auth_token is None:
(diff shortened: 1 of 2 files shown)


rattrayalex:
Thank you for sharing your experience and putting up a PR! We'll take a look.

RobertCraigie:
Thanks for the PR & for sharing context. I don't think the current implementation in this PR is quite correct as both api key and auth token can be `None`.

Unfortunately this is something we'll also want to change in our codegen and we won't be able to prioritise changing it any time soon so I'm going to close this PR for now. I really appreciate the effort put into creating this PR though!

Issue #424, opened 2024-03-28
stream.get_final_message() does not return the correct usage of output_tokens


sirius422:
Like the title said, `stream.get_final_message()` always return the `output_tokens` with the value of `1`. 
running the exmaple code [examples/messages_stream.py](https://github.com/anthropics/anthropic-sdk-python/blob/main/examples/messages_stream.py), the output would look like:
```
Hello there!
accumulated message:  {
  "id": "REDACTED",
  "content": [
    {
      "text": "Hello there!",
      "type": "text"
    }
  ],
  "model": "claude-3-opus-20240229",
  "role": "assistant",
  "stop_reason": "end_turn",
  "stop_sequence": null,
  "type": "message",
  "usage": {
    "input_tokens": 11,
    "output_tokens": 1
  }
}
```
However the actual `output_tokens` should be `6` according to the raw HTTP stream response
```
event: message_start
data: {"type":"message_start","message":{"id":"REDACTED","type":"message","role":"assistant","content":[],"model":"claude-3-opus-20240229","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":11,"output_tokens":1}}   }

event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}     }

event: ping
data: {"type": "ping"}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello"}              }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":" there"}  }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"!"}            }

event: content_block_stop
data: {"type":"content_block_stop","index":0         }

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":6}               }

event: message_stop
data: {"type":"message_stop"           }
```

So, is this a bug or a feature? I've seen someone in issue #417 using `stream.get_final_message()` to obtain the usage information. If `output_tokens ` always returns 1, this won't work properly, I guess.

sirius422:
Adding a line under `accumulate_event` in [src/anthropic/lib/streaming/_messages.py](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/streaming/_messages.py) file seems to fix the issue. Should I submit a pull request?

```
[...]

PR #425, opened 2024-03-28
Update the output_tokens after the stream has finished.

sirius422:
Fix issue #424, probably. I'm not sure if there are any omissions.


rattrayalex:
Thanks, we'll take a look!

RobertCraigie:
Thanks! I'll make this change on our side.

PR #458, opened 2024-04-17
fix: bedrock auth

Crescent617:
image


asengupta-pcty:
Hi, Do we know how long this will take to merge? It's blocking some dev work for our project

Also in the meantime, is there an earlier anthropic version we can pivot to for the time being?

rattrayalex:
Thanks for the PR @Crescent617 .

I expect we'll have the fix for this out later today (though we may fix in a separate branch).

RobertCraigie:
this has been fixed in 0.25.4

PR #468, opened 2024-04-19
Remove checks for `project_id` in `load_auth()`

ushuz:
`AnthropicVertex` accepts `project_id` from the caller. So whether `load_auth()` could load the `project_id` from the Google credential should not be a hard requirement.

Right now, if `load_auth()` can't load the `project_id`, it raises `ValueError` and aborts any further execution.


RobertCraigie:
Hey @ushuz, sorry for the delay in getting to this & thanks for the PR! I've merged a separate change which addresses this issue: https://github.com/anthropics/anthropic-sdk-python/pull/513

Issue #482, opened 2024-04-29
Source distribution tarballs on pypi do not contain README.md, causing hatch-fancy-pypi-readme plugin to fail


charles-dyfis-net:
The source distributions published to pypi presently cannot be built; attempts fail with:

```none
Processing /private/tmp/nix-build-python3.12-anthropic-0.25.7.drv-0/anthropic-0.25.7
  Running command Preparing metadata (pyproject.toml)
  Traceback (most recent call last):
    File "/nix/store/brnl7k3jxym5gyla9yh05xz5c4vhn1rz-python3.12-pip-23.2.1/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
      main()
    File "/nix/store/brnl7k3jxym5gyla9yh05xz5c4vhn1rz-python3.12-pip-23.2.1/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/nix/store/brnl7k3jxym5gyla9yh05xz5c4vhn1rz-python3.12-pip-23.2.1/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 152, in prepare_metadata_for_build_wheel
      whl_basename = backend.build_wheel(metadata_directory, config_settings)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/nix/store/w6r9jmj192v8mwga0zll0wq0kw02xnvn-python3.12-hatchling-1.18.0/lib/python3.12/site-packages/hatchling/build.py", line 56, in build_wheel
      return os.path.basename(next(builder.build(wheel_directory, ['standard'])))
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/nix/store/w6r9jmj192v8mwga0zll0wq0kw02xnvn-python3.12-hatchling-1.18.0/lib/python3.12/site-packages/hatchling/builders/plugin/interface.py", line 93, in build
      self.metadata.validate_fields()
[...]

PR #483, opened 2024-04-30
Include almost all source files in sdist (#482)

charles-dyfis-net:
Presently, the sdist published to pypi cannot be built (failing as shown in #482), because hatch-fancy-pypi-readme is configured to read `README.md` but that file is not included in the source distribution.

Here we make the sdist _almost_ all-inclusive (differing from `git archive` output only by leaving out only hidden files and directories -- presently `.github`, `.devcontainers`, `.python-version`, and `.stats.yml`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,20 @@ include = [
 [tool.hatch.build.targets.wheel]
 packages = ["src/anthropic"]
 
+[tool.hatch.build.targets.sdist]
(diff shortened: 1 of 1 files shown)


rattrayalex:
Thanks for the suggestion/contribution, we'll take a look!

rattrayalex:
thanks for the suggestion/contribution, this change has been incorporated!

PR #519, opened 2024-05-23
Move from `rye-up.com` to `rye.astral.sh`

charliermarsh:
## Summary

We're having issues with `rye-up.com` and so are moving Rye under the `astral.sh` domain. These changes should ensure that your CI keeps working with disruption.


rattrayalex:
Thanks Charlie! This change is already present on `next` and will be merged into `main` as part of the next release: https://github.com/anthropics/anthropic-sdk-python/pull/515

Issue #537, opened 2024-06-12
Add support for resolving AWS region from ~/.aws/config


jamesls:
The `AnthropicBedrock` client resolves the AWS region using the `AWS_REGION` env var or defaulting to `us-east-1`:

https://github.com/anthropics/anthropic-sdk-python/blob/9652512ed83473f359fc7a2fbf89d81f167a69c1/src/anthropic/lib/bedrock/_client.py#L119-L121

However, there are additional ways to configure region ([docs](https://docs.aws.amazon.com/sdkref/latest/guide/feature-region.html)).  In my case, I put all my configuration in `~/.aws/config` and don't set any environment variables other than `AWS_PROFILE` so `AnthropicBedrock` was defaulting to `us-east-1` instead of using my configured region.

You can get botocore to resolve the region using `get_config_variable` on the session instance:

```
>>> botocore.session.get_session().get_config_variable('region')
'us-west-2
```

namanvats:
Fixed it here : #539, please review and merge

EktorTyp:
Hey team, any updates on this? It seems to also be an issue with 3.7 on Bedrock. Even after setting the region in:

`
client = AnthropicBedrock(
    aws_region="us-east-1",
)
`

it always uses `us-east-2`

Same with 3.5 v2 that always defaults to `us-west-1`

RobertCraigie:
This was added recently!

PR #539, opened 2024-06-16
Added Support to fetch aws region from aws config file .aws/config #537

namanvats:
What does this PR do?
This PR adds support to fetch the AWS region name from the ~/.aws/config file and use the environment variable as a fallback if the aws config is absent. It defaults to us-east-1 in case both of them are not present. It leverages boto3 which is already present in requirements. So no changes to requirements are required.

Tries to fix the issue raised here : #537

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import boto3
 from typing import Any, Union, Mapping, TypeVar
 from typing_extensions import Self, override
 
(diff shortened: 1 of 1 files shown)


src/anthropic/lib/bedrock/_client.py
```diff
@@ -116,6 +117,8 @@ def __init__(
 
         self.aws_access_key = aws_access_key
 
+        session = boto3.Session()
+        aws_region = session.region_name
```
JDKardia, line 121: This is almost what we want! Ideally we'd want to set `aws_region` in order of specificity: 
1. specific region passed as argument
2. specific region passed as environment variable
3. specific region within config
4. default to us-east-1
 
however, as written in this PR, this would be:

1. specific region within config
2. specific region passed as argument
3. specific region passed as environment variable
4. default to us-east-1

if you could correct the ordering of assignment, that would be wonderful.

JDKardia (changes requested):
I like this, just needs a tweak!

rattrayalex:
Thank you for the PR. We hope to take a look soon.

src/anthropic/lib/bedrock/_client.py
feranmiodugbemi: @JDKardia, I fixed the issue here at #669, please review and merge

RobertCraigie:
Thanks for the PR and sorry for the delay here, this was shipped separately recently.

PR #542, opened 2024-06-17
feat(vertex): add credentials argument

mitraan-deshaw:
Currently, the `AnthropicVertex` class initializes its own default google.auth `Credentials` object. However, there are instances where we want to pass a `Credentials` object with custom service account information and scopes. For example:
```python
import json
from anthropic import AnthropicVertex
from google.oauth2 import service_account

credentials = service_account.Credentials.from_service_account_info(
    json.loads(google_credentials), 
    scopes=['https://www.googleapis.com/auth/cloud-platform.read-only']
)

client = AnthropicVertex(
    project_id=project_id,
    credentials=credentials,
    region=region_name
)
```

This PR implements the same for `AnthropicVertex` and `AsyncAnthropicVertex`.


RobertCraigie (approved):
Thanks!

Issue #564, opened 2024-06-27
AnthropicVertex stream chat generation is taking too much time


DhruvThu:
Recently, i have started using AnthropicVertex instead of direct anthropic. When I try to generate some data through AnthropicVertex client, it is taking around 2s to start streaming. However, in case of direct anthropic, it is not taking this much time. Also 2s duration is random, sometime it takes quite large amount of time and goes upto 6-10s. In worse case, it goes upto 20s. So, is there any que kind of stuff? I am using same code given in vertex ai anthropic notebook to generate responses. Is there any workaround which i need to complete to get response as fast as direct anthropic? If someone could guide me on this, it would be really helpful.

Thanks !!

aaron-lerner:
Hey @DhruvThu, can you share a few `id`s from the responses you get back on vertex requests? Or share a few [request ids](https://docs.anthropic.com/en/api/errors#request-id)? This will help us debug.

DhruvThu:
Thanks for responding. Also, I am using streaming response from vertex anthropic and these are some of the message ids which i got in first chunk. msg_01D2jNpu4rUZMXUvwtpipMnx, msg_01CMjRdPAhDQaWELbrgSirS8

aaron-lerner:
Hmm, message ids from vertex should look like `msg_vrtx_...`. The ids you shared are for the direct (1P) Anthropic API.

DhruvThu:
Could you check for this? msg_vrtx_01AaDL52fwpTrqFftLMxxQ1e. Sorry for the previous one. In this message, it took around 2.4s to start streaming. 
The streaming response through direct Anthropic API took around 0.89s. The message id for that is msg_01AWgnspZ2w5NhzE92uL7VZ9.


The code I am using is as follows, 
```
class AnthropicLLM:
    def __init__(self, anthropic_client : Anthropic, anthropic_vertex_client : AnthropicVertex) -> None:
        self.anthropic = anthropic_client
        credentials, project_id = google.auth.load_credentials_from_dict(google_credentials_info, scopes=["https://www.googleapis.com/auth/cloud-platform"],)
        anthropic_vertex_client._credentials = credentials
        self.vertex_anthropic = anthropic_vertex_client
        self.messages = Messages(self)

class Messages:
    def __init__(self, client : AnthropicLLM) -> None:
        self.client = client

[...]

PR #575, opened 2024-07-03
fix(vertex): avoid credentials refresh on every request

RobertCraigie:
Fixes https://github.com/anthropics/anthropic-sdk-python/issues/564. There will still be a delay for the first request but then the credentials will be cached.

I'm fairly nervous with this change will result in failures for certain cases as I'm not familiar enough with the google credentials system. Maybe we should ensure we try auth failures?

### DIFF
--- a/src/anthropic/lib/vertex/_client.py
+++ b/src/anthropic/lib/vertex/_client.py
@@ -168,13 +168,11 @@ class AnthropicVertex(BaseVertexClient[httpx.Client, Stream[Any]], SyncAPIClient
 
     @override
     def _prepare_request(self, request: httpx.Request) -> None:
-        access_token = self._ensure_access_token()
-
         if request.headers.get("Authorization"):
             # already authenticated, nothing for us to do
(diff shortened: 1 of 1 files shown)


aaron-lerner (approved):
This LGTM and I had a coworker take a look to confirm as well. He pointed out that there may be a slight danger of tokens expiring mid-request. You can have some multi-minute requests if token counts are high enough, but I'm not sure exactly what the failure mode looks like if that happens. It also looks like the google auth lib [gives](https://github.com/googleapis/google-auth-library-python/blob/1270217556a0249b6be8824e2b6209371f24e400/google/auth/credentials.py#L84) a 3.75 min buffer, so I think this is fine.

Thanks for the quick fix!

RobertCraigie:
thanks for the review! mid-request expiries is a good callout, that was my main concern as well, maybe we could perform our own expiry calculation with an even bigger skew? e.g. 10 mins as thats our default request timeout?

Issue #432, opened 2024-04-02
Bedrock Frequently Has Server-side Errors


kevinlu1248:
Likely unrelated to the client but I'm getting this pretty frequently on Anthropic Bedrock.

anthropic.APIStatusError: Error code: 424 - {'message': 'The system encountered an unexpected error during processing. Try your request again.'}

Also, it seems like this replace the other issue from: https://github.com/anthropics/anthropic-sdk-python/issues/418, so that issue should be resolved now.

costa-paxton:
I'm also coming up against this error though Bedrock. Would it make sense to simply add handling for 424 [in `BaseClient._should_retry`](https://github.com/anthropics/anthropic-sdk-python/blob/6cccfa7e76470f7ff3fd4e04889e904d11aa0bf4/src/anthropic/_base_client.py#L679)?

rattrayalex:
Ah, yes, that does sound like a good idea. We'll take a look at that.

mtliendo:
Is this something we should handle on our own in the meantime? Any guidance?

rattrayalex:
Hey, sorry for the delay in addressing this. This is really more of a Bedrock API issue – they should probably be returning a 5xx for this – but we'll take another look at merging #583 or equivalent.

sihil:
I added logic to work around this over six months ago and I've just persuaded my team not to merge a PR that layered on our own retries on top of the Anthropic retry system but now I'm not convinced that's right coming back here and realising this still isn't fixed after two years.

It would appear that the SDK's retry mechanisms are not being given very much attention and perhaps we would be best to disable them and just roll our own handling. At the moment we're already having to handle this 429 plus some other corner cases where the SDK seems to not retry on streaming operations.

Can Anthropic commit to getting on top of this, or perhaps they can more clearly document the limitations of the retry mechanisms embedded in the SDK so users know what else to layer on?

Edit: Just seen this seems to be at least partially owned by Stainless. Perhaps Anthropic should take ownership of their SDK.

dtmeadows-ant:
Gonna close this one given the staleness. Feel free to open a new one if these errors are still showing up!

PR #583, opened 2024-07-10
Feat: retry request on error 424

richzw:
Try to solve the issue https://github.com/anthropics/anthropic-sdk-python/issues/432


amadabhu (approved):
This looks ready to merge.

localden:
Thank you for the PR - `src/anthropic/__init__.py`, `src/anthropic/_base_client.py` are auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Closing.

PR #599, opened 2024-07-17
Prevent KeyError on Vertex retry.

yasyf:
The `json_data.pop("model")` currently causes `KeyError: 'model'` when the request is internally retried after a server error.


RobertCraigie:
Thanks for the PR! We've fixed this separately in #600

PR #648, opened 2024-08-30
feat(bedrock): add `profile` argument to client

khalilfiremind:
Add profile option to Bedrock client for profile selection

This update introduces a profile option to the Bedrock client, allowing users to select which profile to use when interacting with the Bedrock client.

In my scenario, I use SSO to access the AWS account. When attempting to run the code locally, it initially failed. By passing the profile_name in the **boto3.Session()**, the issue was resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,28 +201,14 @@ message = client.messages.create(
 print(message)
 ```
 
-or if you want to use a profile and region
+The bedrock client supports the following arguments for authentication
 
 ```py
-import os
-from anthropic import AnthropicBedrock
-
-PROFILE_NAME = os.getenv("PROFILE_NAME")
-REGION_NAME = os.getenv("REGION_NAME")
-
-client = AnthropicBedrock(aws_profile=PROFILE_NAME, aws_region=REGION_NAME)
-
-message = client.messages.create(
-    max_tokens=1024,
-    messages=[
-        {
-            "role": "user",
-            "content": "Hello!",
-        }
-    ],
-    model="anthropic.claude-3-sonnet-20240229-v1:0",
+AnthropicBedrock(
+  aws_profile='...',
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -201,6 +201,30 @@ message = client.messages.create(
...
+    messages=[
+        {
+            "role": "user",
+            "content": "Hello!",
+        }
+    ],
+    model="anthropic.claude-3-sonnet-20240229-v1:0",
+)
+print(message)
+```
```
RobertCraigie: nit: I think this example is slightly too big, having either a list of supported additional client arguments or a small snippet like this would be best imo:

The bedrock client supports the following arguments for authentication
```py
AnthropicBedrock(
  aws_profile='...',
  aws_region='us-east'
  ... etc
)
```

RobertCraigie (commented):
Looks good, thanks! Just a minor comment on the docs

README.md
khalilfiremind: done ✅ @RobertCraigie thank you!

README.md
```diff
@@ -201,6 +201,16 @@ message = client.messages.create(
 print(message)
 ```
 
+The bedrock client supports the following arguments for authentication
+
+```py
+AnthropicBedrock(
+  aws_profile='...',
+  aws_region='us-east'
+  ... etc
```
RobertCraigie: ah sorry I wasn't clear in my comment, I intended for `... etc` to be replaced with an exhaustive list of the supported options
```suggestion
  aws_secret_key='...',
  aws_access_key='...',
  aws_session_token='...',
```

khalilfiremind: Ahh fine, sorry, i did it @RobertCraigie

RobertCraigie (approved):
Thanks!

PR #649, opened 2024-09-01
chore: add example demonstrating batching many requests without hitting usage limits

mrkiura:
Adds an example that batches prompts to customizable batch sizes to avoid hitting rate limits.

The example assumes a user on the free tier, who is limited to 5 requests per minute. Other usage tiers can adjust the `batch_size` argument to the `make_batch` function based on the respective rates per minute.

### DIFF
--- a/examples/batch_rate_limit.py
+++ b/examples/batch_rate_limit.py
@@ -3,41 +3,51 @@
 import pprint
 import asyncio
 import itertools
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR! There is now a [Message Batches API](https://docs.anthropic.com/en/docs/build-with-claude/message-batches) so I'm going to close this.

PR #669, opened 2024-09-25
Fix #537: Update AWS region logic for flexibility and defaults

### LINKED ISSUE #537: Add support for resolving AWS region from ~/.aws/config

feranmiodugbemi:
This pull request fixes issue #537 by updating the AWS region logic in the `AnthropicBedrock` class. The new logic supports the following priority for setting the region:

- Argument passed during initialization
- Environment variable `AWS_REGION`
- AWS config file (`~/.aws/config`) using `boto3`
- Default to 'us-east-1' if none of the above are set

This ensures a more flexible setup for AWS region configuration.

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 from typing import Any, Union, Mapping, TypeVar
 from typing_extensions import Self, override
+import boto3
 
 import httpx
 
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR and sorry for the delay, this was fixed separately recently!

PR #682, opened 2024-10-08
refactor(types): improve metadata types

RobertCraigie:
We now properly define a single `MetadataParam` type instead of duplicating the `Metadata` param type everywhere.


kiankyars:
MetadataParam: It would be great to have some common documentation examples for how to use a metadata param.

PR #798, opened 2024-12-17
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Feel free to ignore... just was lurking around and spotted typos.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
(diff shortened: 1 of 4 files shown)


RobertCraigie:
Thanks for the PR! I don't think we'll want to add this just yet, we use cspell internally in our codegen system and I suspect maintaining separate word lists will be more trouble than it's worth.

Thanks for spotting the `overriden` typo though, I'll fix that separately :)

PR #805, opened 2024-12-28
refactor(stream): make `MessageStream` wrap `Stream` directly

alexmojaki:
This is intended to address https://github.com/pydantic/logfire/issues/672. We patch the client `_request` methods to replace the `stream_cls` argument with a wrapper class that logs streamed values. But that wrapper gets lost by `MessageStream` and `MessageStreamManager`. This PR not only fixes the issue for us but also seems cleaner and more logical within this codebase - why recreate the `Stream` with the same arguments? It seems like a relic from when `MessageStream` was a subclass of `Stream` and passing the same arguments to `__init__` made more sense.

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -18,9 +18,6 @@ from ..._utils import consume_sync_iterator, consume_async_iterator
 from ..._models import build, construct_type
 from ..._streaming import Stream, AsyncStream
 
-if TYPE_CHECKING:
-    from ..._client import Anthropic, AsyncAnthropic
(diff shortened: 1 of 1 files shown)


RobertCraigie (approved):
Thanks!

> This PR not only fixes the issue for us but also seems cleaner and more logical within this codebase - why recreate the Stream with the same arguments? It seems like a relic from when MessageStream was a subclass of Stream and passing the same arguments to __init__ made more sense.

yup spot on

---

note that this is technically a breaking change as `MessageStream.__init__` is public API but I'd be surprised if anyone is relying on it

Issue #485, opened 2024-05-01
Casting to InputSchema drops $defs, causing an InternalServerError


charles-dyfis-net:
Consider a tool-calling request with the following schema:

```
{'$defs': {'TestAnimalEnum': {'enum': ['cat', 'dog', 'bird'], 'type': 'string'}},
 'description': 'Describe a pet',
 'properties': {'name': {'description': 'Name of the pet', 'title': 'Name', 'type': 'string'},
                'species': {'allOf': [{'$ref': '#/$defs/TestAnimalEnum'}], 'description': 'species of the pet'}},
 'required': ['species', 'name'],
 'title': 'describe_pet',
 'type': 'object'}
```

Anthropic correctly handles this so long as `$defs` is present; should it be dropped, the `enum` fields become unresolvable, causing an error of the form of:

```none
anthropic.InternalServerError: Error code: 500 - {'type': 'error', 'error': {'type': 'api_error', 'message': 'Internal server error'}}
```

Unfortunately, using `pydantic.TypeAdapter(anthropic.types.beta.tools.tool_param.InputSchema).validate_python(tool.model_json_schema())` drops the `$defs` field from the representation -- after getting the aforementioned HTTP 500 error, if we walk the stack up to `anthropic._base_client.post`, we see a body where `$defs` is not present at all, despite references to it existing:

```
> /nix/store/kpyj1ihx319v9y8kcq0bxvn54zvh52s5-python3-3.12.0-env/lib/python3.12/site-packages/anthropic/_base_client.py(1782)post()
-> return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
(Pdb) pprint.pprint(body)
{'max_tokens': 2048,
 'messages': [{'content': 'Describe a typical pet owned by an 12-year-old', 'role': 'user'}],
 'metadata': NOT_GIVEN,
 'model': 'claude-3-haiku-20240307',
 'stop_sequences': NOT_GIVEN,
 'stream': NOT_GIVEN,
 'system': NOT_GIVEN,
 'temperature': NOT_GIVEN,
 'tools': [{'description': 'Describe a pet',
            'input_schema': {'properties': {'name': {'description': 'Name of the pet',
                                                     'title': 'Name',
                                                     'type': 'string'},
                                            'species': {'allOf': [{'$ref': '#/$defs/TestAnimalEnum'}],
                                                        'description': 'species of the pet'}},
                             'type': 'object'},
[...]

PR #821, opened 2025-01-14
anthropic.types.tool_param.InputSchema: Explicitly define allowed top-level keys (#485)

charles-dyfis-net:
The original fix to #485 does not actually work: pydantic discards fields other than `type` and `properties` when coercing to an `InputSchema`.

This is an alternate approach that explicitly sets out the fields allowed by the JSONSchema specification, ensuring that none of these fields will be stripped.


charles-dyfis-net:
Mooted by d0961c2fcbe7370f145512facad8aab175798158

PR #910, opened 2025-03-23
feat: Add offline mode for testing and development

campaign940:
## Summary
  - Added an offline mode feature that allows testing and development without making actual API calls
  - Implemented mock responses for easier local development and testing

## Test plan
  - Verify offline mode correctly returns mock responses
  - Check that example code works as expected in offline mode
  - Ensure proper documentation of the new feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,742 +1,204 @@
-# Anthropic Python API library
+# Python SDK for Anthropic API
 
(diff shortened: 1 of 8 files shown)


RobertCraigie:
Thanks for the PR! However I don't think we'll want to add this to the SDK as you can already achieve the same thing using respx https://github.com/lundberg/respx

Issue #930, opened 2025-04-08
AttributeError: module 'socket' has no attribute 'TCP_KEEPINTVL' on Windows and macOS


ksaurabh02:
I'm trying to use the anthropic-sdk on both Windows and macOS, but I'm encountering the following error on both platforms:

```
anthropic\_base_client.py:795 Message: AttributeError: module 'socket' has no attribute 'TCP_KEEPINTVL'
```

After investigating, I found that the socket.TCP_KEEPINTVL attribute isn't available on all platforms, particularly Windows and macOS.

To resolve the issue locally, I modified _base_client.py as follows:

```
if "transport" not in kwargs:
    socket_options = [
        (socket.SOL_SOCKET, socket.SO_KEEPALIVE, True)
    ]

    if getattr(socket, 'TCP_KEEPINTVL', None) is not None:
        socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 60)]
    elif sys.platform == 'darwin':
        TCP_KEEPALIVE = getattr(socket, 'TCP_KEEPALIVE', 0x10)
        socket_options += [(socket.IPPROTO_TCP, TCP_KEEPALIVE, 60)]

    if getattr(socket, 'TCP_KEEPCNT', None) is not None:
        socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 5)]

    TCP_KEEPIDLE = getattr(socket, "TCP_KEEPIDLE", None)
    if TCP_KEEPIDLE is not None:
        socket_options.append((socket.IPPROTO_TCP, TCP_KEEPIDLE, 60))
```
With this change, the SDK works as expected on both platforms.

Would this be a valid fix to submit as a PR? Happy to open one if this approach looks good.

RobertCraigie:
I'm surprised you're seeing that error on macos, I was under the impression it was always available on macos 🤔

PR would be great, thank you!

ksaurabh02:
@RobertCraigie Sure, created a PR https://github.com/anthropics/anthropic-sdk-python/pull/935

ksaurabhAparavi:
@RobertCraigie Please have a look at the PR.

RobertCraigie:
This should be fixed as the PR was merged!

PR #935, opened 2025-04-10
fix(sockets): handle non-portable socket flags

ksaurabh02:
I'm trying to use the anthropic-sdk on both Windows and macOS, but I'm encountering the following error on both platforms:

```
anthropic\_base_client.py:795 Message: AttributeError: module 'socket' has no attribute 'TCP_KEEPINTVL'
```

After investigating, I found that the socket.TCP_KEEPINTVL attribute isn't available on all platforms, particularly Windows and macOS.

To resolve the issue locally, I modified _base_client.py as follows:

```
if "transport" not in kwargs:
    socket_options = [
        (socket.SOL_SOCKET, socket.SO_KEEPALIVE, True)
    ]

    if getattr(socket, 'TCP_KEEPINTVL', None) is not None:
        socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 60)]
    elif sys.platform == 'darwin':
        TCP_KEEPALIVE = getattr(socket, 'TCP_KEEPALIVE', 0x10)
        socket_options += [(socket.IPPROTO_TCP, TCP_KEEPALIVE, 60)]

    if getattr(socket, 'TCP_KEEPCNT', None) is not None:
        socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 5)]

    TCP_KEEPIDLE = getattr(socket, "TCP_KEEPIDLE", None)
    if TCP_KEEPIDLE is not None:
        socket_options.append((socket.IPPROTO_TCP, TCP_KEEPIDLE, 60))
```
With this change, the SDK works as expected on both platforms.

Issue link: https://github.com/anthropics/anthropic-sdk-python/issues/930

### DIFF
--- a/src/anthropic/_base_client.py
+++ b/src/anthropic/_base_client.py
@@ -794,14 +794,17 @@ class _DefaultHttpxClient(httpx.Client):
                 (socket.SOL_SOCKET, socket.SO_KEEPALIVE, True)
             ]
 
-            if getattr(socket, 'TCP_KEEPINTVL', None) is not None:
-                socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 60)]
-            elif sys.platform == 'darwin':
-                TCP_KEEPALIVE = getattr(socket, 'TCP_KEEPALIVE', 0x10)
-                socket_options += [(socket.IPPROTO_TCP, TCP_KEEPALIVE, 60)]
+            TCP_KEEPINTVL = getattr(socket, "TCP_KEEPINTVL", None)
+            
+            if TCP_KEEPINTVL is not None:
+                socket_options.append((socket.IPPROTO_TCP, TCP_KEEPINTVL, 60))
+            elif sys.platform == "darwin":
+                TCP_KEEPALIVE = getattr(socket, "TCP_KEEPALIVE", 0x10)
+                socket_options.append((socket.IPPROTO_TCP, TCP_KEEPALIVE, 60))
 
-            if getattr(socket, 'TCP_KEEPCNT', None) is not None:
-                socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 5)]
+            TCP_KEEPCNT = getattr(socket, "TCP_KEEPCNT", None)
+            if TCP_KEEPCNT is not None:
+                socket_options.append((socket.IPPROTO_TCP, TCP_KEEPCNT, 5))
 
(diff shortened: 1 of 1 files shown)


src/anthropic/_base_client.py
```diff
@@ -791,10 +791,21 @@ def __init__(self, **kwargs: Any) -> None:
...
+                (socket.SOL_SOCKET, socket.SO_KEEPALIVE, True)
             ]
+
+            TCP_KEEPINTVL = getattr(socket, "TCP_KEEPINTVL", None)
+            
+            if TCP_KEEPINTVL is not None:
+                socket_options.append((socket.IPPROTO_TCP, TCP_KEEPINTVL, 60))
+            elif sys.platform == "darwin":
+                TCP_KEEPALIVE = getattr(socket, "TCP_KEEPALIVE", 0x10)
+                socket_options.append((socket.IPPROTO_TCP, TCP_KEEPALIVE, 60))
```
RobertCraigie, line 803: question: why is this an `elif` and not just an `if TCP_KEEPALIVE is not None` check?

ksaurabhAparavi: > question: why is this an `elif` and not just an `if TCP_KEEPALIVE is not None` check?

I used `elif sys.platform == "darwin"` instead of a separate if to ensure that only set one keepalive socket option.
If TCP_KEEPINTVL is available (e.g., on Linux), set it and skip the macOS-specific fallback.
Using elif prevents setting both TCP_KEEPINTVL and TCP_KEEPALIVE together, which could lead to conflicting socket behavior.

RobertCraigie (commented):
Thanks! Looks like you have a lint error and I think we'll need to apply this change to the [async](https://github.com/ksaurabh02/anthropic-sdk-python/blob/d7b9d711e4643291476000f21641971a7fca6968/src/anthropic/_base_client.py#L1405) client as well.

Let me know if you need any help resolving the lint errors.

ksaurabhAparavi:
> Thanks! Looks like you have a lint error and I think we'll need to apply this change to the [async](https://github.com/ksaurabh02/anthropic-sdk-python/blob/d7b9d711e4643291476000f21641971a7fca6968/src/anthropic/_base_client.py#L1405) client as well.
> 
> Let me know if you need any help resolving the lint errors.

@RobertCraigie: Added change for the async function and fixed the lint errors.

RobertCraigie (approved):
Thanks!

PR #946, opened 2025-05-07
fix(vertex,bedrock): Added missing stream and batches method to vertex and bedrock beta libs

rusiaaman:
Bedrock and vertex client doesn't currently expose .beta.messages.stream and .beta.messages.batches


RobertCraigie:
Thanks! Have you tested these methods against the live API for both Bedrock and Vertex? I suspect there might be some more changes required for the batches API to work.

rusiaaman:
I have tested the stream method for vertex api successfully. I haven't tested the other methods yet. (I'll update)

rusiaaman:
I've tested both vertex and bedrock for sync stream functions successfully.

Batch api is not working for both. I've removed the batch method.

I haven't tested async stream method.

rusiaaman:
Should this be converted to an issue instead?

PR #959, opened 2025-05-27
feat(vertex): support global region endpoint

StephenGrider:
Vertex supports a `global` region, which redirects requests to any available region serving the model you're trying to access.  

When specifying a `global` region, the expected endpoint URL changes from `https://{region}-aiplatform.googleapis.com/v1` to `https://aiplatform.googleapis.com/v1`.

This PR adds support for the `global` region.


mitraan-deshaw:
Ran into this while working on https://github.com/langchain-ai/langchain-google/pull/983 - when specifying the global region, it shouldn't be necessary to set the `base_url`. 

Could someone please review this PR? Tagging @RobertCraigie.

If it helps, langchain-google-vertexai does something similar as well - https://github.com/langchain-ai/langchain-google/blob/dca0c9580ba09053d155fd07df26c0a7a30a107a/libs/vertexai/langchain_google_vertexai/_base.py#L137-L140

StephenGrider:
Fixed in #983

PR #962, opened 2025-05-31
fix(docs): remove reference to rye shell

TurtleShip:
👋 Hi, this small PR removes a reference to 'rye shell'

The command was removed from rye due to https://github.com/astral-sh/rye/issues/601 ( and through https://github.com/astral-sh/rye/pull/602 )

The Rye's author recommends using venv, which this contributing guide already recommends, so I just deleted reference to `rye shell`.


RobertCraigie (approved):
Thanks!

PR #969, opened 2025-06-09
fix(httpx): resolve conflict between default transport and proxy settings

### LINKED ISSUE #878: https_proxy not working when upgrade to v0.47.0
https_proxy not working when upgrade to v0.47.0, it works fine in v0.46.0

karpetrosyan:
closes #878


src/anthropic/_base_client.py
```diff
@@ -1364,11 +1378,23 @@ def __init__(self, **kwargs: Any) -> None:
...
-                socket_options=socket_options,
-            )
+            proxy_map = {key: None if url is None else Proxy(url=url) for key, url in get_environment_proxies().items()}
+
+            transport_kwargs = {
+                arg: kwargs[arg] for arg in ("verify", "cert", "trust_env", "http1", "http2", "limits") if arg in kwargs
+            }
+
+            proxy_mounts = {
+                key: None if proxy is None else HTTPTransport(proxy=proxy, **transport_kwargs)
```
RobertCraigie: ah I think you forgot to update this to `AsyncHTTPTransport`?

src/anthropic/_base_client.py
```diff
@@ -38,9 +38,11 @@
 import httpx
 import distro
 import pydantic
-from httpx import URL
+from httpx import URL, Proxy, HTTPTransport
 from pydantic import PrivateAttr
 
+from anthropic._utils._httpx import get_environment_proxies
```
RobertCraigie: nit: we use relative imports everywhere

src/anthropic/_utils/_httpx.py
RobertCraigie, line 1: nit: would be helpful to include a comment at the top of this file linking it to httpx and the commit you copied it from

tests/test_client.py
```diff
@@ -928,6 +928,14 @@ def retry_handler(_request: httpx.Request) -> httpx.Response:
             assert response.retries_taken == failures_before_success
             assert int(response.http_request.headers.get("x-stainless-retry-count")) == failures_before_success
 
+    def test_proxy_environment_variables(self, monkeypatch: Any):
+        monkeypatch.setenv("HTTPS_PROXY", "https://example.org")
+        client = Anthropic()
+
+        mounts = tuple(client._client._mounts.items())
+        assert len(mounts) == 1
+        assert mounts[0][0].pattern == "https://"
```
RobertCraigie, line 938: nit: can we include a check for the hostname here as well to make sure it's set to example.org?

src/anthropic/_base_client.py
```diff
@@ -1367,11 +1380,23 @@ def __init__(self, **kwargs: Any) -> None:
...
+            }
+
+            proxy_mounts = {
+                key: None if proxy is None else AsyncHTTPTransport(proxy=proxy, **transport_kwargs)
+                for key, proxy in proxy_map.items()
+            }
+            default_transport = AsyncHTTPTransport(**transport_kwargs)
+
+            # Prioritize the mounts set by the user over the environment variables.
+            kwargs["mounts"] = proxy_mounts | kwargs.get("mounts", {})
```
RobertCraigie: nit: I think the `|` syntax here is Python 3.9 only? We currently support 3.8

src/anthropic/_base_client.py
```diff
@@ -1367,11 +1380,23 @@ def __init__(self, **kwargs: Any) -> None:
...
 
-            kwargs["transport"] = httpx.AsyncHTTPTransport(
-                # note: limits is always set above
-                limits=kwargs["limits"],
-                socket_options=socket_options,
-            )
+            proxy_map = {key: None if url is None else Proxy(url=url) for key, url in get_environment_proxies().items()}
+
+            transport_kwargs = {
+                arg: kwargs[arg] for arg in ("verify", "cert", "trust_env", "http1", "http2", "limits") if arg in kwargs
```
RobertCraigie, line 1387: would it be feasible to add unit tests for all of these args as well?

tests/test_client.py
karpetrosyan: It's actually not a string, but a `transport` object that hides how it retrieves the response. I decided to avoid checking the hostname to prevent diving into deep internals that might change frequently and break compatibility. I'm only checking the `._mounts` property, which I believe will remain stable.

src/anthropic/_base_client.py
karpetrosyan: I guess the `|` syntax is used in many places; that's because you have future annotations included at the top level, so I guess it's okay.

src/anthropic/_base_client.py
karpetrosyan: I guess it's the same problem here. The transport doesn't expose any attributes, so we could write some references to the internal variables if it's worth it. But I think we can avoid that because we are just testing here whether the arguments were correctly passed to httpx.

RobertCraigie: maybe the best tests here would just be 2 tests like this?
1. you can pass the arguments without any errors happening
2. mock the transport class to just make sure the arguments are passed to it?

I think #2 could be nice to have but #1 would provide the most value imo

karpetrosyan: Added a simple smoke test to ensure there are no exceptions when providing arguments to the default client. Also suppressed the deprecation warning because httpx deprecated cert and will remove it in the major release.

RobertCraigie (approved):
Very nice, thank you!

Issue #892, opened 2025-03-04
Bedrock client failing to detect AWS region correctly can cause cross-region inference to fail


charles-dyfis-net:
I'm using an `AWS_PROFILE` environment variable where `~/.aws/config` specifies `region = "us-west-2"`.

Using `boto3.client("bedrock-runtime")` to make requests of cross-region endpoints, it properly attempts to communicate with `https://bedrock-runtime.us-west-2.amazonaws.com`.

By contrast, using `anthropic.AnthropicBedrock()` is defaulting to `us-east-1`, and when attempting to use a system-defined cross-region inference profile (as required for Sonnet 3.7) with credentials generated for us-west-2, https://bedrock-runtime.us-east-1.amazonaws.com is throwing a HTTP 403 (without enough supporting detail to make the cause of this failure readily apparent).

Insofar as the dependency set of anthropic-sdk-python includes both `boto3` and `botocore`, consider using Amazon's libraries to more accurately populate the region, rather than falling back to hardcoded `us-east-1` whenever `AWS_REGION` is unset.

Code to do this might look like:

```
try:
    default_bedrock_region = typing.cast(str, boto3.client("bedrock").meta.region_name)
except botocore.exceptions.NoRegionError:
    default_bedrock_region = 'us-east-1' # fall back to legacy behavior; maybe log?
```

...whereafter `default_bedrock_region` can be used as the fallback when no region is explicitly set, rather than the hardcoded `us-east-1` literal.

Ashutosh0x:
Hey! This issue has been fixed.

Looking at the current code in [`_client.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/bedrock/_client.py#L70-L90), there's now an `_infer_region()` function that does exactly what you suggested:

```python
def _infer_region() -> str:
    aws_region = os.environ.get("AWS_REGION")
    if aws_region is None:
        try:
            import boto3
            session = boto3.Session()
            if session.region_name:
                aws_region = session.region_name
        except ImportError:
            pass

    if aws_region is None:
        log.warning("No AWS region specified, defaulting to us-east-1")
        aws_region = "us-east-1"

    return aws_region
```

This is called in the `AnthropicBedrock` constructor (line 164):
```python
self.aws_region = _infer_region() if aws_region is None else aws_region
```

So the SDK now properly detects the region from your AWS profile via `boto3.Session().region_name` before falling back to `us-east-1`.

[...]

PR #974, opened 2025-06-13
feat(bedrock): automatically infer AWS Region in Bedrock clients

karpetrosyan:
closes #892 and #537


.github/workflows/ci.yml
```diff
@@ -7,6 +7,8 @@ on:
       - 'integrated/**'
       - 'stl-preview-head/**'
       - 'stl-preview-base/**'
+  pull_request:
```
dtmeadows: mind removing this before merging?

dtmeadows (changes requested):
This overall LGTM @karpetrosyan! Could we just remove the workflow change before merging?

.github/workflows/ci.yml
karpetrosyan: yes, sure!

karpetrosyan: done

dtmeadows (approved):

karpetrosyan:
@dtmeadows fyi, seems like I can't merge this

dtmeadows:
@michael-cohen-io mind reviewing and merging this when you get a sec?

michael-cohen-io (approved):

src/anthropic/lib/bedrock/_client.py
```diff
@@ -64,6 +67,29 @@ def _prepare_options(input_options: FinalRequestOptions) -> FinalRequestOptions:
...
+    """
+    Infer the AWS region from the environment variables or
+    from the boto3 session if available.
+    """
+    aws_region = os.environ.get("AWS_REGION")
+    if aws_region is None:
+        try:
+            import boto3
+
+            session = boto3.Session()
```
RobertCraigie, line 80: actually do you know if this can make fs calls?

karpetrosyan: I don't fully understand what 'making fs calls' means. It has access to the standard library, which I assume it uses to read the AWS config file.

RobertCraigie: ah sorry, yeah I meant "does it call blocking IO functions" which if it does we should try and call it asynchronously...

but that might be difficult?

karpetrosyan: It definitely does, but I think it's okay. It won't block the event loop repeatedly—just once during initialization

RobertCraigie: users may instantiate the client in their business logic inside async functions so I think it is somewhat important unfortunately.

if wrapping this in async is too annoying, I think for now we could cache the `boto3.Session` object?

karpetrosyan: > if wrapping this in async is too annoying, I think for now we could cache the boto3.Session object?

In that case, users with an async codebase will still be blocked the first time they call it — am I getting that right? Honestly, I don't think it's worth doing anything about. Many clients block for a few milliseconds during initialization. For example, `httpx.AsyncClient` takes around `50-150ms` to create (which means the event loop is blocked for that time), while `boto3.Session()` blocks for just 5ms. Programmers should avoid creating such objects inside a (for/while) loop — that would kill performance anyway (they're supposed to be created once). So if it's done only once, is a 5ms delay really that critical?

you can check setup time with this script:

```python
# /// script
# requires-python = ">=3.9"
# dependencies = [
#     "boto3",
#     "httpx",
# ]
# ///

import boto3
import time
import httpx

t1 = time.monotonic_ns()
boto3.Session()
t2 = time.monotonic_ns()

print(f"boto3 Time taken: {(t2 - t1) / 1_000_000} ms")

t1 = time.monotonic_ns()
httpx.Client()
t2 = time.monotonic_ns()

print(f"httpx Time taken: {(t2 - t1) / 1_000_000} ms")
```

PR #978, opened 2025-06-18
Rename `http_client` on `http_async_client` in Anthropic(Client)

TulupovKirill:
Split _http_client_ for each client (sync and async) in the Client(Anthropic) constructor.


RobertCraigie:
Thanks for the PR but we don't want to do this, it'll be a breaking change.

PR #981, opened 2025-06-23
fix(tests): fix some failing tests on the latest python versions

karpetrosyan:
`test_copy_build_request` test fails when running tests with Python ≥ 3.10 (by changing the .python-version file or using the UV_PYTHON environment variable) due to a memory leak.


RobertCraigie:
I think this just needs to include the deprecation warning config change? The others should be part of the codegen change

karpetrosyan:
Ah, yes... I'm lost in all of this, sorry:d

RobertCraigie:
no worries!

PR #1037, opened 2025-09-24
Fix docstring inconsistencies in `src/antrhopic/lib` folder.

northern-64bit:
Fixes minor docstring inconsistencies in the `src/anthropic/lib` folder, since that code is not generated.


karpetrosyan:
Thanks for the PR, but I think we have docstrings without a period at the end in our SDK, so if we want to follow a specific format, we should apply it everywhere.

PR #1038, opened 2025-09-28
fix file upload 404 due to Omit not being stripped from file upload request

BryanMakers:
The function `anthropic_client.beta.files.upload` returns error code 404. Because the object Omit is not being stripped away from the request header, which overrides the otherwise hardcoded beta string for file upload, [see](https://github.com/anthropics/anthropic-sdk-python/blob/8bd83a10f22ee8ca8564d4bee99c0451aaace849/src/anthropic/resources/beta/files.py#L303C1-L304C1).

Without this fix, it renders [this piece](https://docs.claude.com/en/docs/build-with-claude/files#uploading-a-file) of the python sdk example from the Anthropic's official documentation unusable

This is my first time contributing to a public git repo, so I hope I'm doing this right :) I do notice the [contributing doc](https://github.com/anthropics/anthropic-sdk-python?tab=contributing-ov-file) says code are being generated, so was I supposed to generate my code using my own code agent or something?

### DIFF
--- a/src/anthropic/_utils/_utils.py
+++ b/src/anthropic/_utils/_utils.py
@@ -295,7 +295,7 @@ def strip_not_given(obj: None) -> None: ...
 
 
 @overload
-def strip_not_given(obj: Mapping[_K, _V | NotGiven]) -> dict[_K, _V]: ...
+def strip_not_given(obj: Mapping[_K, _V | NotGiven | Omit]) -> dict[_K, _V]: ...
 
 
 @overload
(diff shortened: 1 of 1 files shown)


RobertCraigie:
Thanks for the PR! This isn't quite the right fix as it means we'd strip header values too early that shouldn't be stripped if they were set to `Omit()`.

The correct fix here is to update our handling of the beta header to fallback to `NotGiven()` instead of `Omit()`, I'm pushing a fix to our codegen with that change.

PR #1057, opened 2025-10-31
fix(bedrock): preserve http_client in copy() and with_options() methods

ljun20160606:
Fixes an issue where the http_client parameter was not being passed through when copying AnthropicBedrock and AsyncAnthropicBedrock instances. This caused the copied client to create a new http_client instead of reusing the existing one.

Adds comprehensive test coverage for copy() and with_options() methods including:
- Basic copy functionality
- AWS credential overrides
- Default options (max_retries, timeout)
- Default headers and query parameters
- set_default_headers and set_default_query behavior


ljun20160606:
Hi @samj-anthropic, could you help to take a look?

PR #1058, opened 2025-11-02
fix: convert paths to POSIX format for cross-platform compatibility

### LINKED ISSUE #1051: skills.create requires posix paths
When calling client.beta.skills.create from Windows using Anthropic's _files APIs, as [this sample does](https://github.com/anthropics/claude-cookbooks/blob/main/skills/notebooks/03_skills_custom_development.ipynb) , I get a  BadRequestError("Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'Skill contains path with invalid characters'},..

I did not see anywhere in the docs that posix paths are required.  Did I miss it? 
If I did not miss it, It might be useful to document that, or to amend the [_collect_files ](https://github.com/anthropics/anthropic-sdk-python/blob/4022488724e368805336a9f1d013135d5035415f/src/anthropic/lib/_files.py#L19) function. 

Cheers!! 
Jaime

yashwantbezawada:
Fix Windows path handling in skills.create

On Windows, calling `client.beta.skills.create()` fails with "Skill contains path with invalid characters". This happens because `_collect_files()` uses `str(path.relative_to())` which produces backslash-separated paths on Windows (e.g., `dir\file.txt`), but the API expects forward slashes.

Changed to use `.as_posix()` instead, which converts paths to POSIX format on all platforms.

Fixes #1051

### DIFF
--- a/src/anthropic/lib/_files.py
+++ b/src/anthropic/lib/_files.py
@@ -22,7 +22,7 @@ def _collect_files(directory: Path, relative_to: Path, files: list[FileTypes]) -
             _collect_files(path, relative_to, files)
             continue
 
-        files.append((str(path.relative_to(relative_to)), path.read_bytes()))
+        files.append((path.relative_to(relative_to).as_posix(), path.read_bytes()))
 
 
 async def async_files_from_dir(directory: str | os.PathLike[str]) -> list[FileTypes]:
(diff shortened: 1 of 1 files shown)


yashwantbezawada:
Closing to restructure commit.

PR #1061, opened 2025-11-05
AttributeError for python 3.14: 'typing.Union' object has no attribute __discriminator__ and no __dict__ for setting new attributes #1

lencshu:
Please using Annotated[Union] with Python 3.14

**Describe the bug**

When casting:

```
cast(CachedDiscriminatorType, union).__discriminator__ = details
```

With union being:
```
anthropic.types.raw_message_start_event.RawMessageStartEvent | anthropic.types.raw_message_delta_event.RawMessageDeltaEvent | anthropic.types.raw_message_stop_event.RawMessageStopEvent | anthropic.types.raw_content_block_start_event.RawContentBlockStartEvent | anthropic.types.raw_content_block_delta_event.RawContentBlockDeltaEvent | anthropic.types.raw_content_block_stop_event.RawContentBlockStopEvent
```

**To Reproduce**

I’m getting an error when running the MCP tool call

```
  File "/app/.venv/lib/python3.14/site-packages/anthropic/_streaming.py", line 185, in __aiter__
    async for item in self._iterator:
        yield item
  File "/app/.venv/lib/python3.14/site-packages/anthropic/_streaming.py", line 214, in __stream__
    yield process_data(data=data, cast_to=cast_to, response=response)
          ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.14/site-packages/anthropic/_base_client.py", line 641, in _process_response_data
    return cast(ResponseT, construct_type(type_=cast_to, value=data))
                           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.14/site-packages/anthropic/_models.py", line 518, in construct_type
    discriminator = _build_discriminated_union_meta(union=type_, meta_annotations=meta)
  File "/app/.venv/lib/python3.14/site-packages/anthropic/_models.py", line 688, in _build_discriminated_union_meta
    cast(CachedDiscriminatorType, union).__discriminator__ = details
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'typing.Union' object has no attribute '__discriminator__' and no __dict__ for setting new attributes
```

`CachedDiscriminatorType` should be an Annotated rather than Union directly

**OS**
macos

**Python version**
Python v3.14.0

**Library version**
anthropic v0.72.0

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -685,7 +685,7 @@ def _build_discriminated_union_meta(*, union: type, meta_annotations: tuple[Any,
         discriminator_field=discriminator_field_name,
         discriminator_alias=discriminator_alias,
     )
-    cast(CachedDiscriminatorType, union).__discriminator__ = details
+    cast(CachedDiscriminatorType, Annotated[union, details])
     return details
 
 
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import inspect
 from typing import TYPE_CHECKING, Any, Type, Union, Generic, TypeVar, Callable, Optional, cast
+from weakref import WeakKeyDictionary
 from datetime import date, datetime
 from typing_extensions import (
     List,
@@ -73,7 +74,6 @@ _T = TypeVar("_T")
 _BaseModelT = TypeVar("_BaseModelT", bound="BaseModel")
 
 P = ParamSpec("P")
-
 
 @runtime_checkable
 class _ConfigProtocol(Protocol):
@@ -584,10 +584,7 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
     return value
 
 
-@runtime_checkable
-class CachedDiscriminatorType(Protocol):
-    __discriminator__: DiscriminatorDetails
-
+_DISCRIMINATOR_CACHE: "WeakKeyDictionary[type, DiscriminatorDetails]" = WeakKeyDictionary()
 
 class DiscriminatorDetails:
     field_name: str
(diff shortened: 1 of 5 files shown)


Natim:
Refs https://github.com/openai/openai-python/pull/2705

Natim:
### 🧩 The Root Cause

Starting with **Python 3.14**, `typing.Union` (and also `types.UnionType` for `|` syntax) is now fully immutable — you can no longer attach arbitrary attributes to them (they have no `__dict__`).
Hence this line fails with:

```python
cast(CachedDiscriminatorType, union).__discriminator__ = details
```

With the error:

```pgsql
AttributeError: 'typing.Union' object has no attribute '__discriminator__' and no __dict__ for setting new attributes
```

This behavior worked in `≤ 3.13` because `Union` objects allowed dynamic attributes, but that’s gone.

Natim:
### ✅ The Correct Fix

We can’t store metadata on the `Union` type itself anymore.
Instead, we can keep **a sidecar cache (a `WeakKeyDictionary`)** keyed by the union type.

Here’s how you can safely patch both OpenAI and Anthropic implementations:

#### Step 1 – Create a global cache

Add this near the top of the module (just after CachedDiscriminatorType is defined):

```python
import weakref

# Global cache since typing.Union objects are now immutable (Python 3.14+)
_DISCRIMINATOR_CACHE: "weakref.WeakKeyDictionary[type, DiscriminatorDetails]" = weakref.WeakKeyDictionary()
```

#### Step 2 – Update `_build_discriminated_union_meta`

Replace

```python
if isinstance(union, CachedDiscriminatorType):
    return union.__discriminator__
```

With 

```python
cached = _DISCRIMINATOR_CACHE.get(union)
if cached is not None:
    return cached
```

#### Step 3 - Update the `___discriminator__` set

Replace 

```python
cast(CachedDiscriminatorType, union).__discriminator__ = details
return details
```

With

```python
_DISCRIMINATOR_CACHE[union] = details
return details
```

Natim:
### 🧠 Why This Works

- `WeakKeyDictionary` lets cached entries disappear **when their type objects are garbage-collected**.
- Works across all supported Python versions.
- Preserves existing “caching” semantics without mutating Union objects.

Natim:
The fix landed upstream as part of the openai-python v2.7.2 release!

https://github.com/openai/openai-python/pull/2744/files

Natim:
Fixed with https://github.com/anthropics/anthropic-sdk-python/commit/bd2a137a46cd899e28e81dd8d445dad23440674c

PR #1070, opened 2025-11-13
Accumulate extra pydantic fields from the sample event

evanmiller-anthropic:
Currently undeclared fields are discarded rather than accumulated in `get_final_message`; this adds code to deep-merge these fields instead.


src/anthropic/lib/streaming/_messages.py
```diff
@@ -401,6 +401,31 @@ def build_events(
...
+def _deep_merge_extra_fields(existing: object, new: object) -> object:
+    """Deep merge new data into existing data, mutating containers in place.
+
+    - Dicts: recursively merge keys (mutates existing dict)
+    - Lists: extend existing with new items (mutates existing list)
+    - Other: replace with new value
+    """
+    if isinstance(existing, dict) and isinstance(new, dict):
+        existing_dict = cast("dict[str, object]", existing)
+        new_dict = cast("dict[str, object]", new)
```
karpetrosyan: nit: we can use `is_dict` utility methods, to avoid cast

src/anthropic/lib/streaming/_messages.py
```diff
@@ -401,6 +401,31 @@ def build_events(
...
+        for key, value in new_dict.items():
+            if key in existing_dict:
+                existing_dict[key] = _deep_merge_extra_fields(existing_dict[key], value)
+            else:
+                existing_dict[key] = value
+        return existing_dict  # Return mutated dict
+    elif isinstance(existing, list) and isinstance(new, list):
+        existing_list = cast("list[object]", existing)
+        new_list = cast("list[object]", new)
```
karpetrosyan: nit: same as `is_dict`, we can use utility function `is_list` to avoid typing.cast

tests/lib/streaming/test_extra_fields.py
```diff
@@ -0,0 +1,149 @@
...
+
+from anthropic import Anthropic, AsyncAnthropic
+
+from .helpers import get_response, to_async_iter
+
+base_url = os.environ.get("TEST_API_BASE_URL", "http://127.0.0.1:4010")
+api_key = "my-anthropic-api-key"
+
+sync_client = Anthropic(base_url=base_url, api_key=api_key, _strict_response_validation=True)
+async_client = AsyncAnthropic(base_url=base_url, api_key=api_key, _strict_response_validation=True)
```
karpetrosyan: nit: we should use fixtures instead of global clients

tests/lib/streaming/test_extra_fields.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def assert_extra_fields_accumulated(message: Any) -> None:
```
karpetrosyan: nit: let's have the correct type here instead of `Any`

karpetrosyan (changes requested):

evanmiller-anthropic:
Passes locally with Pydantic v2; I added a short-circuit for Pydantic v1 in the test suite since this is not supported. (It looks like CI uses V1.)

src/anthropic/lib/streaming/_messages.py
```diff
@@ -481,4 +502,19 @@ def accumulate_event(
         if event.usage.server_tool_use is not None:
             current_snapshot.usage.server_tool_use = event.usage.server_tool_use
 
+    # Accumulate any extra fields from the event into the snapshot
```
karpetrosyan, line 505: question: should we just accumulate all the extra fields from events into the message? 
TBH, what I really expect here is to see the same extra fields I would get if I made a non-streaming request. Maybe we shouldn't accumulate all the extra kwargs from events, but only the ones that make sense (which are extra kwargs for the non-streaming case too).

- Does that make sense?
- Do we know which events are responsible for transferring the extra fields we would see in the non-streaming case?

evanmiller-anthropic: I believe non-streaming requests already copy over these extra fields (the code I'm working with broke when we moved from non-streaming to streaming). I'm not sure exactly where that happens, though.

karpetrosyan: Sorry for the confusion! Let me explain what I'm seeing.
So in the non-streaming case, we have extra fields A and B, right? I'd expect the streaming case to have those same fields – no more, no less. But right now it looks like it might be getting more.
It seems like we're accumulating extra fields from all the events into the final message, including stuff from the message-stop event and other events that don't actually hold message data. I don't think that's what we want here – or am I missing something?

Also, could you share the SDK version and the requests you made for both streaming and non-streaming? That would be super helpful for figuring this out!

tests/lib/streaming/fixtures/extra_fields_response.txt
karpetrosyan, line 1: I think we don't use this file anymore, right? If so, let's remove it

tests/lib/streaming/test_extra_fields.py
```diff
@@ -0,0 +1,128 @@
...
+from anthropic._compat import PYDANTIC_V1
+from anthropic.lib.streaming._messages import accumulate_event
+from anthropic.types.message_delta_usage import MessageDeltaUsage
+from anthropic.types.raw_message_delta_event import Delta, RawMessageDeltaEvent
+from anthropic.types.raw_message_start_event import RawMessageStartEvent
+from anthropic.types.raw_content_block_delta_event import RawContentBlockDeltaEvent
+from anthropic.types.raw_content_block_start_event import RawContentBlockStartEvent
+
+
+class TestExtraFieldsAccumulation:
```
karpetrosyan: nit: let's remove this class definition, it just holds a single test

tests/lib/streaming/test_extra_fields.py
```diff
@@ -0,0 +1,128 @@
...
+            type="message_delta",
+            delta=Delta(stop_reason="end_turn", stop_sequence=None),
+            usage=MessageDeltaUsage(output_tokens=3),
+            private_field={"nested": {"values": [6]}},  # type: ignore[call-arg]
+        )
+        snapshot = accumulate_event(event=message_delta, current_snapshot=snapshot)
+
+        # This feature requires Pydantic v2
+        if PYDANTIC_V1:
+            return
```
karpetrosyan: we should use `pytest.mark.skipif` if we want to skip the test for some cases, like [here](https://github.com/evanmiller-anthropic/anthropic-sdk-python/blob/829486182c91de312d80cdc5dd1032d654447e37/tests/test_models.py#L836)

karpetrosyan (changes requested):

src/anthropic/lib/streaming/_messages.py
evanmiller-anthropic: Would it be enough to gate this logic on the event.type?

I will have to dig up the version numbers and test cases later.

karpetrosyan: It'd be great to understand which particular events are responsible for transferring message extra fields. From the snapshots you've created for the test, I see that these are `message_start`, `content_block_delta`, and `message_delta`. We can potentially update `kwargs` only if the event is one of these as you mentioned. It would help us avoid filling up the extra `kwargs` of the final message with unnecessary data.
Also, I noticed that the same `kwargs` could be sent through different event types. For example, in your snapshot, `message_start`, `content_block_delta`, and `message_delta` are all sending the `private_field` extra field. Is that intentional?

karpetrosyan: P.S. I see that `message_delta` can send some fields which have already been sent by `message_start`, such as usage or context management fields, so we need to accumulate at least these two events. But I'm not sure about `content_block_delta`.

PR #1076, opened 2025-11-19
feat: support special types for structured outputs

karpetrosyan:
Fixes typing so it also works with special types like `Literal`, `Union`, etc. Currently, we use `type[T]` and return `T` to unwrap the class type, telling the type checker that we return the instance, not the type. When special types are passed, the type checker fails because special types are not subclasses of type, and there is nothing to unwrap in this case.

[PEP 747](https://peps.python.org/pep-0747/) would solve this problem by providing a way to annotate these types. Until then, this workaround can be used:

```python
from __future__ import annotations

from typing import Union, TypeVar, cast, overload

T = TypeVar("T")


class Omit: ...


omit = Omit()


class SomeModel: ...


# This overload should be listed first to prioritize it over generic overload
# so calls with Omit() resolve to ParsedBetaMessage[None] not ParsedBetaMessage[Omit]
@overload
def parse(output_format: None | Omit = omit) -> None: ...
# This should be before the generic overload, so if the type is a class type, we unwrap it.
# Otherwise, something like Union[str, int] would incorrectly resolve to type[str] | type[int].
@overload
def parse(output_format: type[T]) -> T: ... # for classes
@overload
def parse(output_format: T) -> T: ...  # for special types
def parse(output_format: None | type[T] | T | Omit = omit) -> T:
    return cast(T, None)


just_none = parse(omit) # None
special_type = parse(Union[str, int]) # str | int 
model_type = parse(SomeModel) # SomeModel
```

### DIFF
--- a/src/anthropic/resources/beta/messages/messages.py
+++ b/src/anthropic/resources/beta/messages/messages.py
@@ -4,10 +4,10 @@ from __future__ import annotations
 
 import inspect
 import warnings
-from typing import TYPE_CHECKING, List, Type, Union, Iterable, Optional, cast
+from typing import TYPE_CHECKING, List, Type, Union, Generic, Iterable, Optional, cast
 from functools import partial
 from itertools import chain
-from typing_extensions import Literal, overload
+from typing_extensions import Unpack, Literal, overload
 
 import httpx
 import pydantic
@@ -77,6 +77,22 @@ if TYPE_CHECKING:
     from ...._client import Anthropic, AsyncAnthropic
 
 __all__ = ["Messages", "AsyncMessages"]
+
+
+class BetaMessagesParseParamsWithoutFormat(message_create_params.MessageCreateParamsNonStreaming):
+    output_format: None | Omit = omit  # type: ignore[assignment]
+
+
+class BetaMessagesParseParamsWithFormatType(
+    message_create_params.MessageCreateParamsNonStreaming, Generic[ResponseFormatT]
+):
+    output_format: ResponseFormatT  # type: ignore[assignment]
+
+
+class BetaMessagesParseParamsIntersection(
+    message_create_params.MessageCreateParamsNonStreaming, Generic[ResponseFormatT]
+):
+    output_format: None | type[ResponseFormatT] | ResponseFormatT | Omit = omit  # type: ignore[assignment]
 
 
 class Messages(SyncAPIResource):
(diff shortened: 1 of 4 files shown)


src/anthropic/resources/beta/messages/messages.py
```diff
@@ -2950,14 +3230,105 @@ def tool_runner(
...
+    ) -> BetaAsyncMessageStreamManager[None]: ...
+    @overload
+    def stream(
+        self,
+        *,
+        max_tokens: int,
+        messages: Iterable[BetaMessageParam],
+        model: ModelParam,
+        metadata: BetaMetadataParam | Omit = omit,
+        output_format: Type[ResponseFormatT],
```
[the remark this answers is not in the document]
karpetrosyan: thanks!

RobertCraigie:
Could you check if we could do something similar to https://github.com/python/mypy/issues/19227#issuecomment-2938604664 where we only support special types for Pyright? It'd be ideal if we could avoid using overloads here imo

karpetrosyan:
I tried it, to be honest, and it doesn't work. 
Note that the workaround here is not the overload—the overload we need in any case because even with `TypeForm[T]` -> `T`, passing something like `omit` will not be resolved to `None` but will be resolved to `Omit`. The workaround here is to add one more overload that tries to extract the instance type, e.g., (`type[int] -> int`), and then falls back to just using the same type, e.g., (`int | str -> int | str`).


```python
# This should be before the generic overload, so if the type is a class type, we unwrap it.
# Otherwise, something like Union[str, int] would incorrectly resolve to type[str] | type[int].
@overload
def parse(output_format: type[T]) -> T: ... # for classes
@overload
def parse(output_format: T) -> T: ...  # for special types
```

src/anthropic/resources/beta/messages/messages.py
```diff
@@ -1061,7 +1064,100 @@ def parse(
...
+        top_k: int | Omit = omit,
+        top_p: float | Omit = omit,
+        betas: List[AnthropicBetaParam] | Omit = omit,
+        # Use the following arguments if you need to pass additional parameters to the API that aren't available via kwargs.
+        # The extra values given here take precedence over values defined on the client or passed to this method.
+        extra_headers: Headers | None = None,
+        extra_query: Query | None = None,
+        extra_body: Body | None = None,
+        timeout: float | httpx.Timeout | None | NotGiven = NOT_GIVEN,
+    ) -> ParsedBetaMessage[ResponseFormatT]: ...
```
RobertCraigie, line 1148: I think we need a fallback overload for the "you're passing any supported thing" case like we have for streaming?

karpetrosyan: I think in the streaming case we do it so we can annotate cases when the boolean is not statically known, so we just say it's either a streaming or non-streaming response, not just because we need to have an 'any supported thing' case.

RobertCraigie: yeah do we not need that same thing here? e.g. if you're writing a wrapper function, you would get type errors?

karpetrosyan: > yeah do we not need that same thing here?

I think this is a slightly different case, so no.

> e.g. if you're writing a wrapper function, you would get type errors?

I'm not quite following this. Can you explain the use case?

karpetrosyan:
@RobertCraigie I’ve tried using `kwargs` with unpack, and it seems to work great—at least in my vscode (with pylance).
It’ll now also work with the codegen changes—no need to port new kwargs into the custom function signatures

PR #1080, opened 2025-11-22
feat(bedrock): support AWS_BEARER_TOKEN_BEDROCK env var

### LINKED ISSUE #1079: AnthropicBedrock: AWS_BEARER_TOKEN_BEDROCK env var not supported
Bedrock supports Bearer tokens as well as SigV4 auth. With the AWS SDK this is handled via the `AWS_BEARER_TOKEN_BEDROCK` environment variable. If set, that token is sent as a bearer token in the `Authorization` header instead of doing SigV4 auth

https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html

davidmhewitt:
Fixes #1079

### DIFF
--- a/src/anthropic/lib/bedrock/_auth.py
+++ b/src/anthropic/lib/bedrock/_auth.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+import os
 import httpx
 
 from ..._utils import lru_cache
@@ -42,6 +43,12 @@ def get_auth_headers(
     profile: str | None,
     data: str | None,
 ) -> dict[str, str]:
+    bedrock_bearer = os.getenv("AWS_BEARER_TOKEN_BEDROCK")
+    if bedrock_bearer is not None:
+        headers = headers.copy()
+        headers["Authorization"] = f"Bearer {bedrock_bearer}"
+        return dict(headers)
+
     from botocore.auth import SigV4Auth
     from botocore.awsrequest import AWSRequest
 
--- a/src/anthropic/lib/bedrock/_auth.py
+++ b/src/anthropic/lib/bedrock/_auth.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-import os
 import httpx
 
 from ..._utils import lru_cache
@@ -42,11 +41,11 @@ def get_auth_headers(
     region: str | None,
     profile: str | None,
     data: str | None,
+    aws_bearer_token_bedrock: str | None = None,
 ) -> dict[str, str]:
-    bedrock_bearer = os.getenv("AWS_BEARER_TOKEN_BEDROCK")
-    if bedrock_bearer is not None:
+    if aws_bearer_token_bedrock is not None:
         headers = headers.copy()
-        headers["Authorization"] = f"Bearer {bedrock_bearer}"
+        headers["Authorization"] = f"Bearer {aws_bearer_token_bedrock}"
         return dict(headers)
 
     from botocore.auth import SigV4Auth
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -305,6 +305,7 @@ class AsyncAnthropicBedrock(BaseBedrockClient[httpx.AsyncClient, AsyncStream[Any
         # outlining your use-case to help us decide if it should be
         # part of our public interface in the future.
         _strict_response_validation: bool = False,
+        api_key: str | None = None,
     ) -> None:
         self.aws_secret_key = aws_secret_key
 
@@ -314,6 +315,11 @@ class AsyncAnthropicBedrock(BaseBedrockClient[httpx.AsyncClient, AsyncStream[Any
         self.aws_profile = aws_profile
 
         self.aws_session_token = aws_session_token
+
+        if api_key is None:
+            api_key = os.environ.get("AWS_BEARER_TOKEN_BEDROCK")
+
+        self.api_key = api_key
 
         if base_url is None:
             base_url = os.environ.get("ANTHROPIC_BEDROCK_BASE_URL")
@@ -359,6 +365,7 @@ class AsyncAnthropicBedrock(BaseBedrockClient[httpx.AsyncClient, AsyncStream[Any
             region=self.aws_region or "us-east-1",
             profile=self.aws_profile,
             data=data,
+            aws_bearer_token_bedrock=self.api_key,
         )
         request.headers.update(headers)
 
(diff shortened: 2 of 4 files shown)


src/anthropic/lib/bedrock/_auth.py
```diff
@@ -42,6 +43,12 @@ def get_auth_headers(
     profile: str | None,
     data: str | None,
 ) -> dict[str, str]:
+    bedrock_bearer = os.getenv("AWS_BEARER_TOKEN_BEDROCK")
+    if bedrock_bearer is not None:
```
riywo: I think `if bedrock_bearer:` would be better here because it's only `True` when `bedrock_bearer` is non-empty while this line can be `True` for the empty string.

> In the context of Boolean operations, and also when expressions are used by control flow statements, the following values are interpreted as false: False, None, numeric zero of all types, and empty strings ...
https://docs.python.org/3/reference/expressions.html#boolean-operations

davidmhewitt: An empty string is a valid value for AWS environment variables, and the AWS SDK uses the `AWS_BEARER_TOKEN_BEDROCK` environment variable, even when it's empty.

So this code just matches the AWS SDK behaviour:

https://github.com/boto/botocore/issues/3603

src/anthropic/lib/bedrock/_auth.py
```diff
@@ -42,6 +43,12 @@ def get_auth_headers(
     profile: str | None,
     data: str | None,
 ) -> dict[str, str]:
+    bedrock_bearer = os.getenv("AWS_BEARER_TOKEN_BEDROCK")
```
karpetrosyan: I think adding `api_key` as a client argument and defaulting to that environment variable would be better than only supporting it through the environment variable

tests/lib/test_bedrock.py
```diff
@@ -195,3 +195,32 @@ def test_region_infer_from_specified_profile(
     client = AnthropicBedrock()
 
     assert client.aws_region == next(profile for profile in profiles if profile["name"] == aws_profile)["region"]
+
+
+@pytest.mark.respx()
+def test_bearer_token_env(monkeypatch: t.Any, respx_mock: MockRouter) -> None:
```
karpetrosyan: let's use `pytest.MonkeyPatch` instead of `t.Any`

tests/lib/test_bedrock.py
```diff
@@ -195,3 +195,32 @@ def test_region_infer_from_specified_profile(
...
+    respx_mock.post(re.compile(r"https://bedrock-runtime\.us-east-1\.amazonaws\.com/model/.*/invoke")).mock(
```
karpetrosyan, line 202: can we simplify this mock, maybe by just using the hostname instead of this regex?

tests/lib/test_bedrock.py
```diff
@@ -195,3 +195,32 @@ def test_region_infer_from_specified_profile(
...
+        extra_headers={"Custom-Header": "CustomValue"},
+    )
+    calls = cast("list[MockRequestCall]", respx_mock.calls)
+    assert len(calls) == 1
+    auth_header = calls[0].request.headers.get("Authorization")
+    assert auth_header == "Bearer test-bearer-token"
+
+    # Verify that custom headers are preserved
+    custom_header = calls[0].request.headers.get("Custom-Header")
+    assert custom_header == "CustomValue"
```
karpetrosyan: Do we actually need to test this? If so, maybe adding an `X-` prefix would make it less confusing, as custom headers without that prefix are getting stripped out by proxy servers. I'd basically recommend just removing it and only testing that the Bearer header was passed.

karpetrosyan (changes requested):

src/anthropic/lib/bedrock/_auth.py
davidmhewitt: Sure, I've added it as the last argument to to the client constructor to preserve the existing ordering. Let me know if that makes sense, given the `_strict_response_validation` arg (and associated comment) above it.

tests/lib/test_bedrock.py
davidmhewitt: Thanks, done!

tests/lib/test_bedrock.py
davidmhewitt: This regex matches all of the existing mocks in the test file. 

If we change to just the hostname, the call is no longer intercepted. I haven't used `resp_mockx` before, so there may be a way to simplify this, but should we just leave it as-is for consistency with the other tests?

tests/lib/test_bedrock.py
davidmhewitt: Sure, removed.

karpetrosyan:
LGTM! Note that it also needs ruff format. Could you please run/commit it?

karpetrosyan (dismissed):

src/anthropic/lib/bedrock/_client.py
```diff
@@ -156,6 +156,7 @@ def __init__(
         # outlining your use-case to help us decide if it should be
         # part of our public interface in the future.
         _strict_response_validation: bool = False,
+        api_key: str | None = None,
```
RobertCraigie, line 159: Can you make the same change for the async bedrock client as well?

RobertCraigie (commented):
Do you know why the AWS naming around this functionality is inconsistent? i.e. why is the env var "bearer token" but the actual value is an "api key"?

src/anthropic/lib/bedrock/_client.py
davidmhewitt: Oops, missed that, thank you!

davidmhewitt:
> Do you know why the AWS naming around this functionality is inconsistent? i.e. why is the env var "bearer token" but the actual value is an "api key"?

Not sure!

I would guess they called it API key on the user-facing docs side to be consistent with other LLM providers that use that terminology. Then someone implementing the feature in the SDK decided to name the env var in the way that describes the way the token is transported :man_shrugging: 

I'd be happy to rename all the variables/params to one or the other here if it helps

devstein:
@davidmhewitt thank you for contributing this! I would love to see this merged.

RiccardoRubini:
Thanks a lot @davidmhewitt. This is a very needed feature. Hope it will be merged soon

fritz-astronomer:
Thanks! Would also love to see this merged.

karpetrosyan:
@RobertCraigie Friendly ping on this! We could also rename `api_key` to `aws_bearer_token` to match AWS naming, instead of adapting their naming to match ours

maxa-adrian-blandin:
Thanks! Could it be approved and merged? :)

**EDIT:** for anyone that might stumble on this PR, it's now supported.
See https://github.com/anthropics/anthropic-sdk-python/blob/717f90bf95b0dd3e1fefb2df06c022f16a03ee96/src/anthropic/lib/bedrock/_client.py#L136-L162

localden:
Thank you for the PR - this was already addressed on `main` - see [`src/anthropic/lib/bedrock/_client.py:161`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/bedrock/_client.py#L161). Closing as resolved.

PR #1086, opened 2025-11-26
feat(tool runner): add support for server-side tools

karpetrosyan:


RobertCraigie (commented):
Instead of using `BetaToolUnionParam` should we only type it to allow server side tools? or rather, not allow client side tools?

I'm concerned about a footgun with surprising behaviour where a user passes in a normal client side tool without using our tool helpers, where if the model tries to call it, you'd probably get a weird error.

So another solution would be to make sure the error message in that case is helpful.

karpetrosyan:
> I'm concerned about a footgun with surprising behaviour where a user passes in a normal client side tool without using our tool helpers, where if the model tries to call it, you'd probably get a weird error.

How would a user pass the client side tools without using tool helpers? It is basically only possible with the tool helpers feature.

RobertCraigie:
cause now you can do `tool_runner(..., tools=[{'name': 'get_weather', 'type': 'custom', 'input_schema': ...}])` right?

karpetrosyan:
Ah, I see now. The server tool use blocks have other types, not 'tool_use', so our tool runner doesn't see them. For example, the [web search tool use block](https://github.com/karpetrosyan/anthropic-sdk-python/blob/09666139fc472c308a91e2cfb4c4aef44c35f594/src/anthropic/types/server_tool_use_block.py#L18) will send 'server_tool_use'

RobertCraigie:
sorry no I mean that you could do this
```py
        max_tokens=1024,
        model="claude-3-5-sonnet-latest",
        tools=[{"name": "get_weather", "input_schema": {...}}],
        messages=[{"role": "user", "content": "What is the weather in SF?"}],
    )
```
which form the anthropic API perspective is the exact same as `@beta_tool()`, so it'll try to call that tool and then we'll get a weird error

karpetrosyan:
In that case, the tool runner will send an error message as a tool call result back to Anthropic, and the user will probably see a final message from Anthropic saying that the user-provided tool wasn't successfully called. Would it be sufficient if we add a warning there, so it would be clear to the user that something is wrong?

src/anthropic/resources/beta/messages/messages.py
```diff
@@ -1315,6 +1322,15 @@ def tool_runner(
             **(extra_headers or {}),
         }
 
+        client_side_tools: list[BetaRunnableTool] = []
+        server_side_tools: list[BetaToolUnionParam] = []
```
RobertCraigie: nit: this is a bit of a misnomer
```suggestion
        runnable_tools: list[BetaRunnableTool] = []
        raw_tools: list[BetaToolUnionParam] = []
```
(not a big fan of `raw_tools` so fft come up with something else)

karpetrosyan: changed
I'm ok with `raw_tools`, it's an internal naming so I wont think so much

RobertCraigie (approved):

PR #1090, opened 2025-12-02
feat: allow raw JSON schema to be passed to `messages.stream()`

### LINKED ISSUE #1071: messages.create() takes a JSON schema dict but messages.stream() takes a Pydantic model
Just ran into this slight inconsistency in the API relating to structured content support.

Here's the `messages.create()` idea of `output_format=`:

https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/resources/beta/messages/messages.py#L106

Which uses:

https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/types/beta/beta_json_output_format_param.py#L11-L15

So it accepts a Python dictionary that represents a JSON schema.

But the `messages.stream()` method:

https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/resources/beta/messages/messages.py#L1351

Later does this: https://github.com/anthropics/anthropic-sdk-python/blob/d9aea38e754d55f8f0875fdf19ee44f78ca7b845/src/anthropic/resources/beta/messages/messages.py#L1385-L1386

`TypeAdapter` is a Pydantic concept, so this has the effect of only allowing Pydantic types to be passed to `messages.stream()`.

It's confusing and inconsistent that `messages.create()` requires a JSON schema dictionary but `messages.stream()` instead requires a Pydantic model.

karpetrosyan:
closes #1071


RobertCraigie (approved):
Nice, would be great to include a unit test as well.

karpetrosyan:
added!

RobertCraigie (approved):

PR #1091, opened 2025-12-02
feat(tool runner): add infinite compaction detection

karpetrosyan:


RobertCraigie:
There's a merge conflict FYI

src/anthropic/lib/tools/_beta_runner.py
```diff
@@ -247,8 +249,19 @@ def __run__(self) -> Iterator[RunnerItemT]:
 
             self._iteration_count += 1
 
+            compacted = self._check_and_compact()
+
+            if compacted:
+                consecutive_compactions += 1
+                log.debug(f"Consecutive compactions: {consecutive_compactions}")
+                if consecutive_compactions >= 2:
+                    raise RuntimeError("Infinite compaction detected: two consecutive iterations performed compaction")
```
RobertCraigie: nit: we should include something actionable in the error message, what could the user do to prevent this?

Also minor but
```suggestion
                    raise RuntimeError("Potentially infinite compaction detected: two consecutive iterations performed compaction")
```

karpetrosyan: This happens mostly because the threshold number is very small. If you set it too low, some extra tokens, for example tool definitions won’t be affected by compaction and won’t fit within that threshold.

PR #1121, opened 2026-01-20
fix(structured outputs): avoid including beta header if `output_format` is missing

karpetrosyan:
Third-party clients (Vertex, Bedrock) don’t support structured outputs and its beta header, so when the SDK sets it, the remote server returns `Error code: 400 - {'message': 'invalid beta flag'}.`

This change just avoids always setting that header, so we can still port the `tool_runner` and `stream` helpers, which use `parse` method internally, with minimal changes.

### DIFF
--- a/src/anthropic/resources/beta/messages/messages.py
+++ b/src/anthropic/resources/beta/messages/messages.py
@@ -2789,7 +2789,7 @@ class AsyncMessages(AsyncAPIResource):
             )
         betas = [beta for beta in betas] if is_given(betas) else []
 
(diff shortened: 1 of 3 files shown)


RobertCraigie (approved):
Nice!

karpetrosyan:
@RobertCraigie merge it please if it's ready. I can't merge unfortunately, even the approved ones 😢

RobertCraigie:
Ah sorry I forgot you can't do that yet :(

PR #1110, opened 2026-01-05
feat(tools): support callable class instances and bound methods

majiayu000:
## Summary

Add support for using callable class instances and bound methods as tools in `beta_tool`. This enables a more natural pattern for tools that need access to instance state or context.

### Changes
- Added `_normalize_callable()` helper function that extracts the bound `__call__` method from callable class instances
- Modified `BaseFunctionTool.__init__` to normalize callables before processing
- Added tests for callable class instances and bound methods

### Usage

**Callable class instance:**
```python
class FetchProduct:
    def __init__(self, ctx: dict):
        self.ctx = ctx
    
    def __call__(self, product_id: int) -> str:
        """Fetch a product by ID."""
        return f"Product {product_id} from {self.ctx['session']}"

instance = FetchProduct({"session": "test"})
tool = beta_tool(instance, name="fetch_product")
```

**Bound method:**
```python
class WeatherService:
    def __init__(self, api_key: str):
        self.api_key = api_key
    
    def get_weather(self, location: str) -> str:
        """Get weather for a location."""
        return f"Weather in {location}"

service = WeatherService("secret-key")
tool = beta_tool(service.get_weather, name="weather")
```

Closes #1087

## Test plan
- [x] Added unit tests for callable class instances
- [x] Added unit tests for bound methods
- [x] Verified existing function tests still pass
- [x] Ran `./scripts/lint` - all checks pass
- [x] Ran `./scripts/test tests/lib/tools/test_functions.py` - 23 tests pass

PR #1122, opened 2026-01-20
feat(tools): support callable class instances and bound methods

### LINKED ISSUE #1087: Support class methods for tool helpers
```
    @beta_tool(name='get_user')
    def get_user(id: int) -> str:
        if user_id != id:
        return ...

    return [get_user]

# in router/interactor
user_id = 123
client.beta.tool_runner(
    ...,
    tools=setup_tools(user_id)
)
```

```
@beta_tool(name='get_user')
    if ctx['user_id'] != id:
    return ...

client.beta.tool_runner(
    ...,
    tools=[get_user],
    context={'user_id': 1},
)
```

```
class FetchProduct
        self.ctx = ctx

tools = [FetchProduct(ctx)]
```

fede-kamel:
## Summary

Adds support for callable class instances and bound methods with `beta_tool()`. 

This makes it easier to create tools that need access to instance state or request-scoped context (database sessions, API keys, user context, etc.) without having to wrap everything in closure factories.

### Before

```python
def setup_tools(session):
    @beta_tool
    def fetch_product(id: int) -> str:
        return session.get(Product, id)
    return [fetch_product]

tools = setup_tools(db_session)
```

### After

```python
class FetchProduct:
    def __init__(self, session):
        self.session = session

        """Fetch a product by ID."""
        return self.session.get(Product, id)

tool = beta_tool(FetchProduct(db_session), name="fetch_product")
```

Bound methods also work:

```python
tool = beta_tool(my_service.get_weather, name="weather")
```

## Test plan

- [x] Unit tests for callable class instances
- [x] Unit tests for bound methods  
- [x] Integration tests with tool_runner
- [x] All existing tests pass

Closes #1087

### DIFF
--- a/src/anthropic/lib/tools/_beta_functions.py
+++ b/src/anthropic/lib/tools/_beta_functions.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from abc import ABC, abstractmethod
 from typing import Any, Union, Generic, TypeVar, Callable, Iterable, Coroutine, cast, overload
-from inspect import iscoroutinefunction
+from inspect import ismethod, isfunction, iscoroutinefunction
 from typing_extensions import TypeAlias, override
 
 import pydantic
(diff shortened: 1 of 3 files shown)


karpetrosyan:
duplicate of #1110

PR #1124, opened 2026-01-20
fix(tools): use filtered messages list in async compaction

fede-kamel:
## Summary

Fixes a bug where async `_check_and_compact()` ignores the tool_use filtering logic.

## The Bug

When compaction runs and the last message is an assistant with only `tool_use` blocks, those blocks should be filtered out before sending the summarization request. The sync version does this correctly, but the async version was using `self._params["messages"]` instead of the filtered `messages` variable.

**Without fix:**
```
BadRequestError: messages.2: `tool_use` ids were found without 
`tool_result` blocks immediately after: toolu_test123
```

**With fix:** Compaction succeeds - tool_use is filtered out.

## Code Change

```python
# Sync (line 202) - correct
messages = [*messages, ...]

# Async (line 453) - was buggy  
messages = [*self._params["messages"], ...]  # ignored filtering!

# Async (line 453) - fixed
messages = [*messages, ...]
```

## Test Plan

- [x] Added regression test `test_async_compaction_filters_tool_use`
- [x] Test fails without fix, passes with fix
- [x] All existing tests pass
- [x] Verified with real API call


fede-kamel:
@RobertCraigie @karpetrosyan @dtmeadows — would appreciate a review on this when you have a moment.

This PR has been rebased on the latest `main` (no conflicts). Created a corresponding issue: #1205.

**The bug:** Async `_check_and_compact()` uses `self._params["messages"]` instead of the filtered local `messages` variable, causing `tool_use` blocks to be sent to the API without corresponding `tool_result` blocks. This results in a `BadRequestError` at runtime. The sync version handles this correctly.

**Verified:**
- Test **fails on main** (sends 3 messages, `tool_use` not filtered)
- Test **passes with fix** (sends 2 messages, `tool_use` correctly filtered)
- Standalone reproduction gist (no API key needed): https://gist.github.com/fede-kamel/175f0e99abdb53a380ded3f41236e7dd

The fix is a one-line change. Happy to address any feedback. Thank you.

fede-kamel:
@RobertCraigie @karpetrosyan Friendly follow-up - this is a one-line bug fix for async message compaction. The bug causes `BadRequestError` at runtime when `tool_use` blocks are sent without corresponding `tool_result`. Test + reproduction gist included. Would appreciate a quick review. Thanks!

tests/lib/tools/test_runners.py
```diff
@@ -623,6 +623,80 @@ async def get_weather(location: str, units: Literal["c", "f"]) -> BetaFunctionTo
     ).until_done()
 
 
+@pytest.mark.skipif(PYDANTIC_V1, reason="tool runner not supported with pydantic v1")
+async def test_async_compaction_filters_tool_use(async_client: AsyncAnthropic) -> None:
```
karpetrosyan: LGTM! Let’s move forward without this test. I’m working on some logic to make runner tests way simpler—no mocks, and much shorter

fede-kamel:
@karpetrosyan Done! Removed the test as requested. The PR now contains only the one-line bug fix. Ready for approval when you get a chance. Thanks!

src/anthropic/lib/tools/_beta_runner.py
karpetrosyan, line 111: @fede-kamel do we need to add a cast here? I think it should work with it

tests/lib/tools/test_runners.py
fede-kamel: You're right, the cast was unnecessary - pyright passes without it. Removed it. Thanks for catching that\!

src/anthropic/lib/tools/_beta_runner.py
fede-kamel: You're right, the cast was unnecessary - pyright passes without it. Removed it. Thanks for catching that\!

karpetrosyan (approved):
LGTM! Thanks @fede-kamel

PR #1126, opened 2026-01-21
fix(types): remove caller field from tool_use Param types

### LINKED ISSUE #1112: Extra 'caller' Field in Request Breaks computer-use Between v0.74.1 and v0.75.0
I observe the following with my computer-use application when I upgrade to v0.75.0 of the Anthropic Python SDK:
```
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.1.content.1.tool_use.caller: Extra inputs are not permitted'}, 'request_id': 'req_011CWtR3e2Pnww7S8JnEs2DE'}
```
This error does not occur w/ v0.74.1 of the same SDK. I have attached some Python code which should enable you to reproduce the problem.

[simple_demo.py](https://github.com/user-attachments/files/24477619/simple_demo.py)

Ashutosh0x:
## Summary

Removes the `caller` field from `BetaToolUseBlockParam` and `BetaServerToolUseBlockParam` since it's output-only.

## Problem

Fixes #1112

When users convert API response content blocks back to params for multi-turn conversations, the `caller` field gets included in the request payload and the API rejects it:

```
'messages.1.content.1.tool_use.caller: Extra inputs are not permitted'
```

## Root Cause

The `caller` field was added to both the response types (`BetaToolUseBlock`) and the request param types (`BetaToolUseBlockParam`). However, `caller` is output-only - it's returned by the API but should not be sent in requests.

## Changes

Removed `caller` field from:
- `src/anthropic/types/beta/beta_tool_use_block_param.py`
- `src/anthropic/types/beta/beta_server_tool_use_block_param.py`

Also cleaned up unused imports (`BetaDirectCallerParam`, `BetaServerToolCallerParam`, `TypeAlias`, `Union`).

> **Note:** These files are auto-generated from OpenAPI spec by Stainless. The permanent fix should mark `caller` as `readOnly: true` in the spec so it's excluded from Param types during generation.

## Checklist
- [x] Changes follow existing code patterns
- [x] No breaking changes
- [x] Tested that the fix resolves the 400 error

### DIFF
--- a/issue_1022_comment.md
+++ b/issue_1022_comment.md
@@ -1,0 +1,7 @@
+Hey! This looks like it's already been fixed.
+
+`BetaContentBlock` is now included in `BetaContentBlockParam` - see [beta_content_block_param.py#L49](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/beta/beta_content_block_param.py#L49)
+
+The fix was added in commit `9843a8e` (Nov 24, 2025), which came after this issue was opened.
+
+Can this issue be closed?
--- a/issue_1112_comment.md
+++ b/issue_1112_comment.md
@@ -1,0 +1,33 @@
+Hey! I investigated this and found the root cause.
+
+## Root Cause
+
+The `caller` field was added to `BetaToolUseBlockParam` (the **request** type) in commit `9843a8e` (Nov 24, 2025), but the API only accepts `caller` as an **output-only** field in responses.
+
+When you convert API response content blocks back to params for multi-turn conversations (e.g., passing `message.content` back), the `caller` field gets included in the request payload and the API rejects it with:
+
+```
+'messages.1.content.1.tool_use.caller: Extra inputs are not permitted'
+```
+
+## Where the issue is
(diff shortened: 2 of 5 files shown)


felixfbecker:
Hi @Ashutosh0x, thanks for the PR! I think we need to fix this in the schema, which we codegen this code from.

Ashutosh0x:
Thanks for the quick response @felixfbecker!

I completely agree - the proper fix should be in the OpenAPI schema by marking caller as 
eadOnly: true so it's excluded from Param types during generation.

I'm happy to help! A few questions:
1. Is the OpenAPI spec in a separate repo, or is there a process for requesting schema changes?
2. Should I close this PR and open an issue in the appropriate place for the schema fix?
3. Or would you prefer to handle the schema change internally?

In the meantime, this PR can serve as a reference for what the generated code should look like after the schema fix.

Ashutosh0x:
Closing in favor of upstream schema fix as requested by @felixfbecker.

Issue #984, opened 2025-06-24
Message Batch Processing - undocumented character limit for custom_id parameter


PaulBFB:
creating a request via 

```

import anthropic
from anthropic.types.message_create_params import MessageCreateParamsNonStreaming
from anthropic.types.messages.batch_create_params import Request

client = anthropic.Anthropic()

message_batch = client.messages.batches.create(
    requests=[
        Request(
            custom_id=f"{'a'*65}",
            params=MessageCreateParamsNonStreaming(
                model="claude-opus-4-20250514",
                max_tokens=1024,
                messages=[{
                    "role": "user",
                    "content": "Hello, world",
                }]
            )
        ),
        Request(
            custom_id=f"{'b'*65}",
            params=MessageCreateParamsNonStreaming(
                model="claude-opus-4-20250514",
                max_tokens=1024,
                messages=[{
                    "role": "user",
                    "content": "Hi again, friend",
                }]
            )
        )
    ]
)
```

Will cause 

```
BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'requests.0.custom_id: String should have at most 64 characters'}}
```

I checked the documentation (first with cmd-f and then using claude research ;) ) for the character limit for the custom_id parameter of 64 characters and it's not listed anywhere.

kevinc13:
Hey Paul, thanks for raising the issue! 

We document the 64 character limit in the API reference, but it's unfortunately a bit hidden in the  child attributes [dropdown](https://docs.anthropic.com/en/api/creating-message-batches#body-requests-custom-id).

Let me know if you have any more concerns / feedback

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

nightcityblade:
Unclaiming this — I noticed there is already a linked PR, so I won’t duplicate the work. Sorry for the noise.

PR #1127, opened 2026-01-21
docs: document custom_id 64 character limit

Ashutosh0x:
Documents the undocumented limit of 64 characters for the \custom_id\ field in batch requests.

References: #984

### DIFF
--- a/pr_1112_body.md
+++ b/pr_1112_body.md
@@ -1,0 +1,32 @@
+## Summary
+
+Removes the `caller` field from `BetaToolUseBlockParam` and `BetaServerToolUseBlockParam` since it's output-only.
+
+## Problem
+
+Fixes #1112
+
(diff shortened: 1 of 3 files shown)


localden:
Thank you for the PR - [`src/anthropic/types/beta/messages/batch_create_params.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/beta/messages/batch_create_params.py), [`src/anthropic/types/messages/batch_create_params.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/messages/batch_create_params.py) are auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Closing.

PR #1129, opened 2026-01-21
feat(bedrock): add support for AWS_BEARER_TOKEN_BEDROCK env var

### LINKED ISSUE #1079: AnthropicBedrock: AWS_BEARER_TOKEN_BEDROCK env var not supported

Ashutosh0x:
# feat(bedrock): add support for AWS_BEARER_TOKEN_BEDROCK env var

## Summary

This PR adds support for the `AWS_BEARER_TOKEN_BEDROCK` environment variable in the Bedrock client, enabling bearer token authentication as an alternative to SigV4.

## Problem

Currently, the Bedrock client only supports SigV4 authentication. Some users need to use bearer token authentication for specific deployment scenarios.

## Solution

- Added `get_bearer_token()` function to `_auth.py` to retrieve the token from the environment
- Updated `_prepare_request` in both sync and async Bedrock clients to prioritize bearer token if `AWS_BEARER_TOKEN_BEDROCK` is set, falling back to SigV4 otherwise

## Usage

```python
import os
os.environ["AWS_BEARER_TOKEN_BEDROCK"] = "your-bearer-token"

from anthropic import AnthropicBedrock

client = AnthropicBedrock()
# Client now uses bearer token auth instead of SigV4
```

Fixes #1079

### DIFF
--- a/src/anthropic/lib/bedrock/_auth.py
+++ b/src/anthropic/lib/bedrock/_auth.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from typing import TYPE_CHECKING
 
 import httpx
@@ -8,6 +9,14 @@ from ..._utils import lru_cache
 
 if TYPE_CHECKING:
     import boto3
+
+
+def get_bearer_token() -> str | None:
+    """Get the bearer token from the AWS_BEARER_TOKEN_BEDROCK environment variable.
+
+    Returns the token if set, otherwise None.
+    """
+    return os.environ.get("AWS_BEARER_TOKEN_BEDROCK")
 
 
 @lru_cache(maxsize=512)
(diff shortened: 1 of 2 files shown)


localden:
Thank you for the PR - this was already addressed on `main` - see [`src/anthropic/lib/bedrock/_client.py:161`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/bedrock/_client.py#L161). Closing as resolved.

PR #1130, opened 2026-01-21
fix(client): add 424 status code to retry logic for transient Bedrock errors

Ashutosh0x:
## Summary

Add status code 424 (Failed Dependency) to the automatic retry logic. This status code is frequently returned by Amazon Bedrock for transient server-side issues.

## Changes

- Added status code 424 to the _should_retry method in _base_client.py

## Motivation

Users have reported transient 424 errors from Bedrock that should be automatically retried.


karpetrosyan:
duplicate of #583

PR #1135, opened 2026-01-21
fix(beta): strip extra 'caller' field from tool use blocks

### LINKED ISSUE #1112: Extra 'caller' Field in Request Breaks computer-use Between v0.74.1 and v0.75.0

Ashutosh0x:
Including the 'caller' field in 'tool_use' blocks within input messages causes API errors (especially for computer-use) because the API expects only standard fields like id, name, type, and input. This field was recently added to the SDK models but should not be sent back to the API.

This PR:
1. Removes 'caller' from BetaToolUseBlockParam and BetaServerToolUseBlockParam.
2. Adds __api_exclude__ = {"caller"} to BetaToolUseBlock and BetaServerToolUseBlock response models.
3. Modifies the internal transformer to explicitly strip 'caller' if it's not annotated in a TypedDict, ensuring raw dicts or unfiltered model dumps don't leak it back to the API.

Fixes #1112

### DIFF
--- a/src/anthropic/_utils/_transform.py
+++ b/src/anthropic/_utils/_transform.py
@@ -274,6 +274,9 @@ def _transform_typeddict(
 
         type_ = annotations.get(key)
         if type_ is None:
+            if key == "caller":
+                continue
+
             # we do not have a type annotation for this field, leave it as is
             result[key] = value
         else:
@@ -440,6 +443,9 @@ async def _async_transform_typeddict(
 
         type_ = annotations.get(key)
         if type_ is None:
+            if key == "caller":
+                continue
+
             # we do not have a type annotation for this field, leave it as is
             result[key] = value
         else:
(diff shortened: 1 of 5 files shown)


Ashutosh0x:
Hi @RobertCraigie, I've implemented a fix for Issue #1112 by stripping the extra `caller` field from tool use blocks in API requests. This ensures that even when model instances or raw dictionaries (from previous responses) are passed back to the SDK, the field is correctly omitted, preventing API errors. Please let me know if you have any feedback!

Ashutosh0x:
Closing in favor of upstream schema fix as requested by @felixfbecker.

PR #1138, opened 2026-01-22
fix(tools): filter self and cls parameters from @beta_tool schema generation

Ashutosh0x:
The @beta_tool decorator currently includes \self\ and \cls\ parameters in the generated JSON schema when used as a class method decorator. This fix filters them out to prevent API errors. Fixes #1128


karpetrosyan:
You have two more duplicates of this exact PR, #1128, #1154 and this. Please stop spamming

Issue #1017, opened 2025-08-18
[Feature] Better thinking usage information


jplu:
### Describe the feature

It would be very useful to have more details on the usage if the thinking is enabled. For now, we have no clue how many tokens have been used in the thinking part of the answer.

### Use Case

This feature is useful for us in order to measure how much of a budget goes to the thinking part in order to better estimate the ratio cost / thinking budget.

### Proposed Solution

Like the OpenAI and Gemini APIs, having a `thinking_tokens` attribute in the `usage` object of the response body that will correspond to the number of token used for generating the "thinking" part of the response.

### SDK version used

0.64.0

### Environment details (OS name and version, etc.)

Ubuntu 25.04

bharathc346:
+1

dtmeadows-ant:
Hey @jplu — sorry for the long delay here. This shipped in v0.105.0: the response `usage` object now carries `output_tokens_details.thinking_tokens` with the thinking portion of the output ([release notes](https://github.com/anthropics/anthropic-sdk-python/releases/tag/v0.105.0)).

```python
response = client.messages.create(..., thinking={"type": "enabled", "budget_tokens": 1024})
print(response.usage.output_tokens_details.thinking_tokens)
```

PR #1143, opened 2026-01-22
feat: add thinking_tokens to usage models and streaming accumulation

Ashutosh0x:
This PR adds support for granular reporting of thinking tokens, addressing Feature Request #1017. Changes: Added thinking_tokens field to Usage models and updated accumulation logic.


karpetrosyan:
Thanks for the contribution—closing this since it’s an API-related improvement and should be done on the codegen side

Ashutosh0x:
Thanks for the clarification! Understood that this should be handled on the codegen side. Appreciate the review 🙌

Issue #1001, opened 2025-07-24
Error 500/Internal server error when using tool with draft-07 JSON schema version since 2025-07-24, ~10AM GMT+2


gerkensm:
When including
```
"$schema": "https://json-schema.org/draft-07/schema#,
```

in a tool's JSON schema in an API call to any model, in both streaming and non-streaming mode, since today, all of a sudden, the API responds with a generic 500 Internal Server Error. 
The request shows as fulfilled in the usage logs on the Anthropic Console (This might be important for debugging this as the actual model appears to answer the request just fine - it might be some tooling, analytics, proxying around that?).

When using 
```
“$schema”: “https://json-schema.org/draft/2020-12/schema#”,
```
**or leaving `$schema` out completely**, the request is answered correctly.

This bug likely also affects MCP Server usage. This behaviour is new and happened since this morning ~10am CEST and we never encountered this before. The documentation just says to use a JSON schema, not that it has to be a specific JSON schema version.

Requests with this field set used to work for months prior to today.

This is on the latest python SDK (anthropic 0.59.0).

### 1. Code that FAILS (Returns HTTP 500) - and worked before July 24th, 2025

This code uses the draft-07 schema and consistently fails.
```python
try:
    response = client.messages.create(
        model="claude-3-5-sonnet-20240620",
        max_tokens=1024,
        messages=[{"role": "user", "content": "Please use the CategorizeTool."}],
        tools=[
            {
                "name": "CategorizeTool",
                "description": "Categorizes the user's choice.",
                "input_schema": {
                    "type": "object",
                    # This specific schema version causes a 500 error.
                    "$schema": "https://json-schema.org/draft-07/schema#",
                    "properties": {
                        "category": {
                            "type": "string",
                            "enum": ["CHOICE_A", "CHOICE_B", "CHOICE_C"],
                        }
                    },
                    "required": ["category"],
                },
            }
        ],
        tool_choice={"type": "tool", "name": "CategorizeTool"},
    )
    print(response)
except Exception as e:
    print(f"Request failed: {e}")
```

**Expected Result**: A successful API response with the model using the CategorizeTool.
**Actual Result**: `Request failed: Error code: 500 - {'type': 'error', 'error': {'type': 'api_error', 'message': 'Internal server error'}}`

### 2. Code that SUCCEEDS

The only change is updating the $schema to the 2020-12 version.
```python
try:
    response = client.messages.create(
        model="claude-3-5-sonnet-20240620",
[...]

PR #1144, opened 2026-01-22
fix: strip schema field from tool definitions to prevent 500 errors

Ashutosh0x:
This PR addresses Issue #1001 where including the schema field in tool input_schema caused 500 Internal Server Errors from the API. Changes: Implemented _strip_tools helper and applied it to Messages methods in standard and beta resources.

### DIFF
--- a/src/anthropic/resources/beta/messages/messages.py
+++ b/src/anthropic/resources/beta/messages/messages.py
@@ -78,6 +78,25 @@ if TYPE_CHECKING:
 __all__ = ["Messages", "AsyncMessages"]
 
 
+def _strip_tools(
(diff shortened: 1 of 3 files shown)


localden:
Thank you for the PR - [`src/anthropic/resources/beta/messages/messages.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/resources/beta/messages/messages.py), [`src/anthropic/resources/messages/messages.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/resources/messages/messages.py) are auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Closing.

PR #1150, opened 2026-01-24
fix(tool_runner): propagate container_id for programmatic tool calling

stephaniepang97:
## Problem

When using programmatic tool calling (PTC) with `tool_runner`, the API returns a `container_id` in the response that must be passed in subsequent requests. Without this, the API returns:

```
container_id is required when there are pending tool uses generated by code execution with tools.
```

## Solution

Update both sync and async tool runners to extract `message.container.id` from responses and set it in `self._params["container"]` for subsequent API calls.

## Changes

- `BaseSyncToolRunner.__run__()`: Added container propagation after getting message
- `BaseAsyncToolRunner.__run__()`: Added container propagation after getting message (async version)

## Testing

This enables using `tool_runner` with programmatic tool calling where client-side tools are called from within code execution. The container lifecycle is now properly managed across requests.

## Example

```python
    model="claude-sonnet-4-5",
    tools=[
        {"type": "code_execution_20250825", "name": "code_execution"},
        {
            "name": "my_tool",
            "description": "...",
            "input_schema": {...},
            "allowed_callers": ["code_execution_20250825"]
        }
    ],
    messages=[...],
    betas=["advanced-tool-use-2025-11-20"],
)

for message in runner:
    # Now works! Container ID is automatically propagated between requests
    print(message)
```

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -246,6 +246,10 @@ class BaseSyncToolRunner(BaseToolRunner[BetaRunnableTool, ResponseFormatT], Gene
                 message = self._get_last_message()
                 assert message is not None
 
+                # Update container from response for programmatic tool calling support
+                if message.container is not None:
+                    self._params["container"] = message.container.id
+
             self._iteration_count += 1
 
             # If the compaction was performed, skip tool call generation this iteration
@@ -496,6 +500,10 @@ class BaseAsyncToolRunner(
                 yield item
                 message = await self._get_last_message()
                 assert message is not None
+
+                # Update container from response for programmatic tool calling support
+                if message.container is not None:
+                    self._params["container"] = message.container.id
 
             self._iteration_count += 1
 


dtmeadows (commented):
This looks good to me and the reasoning makes sense. @karpetrosyan can I leave the final review to you? If it's helpful here's the specific place in the docs about having to pass this: https://platform.claude.com/docs/en/agents-and-tools/tool-use/programmatic-tool-calling#step-3-provide-tool-result

karpetrosyan:
LGTM!
nit: I think it's better to commit as a feature, like `feat(tool runner): add support for programmatic tool calling`

karpetrosyan (approved):

src/anthropic/lib/tools/_beta_runner.py
```diff
@@ -246,6 +246,10 @@ def __run__(self) -> Iterator[RunnerItemT]:
                 message = self._get_last_message()
                 assert message is not None
 
+                # Update container from response for programmatic tool calling support
+                if message.container is not None:
```
karpetrosyan, line 250: @stephaniepang97 Can we, instead of using the `container.id` from the last API response, rely on the last assistant message (`runner._get_last_assistant_message_content`)? The user might change the state during the loop (e.g., in a for block via `set_messages_params`), so we should take the ID from the last message in the current state, not from the last API response

src/anthropic/lib/tools/_beta_runner.py
```diff
@@ -497,6 +501,10 @@ async def __run__(self) -> AsyncIterator[RunnerItemT]:
                 message = await self._get_last_message()
                 assert message is not None
 
+                # Update container from response for programmatic tool calling support
+                if message.container is not None:
```
karpetrosyan, line 505: Can we add a simple test for this case to make sure we’re sending container.id back? We can use a snapshot with a real example and record it, like we do in the other tests.

karpetrosyan:
closing as this was resolved [here](https://github.com/anthropics/anthropic-sdk-python/commit/3ae7ff6ff7af8a881706ae8068b1040a23c96fbd)

PR #1151, opened 2026-01-24
fix(compaction): also filter out server_tool_use blocks before compaction

stephaniepang97:
## Problem

When compaction is triggered while the last assistant message contains `server_tool_use` blocks (for server-side tools like `web_fetch` or `web_search`), the compaction API call fails with:

```
messages.X: `web_fetch` tool use with id `srvtoolu_XXX` was found without a corresponding `web_fetch_tool_result` block
```

## Root Cause

The compaction code filters out `tool_use` blocks from the last assistant message before making the compaction request (since they require corresponding `tool_result` blocks). However, it wasn't filtering out `server_tool_use` blocks, which also require corresponding result blocks (`web_fetch_tool_result`, `web_search_tool_result`, etc.).

## Solution

Changed the filter from:
```python
block.get("type") != "tool_use"
```

to:
```python
block.get("type") not in ("tool_use", "server_tool_use")
```

## Changes

- Updated `BaseSyncToolRunner._check_and_compact()` to filter out both `tool_use` and `server_tool_use`
- Updated `BaseAsyncToolRunner._check_and_compact()` with the same fix

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -186,12 +186,12 @@ class BaseSyncToolRunner(BaseToolRunner[BetaRunnableTool, ResponseFormatT], Gene
         messages = list(self._params["messages"])
 
         if messages[-1]["role"] == "assistant":
-            # Remove tool_use blocks from the last message to avoid 400 error
-            # (tool_use requires tool_result, which we don't have yet)
+            # Remove tool_use and server_tool_use blocks from the last message to avoid 400 error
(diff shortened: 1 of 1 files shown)


stephaniepang97:
@kevinc13 do we still need this fix?

tomer-ant:
Thanks, but this patches the client-side compaction path (compaction_control in the tool runner), which was deprecated in https://github.com/anthropics/anthropic-sdk-python/commit/f1f1a3f773a7fe7dd5198fd96d83c4399f8f0ca0 in favor of server-side compaction via edits=[{"type": "compact_20260112"}]. We're not taking further changes to the deprecated helper, so I'm closing this.

PR #1154, opened 2026-01-26
feat(beta): support class methods and instance methods for tool helpers

### LINKED ISSUE #1087: Support class methods for tool helpers

Ashutosh0x:
Fixes #1087. Adds comprehensive support for class methods, instance methods, and static methods when using @beta_tool decorators. 

Key changes:
- Implemented the descriptor protocol (__get__) on tool classes to handle automatic binding.
- Updated JSON schema generation to exclude 'self' and 'cls' parameters.
- Updated decorators to correctly unwrap and re-wrap classmethod/staticmethod objects.
- Changed tool objects to be directly callable (removing @property on __call__) to fix method binding issues.

### DIFF
--- a/src/anthropic/lib/tools/_beta_functions.py
+++ b/src/anthropic/lib/tools/_beta_functions.py
@@ -87,6 +87,8 @@ class BaseFunctionTool(Generic[CallableT]):
         self._func_with_validate = pydantic.validate_call(func)
         self.name = name or func.__name__
         self._defer_loading = defer_loading
+        self._is_classmethod = False
+        self._is_staticmethod = False
 
         self.description = description or self._get_description_from_docstring()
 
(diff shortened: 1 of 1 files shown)


karpetrosyan:
@Ashutosh0x There’s also PR #1128, which you created, and it solves the same issue. Was that intentional?

PR #1158, opened 2026-02-03
fix(structured outputs): send structured output beta header when format is omitted

karpetrosyan:
This reverts commit 062077e50d182719637403576f59761999b3b2f5.

### DIFF
--- a/src/anthropic/resources/beta/messages/messages.py
+++ b/src/anthropic/resources/beta/messages/messages.py
@@ -1118,7 +1118,7 @@ class Messages(SyncAPIResource):
 
         betas = [beta for beta in betas] if is_given(betas) else []
 
-        if "structured-outputs-2025-12-15" not in betas and is_given(output_format):
+        if "structured-outputs-2025-11-13" not in betas:
             # Ensure structured outputs beta is included for parse method
             betas.append("structured-outputs-2025-12-15")
 
@@ -2813,7 +2813,7 @@ class AsyncMessages(AsyncAPIResource):
             )
         betas = [beta for beta in betas] if is_given(betas) else []
 
-        if "structured-outputs-2025-12-15" not in betas and is_given(output_format):
+        if "structured-outputs-2025-11-13" not in betas:
             # Ensure structured outputs beta is included for parse method
             betas.append("structured-outputs-2025-12-15")
 
--- a/src/anthropic/resources/beta/messages/messages.py
+++ b/src/anthropic/resources/beta/messages/messages.py
@@ -1118,7 +1118,7 @@ class Messages(SyncAPIResource):
 
         betas = [beta for beta in betas] if is_given(betas) else []
 
-        if "structured-outputs-2025-11-13" not in betas:
+        if "structured-outputs-2025-12-15" not in betas:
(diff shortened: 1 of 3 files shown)


karpetrosyan:
That change made the normal `tool_runner` and `parse` usage fail because our codegen sends an empty string as a header value when the header value is an empty array, which is invalid for Anthropic.

Now, when a user calls `tool_runner` without providing `output_format`, they’ll get this:

```
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'Unexpected value(s) `` for the `anthropic-beta` header. Please consult our documentation at docs.claude.com or try again without the header.'}, 'request_id': 'req_011CXmHfwd5GrhJkdjGMDod6'}
```

Simple repro script:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "anthropic==0.77.0",
#     "rich",
# ]
# ///

import json
from typing_extensions import Literal

import rich

from anthropic import Anthropic, beta_tool

client = Anthropic()


@beta_tool
def get_weather(location: str, units: Literal["c", "f"]) -> str:
    """Lookup the weather for a given city in either celsius or fahrenheit

    Args:
        location: The city and state, e.g. San Francisco, CA
        units: Unit for the output, either 'c' for celsius or 'f' for fahrenheit
    Returns:
        A dictionary containing the location, temperature, and weather condition.
    """
    # Simulate a weather API call
    print(f"Fetching weather for {location} in {units}")

    # Here you would typically make an API call to a weather service
    # For demonstration, we return a mock response
    if units == "c":
        return json.dumps(
            {
                "location": location,
                "temperature": "20°C",
                "condition": "Sunny",
            }
        )
    else:
        return json.dumps(
            {
                "location": location,
                "temperature": "68°F",
                "condition": "Sunny",
            }
        )


def main() -> None:
        max_tokens=1024,
        model="claude-sonnet-4-5",
        tools=[get_weather],
        messages=[{"role": "user", "content": "What is the weather in SF?"}],
    )
    for message in runner:
        rich.print(message)


main()
```

RobertCraigie:
is it feasible to do the more correct thing and not send an empty string? or is that much more involved

karpetrosyan:
I’m also preparing a more accurate fix [in the codegen,](https://github.com/stainless-api/stainless/pull/15569) though I’m not sure what we should do: 
- should we change the default behavior and avoid sending an empty header value for array headers
- or introduce a flag so customers can configure the behavior? 

Just changing it could probably break some code, so the second option seems safer

karpetrosyan:
I don’t want to just change the codegen behavior to stop sending an empty header value in this case, because empty values are valid in HTTP and might be useful in some cases

RobertCraigie:
fwiw I think it'd be pretty safe to just not send an empty header by default, that seems like more sane behaviour to me.

RobertCraigie:
sounds like this is a bad break though so we can just merge this for now

RobertCraigie (approved):

RobertCraigie:
can you also add a test for this case as well so we can't regress in the future?

PR #1165, opened 2026-02-06
fix: handle None by_alias in model_dump for FastAPI contexts

### LINKED ISSUE #1160: TypeError: argument 'by_alias': 'NoneType' object cannot be converted to 'PyBool' when using SDK in FastAPI context
Environment:
  - anthropic SDK: 0.77.1
  - pydantic: 2.9.2
  - pydantic-core: 2.23.4
  - Python: 3.12
  - FastAPI + uvicorn

Description:

To Reproduce:

Key Finding:
  - Direct CLI test: ✅ Works

Traceback:
      return model.model_dump(

Workaround:

thecaptain789:
When calling `model_dump` with `by_alias=None` (the default), the pydantic-core serializer in certain async contexts (like FastAPI) can fail with `'NoneType' cannot be converted to 'PyBool'`.

This applies the same pattern already used in `_models.py` where `by_alias` is explicitly converted:
```python
by_alias=by_alias if by_alias is not None else False
```

Fixes #1160

### DIFF
--- a/src/anthropic/_compat.py
+++ b/src/anthropic/_compat.py
@@ -155,7 +155,7 @@ def model_dump(
             exclude_defaults=exclude_defaults,
             # warnings are not supported in Pydantic v1
             warnings=True if PYDANTIC_V1 else warnings,
-            by_alias=by_alias,
+            by_alias=by_alias if by_alias is not None else False,
         )
     return cast(
         "dict[str, Any]",


gdellolio-mariner:
when can we merge this? i'm also having this issue while using FastAPI

gdellolio-mariner (approved):

dgellow:
cc @RobertCraigie

localden:
Thank you for the PR - this was already addressed on `main` - see [`src/anthropic/_compat.py:155`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/_compat.py#L155). Closing as resolved.

PR #1167, opened 2026-02-07
fix(models): correctly unwrap Annotated types for RawMessageStreamEvent deserialization

### LINKED ISSUE #941: content_block_delta event not deserialized correctly during streaming

Event -
```
{
  delta: {
    text: "something something",
    type: "text_delta"
  },
    index: 0,
    type: "content_block_delta"
}
```

Error 
```
```

Ashutosh0x:
This PR fixes a bug where construct_type failed to correctly unwrap Annotated types when they were wrapped in TypeAlias (common in newer Python versions/setups) or when metadata was explicitly provided. This ensures that discriminated unions like RawMessageStreamEvent are correctly processed, preventing intermittent Unexpected event runtime type errors during streaming (fixes #941).

### DIFF
--- a/src/anthropic/_compat.py
+++ b/src/anthropic/_compat.py
@@ -155,7 +155,7 @@ def model_dump(
             exclude_defaults=exclude_defaults,
             # warnings are not supported in Pydantic v1
             warnings=True if PYDANTIC_V1 else warnings,
-            by_alias=by_alias,
+            by_alias=by_alias if by_alias is not None else False,
         )
     return cast(
         "dict[str, Any]",
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -503,29 +503,31 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
 
     # store a reference to the original type we were given before we extract any inner
     # types so that we can properly resolve forward references in `TypeAliasType` annotations
-    original_type = None
+    original_type = type_
 
+    # unwrap `Annotated[T, ...]` -> `T` and `TypeAliasType` -> `T.__value__`
+    # while collecting metadata from `Annotated`
+    #
     # we allow `object` as the input type because otherwise, passing things like
     # `Literal['value']` will be reported as a type error by type checkers
     type_ = cast("type[object]", type_)
-    if is_type_alias_type(type_):
-        original_type = type_  # type: ignore[unreachable]
-        type_ = type_.__value__  # type: ignore[unreachable]
 
-    # unwrap `Annotated[T, ...]` -> `T`
-    if metadata is not None and len(metadata) > 0:
-        meta: tuple[Any, ...] = tuple(metadata)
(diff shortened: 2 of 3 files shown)


Ashutosh0x:
I've verified the fix locally with reproduction scripts specifically targeting the TypeAliasType + Annotated nesting scenario. This should resolve the reported intermittent issues with RawMessageStreamEvent deserialization.

felixfbecker:
Thank you for your contribution! Would it be possible to add a test for this?

Ashutosh0x:
@felixfbecker I've added comprehensive regression tests to `tests/test_models.py` covering several nested `TypeAliasType` and `Annotated` scenarios. These verify that the fix correctly unwraps the types and collects the necessary metadata for discriminated unions like `RawMessageStreamEvent`. I also fixed a minor regression in the `model_dump` shim that was causing issues on older Python versions (and Pydantic v1). Ready for another look! 🚀

Ashutosh0x:
@felixfbecker I've refined the fix to address the failure observed in CI. The issue was that passing a `TypeAliasType` directly to Pydantic's `validate_type` (which uses `TypeAdapter`) was inconsistent across Python versions/Pydantic environments, sometimes returning a raw `dict` instead of co-ercing to a model instance.

I've updated `construct_type` to:
1.  **Fully unwrap** the type before passing it to `validate_type`. This ensures Pydantic sees a standard `Union` (or model class) and correctly identifies if the data is structurally valid.
2.  **Collect all metadata** from all `Annotated` layers, ensuring discriminators are never missed.
3.  **Fall back reliably** to the manual construction logic if `validate_type` returns a `dict` for a union that has a discriminator.

This improved approach is more robust and passes correctly in environments where `TypeAliasType` might be handled differently by Pydantic.

Ashutosh0x:
## CI Passing ✅

All CI checks are now passing after the latest commits. Here's a summary of the changes made:

### Changes Summary

**`src/anthropic/_models.py`:**
- Refactored `construct_type` to use a local variable `_type` for the unwrapped type, ensuring `issubclass()` is only called on actual classes
- Removed unnecessary `cast(Any, ...)` calls that pyright flagged as redundant
- Added `# pyright: ignore[reportUnknownVariableType]` comments on lines where dynamic method calls on `Any` types return unknown types (lines 531, 579, 585)

**`tests/test_models.py`:**
- Moved all `TypeAliasType` definitions to module level to satisfy pyright's requirement that type aliases be assigned to variables with matching names
- Added regression tests for nested `TypeAliasType` and `Annotated` unwrapping

### Test Coverage
All 51 tests pass, including the three new regression tests:
- `test_type_alias_annotated_unwrapping` - TypeAliasType of Annotated Union
- `test_annotated_type_alias_unwrapping` - Annotated of TypeAliasType of Union
- `test_deeply_nested_unwrapping` - Deeply nested TypeAliasType and Annotated

Ready for review!

PR #1171, opened 2026-02-08
Fix issue #1160: Handle None value for by_alias in model_dump

### LINKED ISSUE #1160: TypeError: argument 'by_alias': 'NoneType' object cannot be converted to 'PyBool' when using SDK in FastAPI context

veeceey:
## Summary

Fixes TypeError when by_alias=None is passed to model_dump() in Pydantic v2 context.

When the Anthropic SDK is used in a FastAPI application with Pydantic v2, the model_dump() function in _compat.py was passing by_alias=None directly to pydantic's model_dump(), which requires a boolean value. This caused a TypeError in Pydantic-Core.

## Changes

- Updated model_dump() to default by_alias to False when None is passed
- Added test cases to verify the fix works with by_alias=None, True, and False

## Related Issues

Fixes #1160

### DIFF
--- a/src/anthropic/_compat.py
+++ b/src/anthropic/_compat.py
@@ -155,7 +155,7 @@ def model_dump(
             exclude_defaults=exclude_defaults,
             # warnings are not supported in Pydantic v1
             warnings=True if PYDANTIC_V1 else warnings,
-            by_alias=by_alias,
+            by_alias=by_alias if by_alias is not None else False,
         )
     return cast(
         "dict[str, Any]",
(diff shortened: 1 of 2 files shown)


veeceey:
Closing as duplicate of #1168 which was opened earlier with the same fix.

webshark:
Has this been released yet as a fix in the main ?

PR #1172, opened 2026-02-08
Fix tool runner exiting early with server_tool_use blocks

### LINKED ISSUE #1170: Tool runner exits early when response contains only server tool use blocks (e.g. web_search, web_fetch) and stop reason: pause_turn
## Description

## Root cause

```python
if not tool_use_blocks:
    return None
```

```python
if response is None:
    return
```

## Reproduction

```python
import anthropic

@anthropic.beta_tool
        f.write(title + "\n")

    max_tokens=1024,
    tools=[
        save_recipe,
    ],
)

final = runner.until_done()
```

## Expected behavior

## Actual behavior

## SDK version

veeceey:
## Summary
Tool runner exits prematurely when the API responds with `server_tool_use` blocks (like web_search, web_fetch) and `stop_reason: pause_turn`, causing incomplete message returns.

## Problem
The `generate_tool_call_response()` method only checked for `tool_use` blocks, ignoring `server_tool_use` blocks. When only server-side tools were present, the method returned `None`, causing the loop to exit incorrectly.

## Solution
- Detect both `server_tool_use` and `tool_use` blocks
- When server tools are present with pause_turn, return an empty tool result list to continue the loop
- Only append messages when there's actual content
- Apply fix to both sync and async runners

## Fixes
Fixes #1170

## Testing
The fix ensures:
- Runner continues when receiving pause_turn with server tools
- Runner still exits correctly when no tools are present
- Both sync and async versions behave consistently

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -255,8 +255,13 @@ class BaseSyncToolRunner(BaseToolRunner[BetaRunnableTool, ResponseFormatT], Gene
                     log.debug("Tool call was not requested, exiting from tool runner loop.")
                     return
 
-                if not self._messages_modified:
-                    self.append_messages(message, response)
+                # Only append message if there's actual content to add
+                if response is not None and response.get("content"):
+                    if not self._messages_modified:
+                        self.append_messages(message, response)
+                elif response is not None:
+                    # Empty content means we have server tools in pause_turn state, just continue
+                    log.debug("Server tools detected with pause_turn, continuing the loop.")
 
             self._messages_modified = False
             self._cached_tool_call_response = None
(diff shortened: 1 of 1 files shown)


src/anthropic/lib/tools/_beta_runner.py
```diff
@@ -255,8 +255,13 @@ def __run__(self) -> Iterator[RunnerItemT]:
                     log.debug("Tool call was not requested, exiting from tool runner loop.")
                     return
 
-                if not self._messages_modified:
-                    self.append_messages(message, response)
+                # Only append message if there's actual content to add
+                if response is not None and response.get("content"):
```
karpetrosyan: Do we need `response is not None` check here? We do it above and returning if it's `None`

veeceey: Good catch @karpetrosyan\! You're right, the `response is not None` check is redundant since we already return early if it's `None` on the line above. Removed it in both the sync and async `__run__` methods. Thanks for the review\!

veeceey:
Hi @karpetrosyan, thanks for the review\! I've pushed a commit removing the redundant `response is not None` checks in both the sync and async `__run__` methods as you suggested. Ready for another look when you get a chance\!

karpetrosyan:
Thanks for the contribution. I think this can be solved differently—I’ve opened #1174, which resolves it, and I’ve added you as a co-author on the commit. Closing this one.

PR #1174, opened 2026-02-10
fix(tool runner): don't exit early on `pause_turn`

### LINKED ISSUE #1170: Tool runner exits early when response contains only server tool use blocks (e.g. web_search, web_fetch) and stop reason: pause_turn

karpetrosyan:
fixes #1170 

## Problem
The tool runner was exiting early when encountering a `pause_turn` stop reason, preventing it from continuing the conversation loop.

`pause_turn` means the API paused some operation and wants the client to send back the message to continue it, which is why we should send it back instead of just exiting the loop.


felixfbecker (dismissed):
how should this interact with the `max_iterations` param? I think the tool runner accepts two (one for the client-side iteration limit, one for the server-side limit). The server one would in effect only control the "chunking" of getting server tool responses.

This behavior should probably be documented

PR #1200, opened 2026-02-24
fix(streaming): preserve BetaCompactionBlock type during streaming accumulation

### LINKED ISSUE #1175: [Opus 4.6] BetaCompactionBlock deserialized as ParsedBetaTextBlock during streaming accumulation
SDK Version [anthropic==0.79.0] (Python 3.12) via vertex

### Description
When using the beta streaming API with **compaction=True**, the **BetaCompactionBlock** in the final **ParsedBetaMessage.content** is incorrectly typed as a **ParsedBetaTextBlock** instead of **BetaCompactionBlock**.

### Reproduction
```
kwargs["betas"]=["compact-2026-01-12"]
kwargs["context_management"]={
    "edits": [{
        "type": "compact_20260112",
        "trigger": {"type": "input_tokens", "value": 50000},
        "pause_after_compaction": pause_after_compaction
    }]}
```
```
client = AsyncAnthropicVertex(
    project_id=project_id,
    region=region,
    credentials=credentials  # <-- Fix: credentials passées directement
)
async with client.beta.messages.stream(**kwargs) as stream:
    message = await stream.get_final_message()
    
    for block in message.content:
        print(type(block), block.type)
        # Expected: <class 'BetaCompactionBlock'> compaction
        # Actual:   <class 'ParsedBetaTextBlock'> compaction
```
### Observed behavior
When a compaction event occurs, the streaming events (content_block_start, content_block_delta and content_block_stop for "compaction" are correctly typed but the final message contains:

`ParsedBetaTextBlock(citations=None, text=None, type='compaction', parsed_output=None, content=None)`

- type(block) is ParsedBetaTextBlock, not BetaCompactionBlock
- The object has parasitic fields from ParsedBetaTextBlock (text, citations, parsed_output) all set to None
- block.type is correctly "compaction"
- Calling model_dump() / model_dump_json() on the parent message triggers Pydantic serialization warnings for every variant in the union

### Expected behavior
message.content[i] should be an instance of BetaCompactionBlock when block.type == "compaction".

bledden:
## Summary

When using the beta streaming API with compaction enabled, the final message from `get_final_message()` incorrectly types `BetaCompactionBlock` content blocks as `ParsedBetaTextBlock`. The block has `type='compaction'` but `isinstance(block, BetaCompactionBlock)` returns False, and parasitic fields like `text`, `citations`, and `parsed_output` appear as None.

The issue is in the streaming accumulator's `content_block_start` handling. It converts every content block to a dict and reconstructs it through the `ParsedBetaContentBlock` discriminated union via `construct_type()`. On some Python versions (reported on 3.12), the generic `ParsedBetaTextBlock[ResponseFormatT]` variant in that union breaks Pydantic's discriminator resolution, causing all blocks to fall through to the first union variant (`ParsedBetaTextBlock`).

The fix: only text blocks need reconstruction through `construct_type` (they're the only type that needs wrapping in `ParsedBetaTextBlock` to add the `parsed_output` field). Non-text blocks like `BetaCompactionBlock`, `BetaToolUseBlock`, `BetaThinkingBlock`, etc. are already the correct type from the streaming event and can be appended directly.

Added a test fixture simulating a compaction streaming response and two tests (sync + async) that verify `BetaCompactionBlock` instances are preserved in the final message. All 14 beta streaming tests pass.

Fixes #1175

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -30,7 +30,7 @@ from ..._streaming import Stream, AsyncStream
 from ...types.beta import BetaRawMessageStreamEvent
 from ..._utils._utils import is_given
 from .._parse._response import ResponseFormatT, parse_text
-from ...types.beta.parsed_beta_message import ParsedBetaMessage, ParsedBetaContentBlock
+from ...types.beta.parsed_beta_message import ParsedBetaMessage, ParsedBetaContentBlock, ParsedBetaTextBlock
 
 
 class BetaMessageStream(Generic[ResponseFormatT]):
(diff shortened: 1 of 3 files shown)


dtmeadows-ant:
@bledden thanks for this. The underlying issue (#1175) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1206, opened 2026-02-26
rewrite test_files: structured, parametrized, edge-case coverage

tpbrown:
Rewrites tests/test_files.py from 6 flat tests (with print debugging) to 50 parametrized, grouped tests covering all public functions in anthropic._files: type guards (is_file_content, is_base64_file_input, assert_is_file_content), sync/async content reading, and sync/async httpx transformation. Adds edge cases: empty inputs, None passthrough, invalid top-level types, tuple passthrough semantics, IO/bytes identity checks.

### DIFF
--- a/tests/test_files.py
+++ b/tests/test_files.py
@@ -1,51 +1,291 @@
+"""Tests for anthropic._files: file transformation utilities for httpx.
+
+Covers to_httpx_files, async_to_httpx_files, read_file_content,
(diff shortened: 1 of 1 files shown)


localden:
Thank you for the PR - these files are auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Closing.

PR #1207, opened 2026-02-26
fix(bedrock): use boto3 client for AWS region detection

micheallam130:
## Description

Fixes issue #892 - Bedrock client failing to detect AWS region correctly when using AWS_PROFILE.

## Changes

- Use `boto3.client("bedrock").meta.region_name` instead of `boto3.Session().region_name` to properly detect AWS region from `~/.aws/config` when using `AWS_PROFILE`
- Add `NoRegionError` exception handling as fallback
- Maintain backward compatibility with the existing fallback to `us-east-1`

## Testing

This fix allows the Bedrock client to correctly detect the AWS region from the user's AWS configuration when using `AWS_PROFILE` environment variable.

### DIFF
--- a/src/anthropic/_utils/_utils.py
+++ b/src/anthropic/_utils/_utils.py
@@ -181,6 +181,7 @@ def deepcopy_minimal(item: _T) -> _T:
 
     - mappings, e.g. `dict`
     - list
+    - tuple (recursively copies contents)
 
     This is done for performance reasons.
     """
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@micheallam130 thanks for this. The underlying issue (#892) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1214, opened 2026-03-01
fix: use filtered messages in async _check_and_compact()

### LINKED ISSUE #1205: Bug: async _check_and_compact() ignores tool_use filtering, causes BadRequestError
## Bug

When using the async tool runner with compaction enabled, `_check_and_compact()` sends unfiltered `tool_use` blocks to the API, causing a `BadRequestError`.

**Error:**
```
BadRequestError: messages.2: `tool_use` ids were found without 
`tool_result` blocks immediately after: toolu_test123
```

## Root Cause

In `src/anthropic/lib/tools/_beta_runner.py`, the async `_check_and_compact()` method (line ~472) uses `self._params["messages"]` instead of the filtered local `messages` variable when building the summarization request.

The filtering logic (lines ~457-469) correctly removes assistant messages containing only `tool_use` blocks, but the result is discarded because `self._params["messages"]` is used instead:

```python
# Line ~472 (async version - BUGGY):
messages = [
    *self._params["messages"],  # ignores the filtering done above
    ...
]

# Line ~211 (sync version - CORRECT):
messages = [
    *messages,  # uses the filtered variable
    ...
]
```

## Reproduction

Gist with standalone reproduction (no API key needed):
https://gist.github.com/fede-kamel/175f0e99abdb53a380ded3f41236e7dd

## Fix

PR #1124 — one-line change: replace `self._params["messages"]` with `messages` on line ~472.

The PR also fixes a similar issue in `append_messages` (line ~120) where a lambda captures `self._params["messages"]` at definition time instead of using `params["messages"]` at call time.

thakoreh:
## Summary

Fixes #1205

The async `_check_and_compact()` method was using `self._params['messages']` instead of the filtered local `messages` variable when building the summarization request. This caused `tool_use` blocks to be sent to the API without corresponding `tool_result` blocks, resulting in `BadRequestError`.

## Root Cause

In `src/anthropic/lib/tools/_beta_runner.py`, line 478:

```python
# Before (BUGGY)
messages = [
    *self._params["messages"],  # <-- Uses unfiltered messages
    BetaMessageParam(...)
]
```

The sync version of this method (line 218) correctly uses `*messages` (the filtered variable), but the async version was using the unfiltered `self._params["messages"]`.

## Fix

Changed line 478 to use the filtered `messages` variable:

```python
# After (FIXED)
messages = [
    *messages,  # <-- Uses filtered messages (tool_use blocks removed)
    BetaMessageParam(...)
]
```

## Testing

- The fix aligns the async implementation with the sync implementation
- The filtering logic (lines 459-476) now has its effect preserved when building the summarization request

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -469,7 +469,7 @@ class BaseAsyncToolRunner(
                 messages.pop()
 
         messages = [
-            *self._params["messages"],
+            *messages,
             BetaMessageParam(
                 role="user",
                 content=self._compaction_control.get("summary_prompt", DEFAULT_SUMMARY_PROMPT),


karpetrosyan:
#1205 was already fixed

PR #1217, opened 2026-03-02
feat(exceptions): export missing HTTP status exceptions to public interface

sbobryshev:
## Problem

Four exception classes are defined in `_exceptions.py` and raised by the SDK
internally, but are not exported in `__all__` or the top-level `anthropic`
package — making them impossible to catch by name:

| Exception | HTTP status |
|---|---|
| `RequestTooLargeError` | 413 |
| `ServiceUnavailableError` | 503 |
| `OverloadedError` | 529 |
| `DeadlineExceededError` | 504 |

This forces users to write:

```python
except APIStatusError as e:
    if e.status_code == 529:
        ...
```

instead of the idiomatic:

```python
except anthropic.OverloadedError:
    ...
```

## Changes

- Added the four missing exceptions to `__all__` in `_exceptions.py`
- Imported and exported them from the top-level `anthropic` package (`__init__.py`)

## Notes

This is a purely additive, non-breaking change. `RateLimitError` (429) is
already public — `OverloadedError` (529) and `ServiceUnavailableError` (503)
are equally common API errors and should be available for the same reasons.


tomer-ant:
Thanks, but this will be fixed in the next release

PR #1218, opened 2026-03-03
fix: prevent dict mutation in deepcopy_minimal

### LINKED ISSUE #1202: Use of `deepcopy_minimal` in `files.beta.upload` mutates dict in place

abdelhadi703:
## Summary
Fix `deepcopy_minimal` mutating user-provided dictionaries in place when used in `files.beta.upload`.

The function previously only deep-copied `dict` and `list` types, but ignored `tuple`. Since `FileTypes` values are commonly tuples like `(filename, content, content_type, headers_dict)`, any nested dicts inside these tuples (such as the headers mapping) were shared by reference between the original and the "copy". When `extract_files` subsequently mutated the body dict, it could inadvertently modify the user's original data structures.

This fix adds tuple support to `deepcopy_minimal` so that all nested structures within tuples are properly deep-copied.

Fixes #1202

## Changes
- Add tuple copying support to `deepcopy_minimal` in `src/anthropic/_utils/_utils.py`
- Add tests for tuple deep-copying in `tests/test_deepcopy.py`
- Remove outdated test that asserted tuples should NOT be copied

## Test plan
- [x] New test `test_simple_tuple` verifies tuples are copied
- [x] New test `test_tuple_with_nested_dict` verifies dicts inside tuples get their own identity
- [x] New test `test_dict_with_tuple_value` verifies the full scenario (dict -> tuple -> dict)
- [x] All existing deepcopy tests continue to pass

### DIFF
--- a/src/anthropic/_utils/_utils.py
+++ b/src/anthropic/_utils/_utils.py
@@ -181,6 +181,7 @@ def deepcopy_minimal(item: _T) -> _T:
 
     - mappings, e.g. `dict`
     - list
+    - tuple
 
     This is done for performance reasons.
     """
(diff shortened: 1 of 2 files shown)


abdelhadi703:
Hi @anthropics/sdk-team,

Gentle ping on this bug fix. It prevents dict mutation in `deepcopy_minimal` which could cause unexpected side effects when copying objects.

Ready to address any review comments. Thanks!

localden:

PR #1221, opened 2026-03-05
add tool use example

ozge-devops:
## Description

Add a tool use (function calling) example to the `examples/` directory.

This example demonstrates the complete tool use workflow:
- Defining tools with JSON Schema input descriptions
- Sending a message that triggers tool use
- Processing tool calls in a loop (handling multiple tool calls)
- Returning tool results back to Claude
- Getting Claude's final response incorporating the tool results

The example uses a simulated weather API to keep it simple and runnable without external dependencies.

## Motivation

The SDK supports tool use but the `examples/` directory doesn't have a dedicated example for it. This makes it harder for new users to get started with function calling. The README references tool use support but links to external docs rather than a working example.

## Testing

Tested locally with `ANTHROPIC_API_KEY` set — Claude correctly

### DIFF
--- a/examples/tool_use.py
+++ b/examples/tool_use.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env -S uv run python
+
+"""Example demonstrating tool use (function calling) with the Anthropic SDK.
+
+This example shows how to:
+1. Define tools with JSON Schema input descriptions
+2. Send a message that triggers tool use
+3. Process the tool call and return results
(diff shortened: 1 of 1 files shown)


localden:
Thank you for the PR - this was already addressed on `main` - see [`examples/tools.py:1`](https://github.com/anthropics/anthropic-sdk-python/blob/main/examples/tools.py#L1). Closing as resolved.

PR #1222, opened 2026-03-05
test: add unit tests for tool use example

ozge-devops:
## Description

Add unit tests for the tool use example (`examples/tool_use.py`).

## Test coverage

- get_weather() – known locations,unknown locations,default values
- Temperature unit conversion(Celsius → Fahrenheit)
- process_tool_call() – routing,error handling, JSON validation
- Full tool use loop with mocked API (no API key required)

## Details

- 5 test classes,16 test functions
- Uses `pytest.mark.parametrize` for data-driven testing
- Uses `unittest.mock` to simulate API responses
- All tests run offline without API credentials

### DIFF
--- a/tests/test_tool_use.py
+++ b/tests/test_tool_use.py
@@ -1,0 +1,278 @@
+"""Unit tests for the tool_use example.
+
+Tests cover:
+- Weather data retrieval with known and unknown locations
+- Temperature unit conversion (Celsius to Fahrenheit)
+- Tool call routing and error handling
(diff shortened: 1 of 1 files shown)


localden:
Thank you for the PR. Closing - this adds tests for [`examples/tool_use.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/examples/tool_use.py), which doesn't exist on `main`, and `examples/` isn't an importable package so the test imports would fail regardless. We don't unit-test demo scripts in this repo.

PR #1226, opened 2026-03-08
fix: handle tuples in deepcopy_minimal to prevent dict mutation

### LINKED ISSUE #1202: Use of `deepcopy_minimal` in `files.beta.upload` mutates dict in place

giulio-leone:
## Summary

Fixes #1202 — `deepcopy_minimal` does not recurse into tuples, causing `files.beta.upload` to mutate the caller's headers dict in-place.

## Root Cause

`deepcopy_minimal` only handles `dict` and `list`, returning everything else as-is. When a `FileTypes` tuple like `(name, content, mime, headers_dict)` is passed, the tuple is shared by reference, so any downstream mutation of the headers dict (element [3]) leaks back to the caller.

## Fix

Add tuple handling to `deepcopy_minimal`:

```python
if isinstance(item, tuple):
    return cast(_T, tuple(deepcopy_minimal(entry) for entry in item))
```

This ensures dicts (and lists) nested inside tuples are recursively copied.

## Tests

- Updated existing tuple test (was asserting `obj3 is obj4` identity, now checks proper copy)
- Added `test_tuple_with_nested_dict` — verifies the exact `FileTypes` scenario from #1202: mutating the copy's headers dict does not affect the original
- All 8 tests pass

### DIFF
--- a/src/anthropic/_utils/_utils.py
+++ b/src/anthropic/_utils/_utils.py
@@ -181,6 +181,7 @@ def deepcopy_minimal(item: _T) -> _T:
 
     - mappings, e.g. `dict`
     - list
+    - tuple
 
     This is done for performance reasons.
     """
@@ -188,6 +189,8 @@ def deepcopy_minimal(item: _T) -> _T:
         return cast(_T, {k: deepcopy_minimal(v) for k, v in item.items()})
     if is_list(item):
         return cast(_T, [deepcopy_minimal(entry) for entry in item])
+    if isinstance(item, tuple):
+        return cast(_T, tuple(deepcopy_minimal(entry) for entry in item))
     return item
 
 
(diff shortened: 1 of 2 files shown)


giulio-leone:
Friendly ping — CI is green, tests pass, rebased on latest. Ready for review whenever convenient. Happy to address any feedback. 🙏

giulio-leone:
Added a real runtime validation pass for this fix using the public `Anthropic(...).beta.files.upload(file=...)` path against a local HTTP server (no mocks/stubs in the reproduction itself).

Results:
- On `upstream/main`, the caller-owned headers dict was mutated in place by the upload call: `{'X-Original': '1'}` became `{'X-Original': '1', 'Content-Type': 'text/plain'}` after the request completed.\n- On this branch, the exact same upload call returned a valid `FileMetadata` response and left the original headers dict unchanged.\n- Targeted regression suite still passes locally: `.venv/bin/python -m pytest tests/test_deepcopy.py -q` -> `8 passed`.\n\nSo this is now backed by both the focused tests and a real before/after reproduction of the public upload flow from issue #1202.

giulio-leone:

PR #1228, opened 2026-03-08
fix: handle tuples in `deepcopy_minimal` to prevent in-place mutation

### LINKED ISSUE #1202: Use of `deepcopy_minimal` in `files.beta.upload` mutates dict in place

MaxwellCalkin:
## Summary

Fixes #1202.

`deepcopy_minimal` only recurses into `dict` and `list`, returning all other types (including `tuple`) as-is. This causes mutable objects nested inside tuples to be shared by reference between the original and the "copy", leading to unexpected in-place mutation of the caller's data.

The most common case is `FileTypes` values of the form `(filename, content, content_type, headers_dict)` passed to `files.beta.upload()`. The headers dict (4th tuple element) is not copied, so the SDK's internal processing mutates the caller's original dict.

## Changes

- **`src/anthropic/_utils/_utils.py`**: Add `tuple` handling to `deepcopy_minimal` using the existing `is_tuple()` helper, so mutable contents within tuples are properly deep-copied.
- **`tests/test_deepcopy.py`**: Add three test cases covering:
  - Tuples with mutable dict contents (the reported bug scenario)
  - Tuples nested inside dicts
  - Tuples nested inside lists

## Before / After

```python
inner_dict = {"content-type": "application/json"}
file_tuple = ("file.txt", b"content", "application/json", inner_dict)
copied = deepcopy_minimal(file_tuple)

# BEFORE: inner_dict is shared — mutation leaks back
copied[3]["x-new"] = "value"
assert "x-new" in inner_dict  # True — bug!

# AFTER: inner_dict is properly copied — no mutation leak
copied[3]["x-new"] = "value"
assert "x-new" not in inner_dict  # True — fixed!
```

---

> **Transparency note:** This PR was authored by an AI (Claude Opus 4.6) as part of an effort to demonstrate value through open-source contributions. The fix and tests were written after careful analysis of the issue report and source code.

### DIFF
--- a/src/anthropic/_utils/_utils.py
+++ b/src/anthropic/_utils/_utils.py
@@ -181,6 +181,7 @@ def deepcopy_minimal(item: _T) -> _T:
 
     - mappings, e.g. `dict`
     - list
+    - tuple
 
     This is done for performance reasons.
     """
@@ -188,6 +189,8 @@ def deepcopy_minimal(item: _T) -> _T:
         return cast(_T, {k: deepcopy_minimal(v) for k, v in item.items()})
     if is_list(item):
         return cast(_T, [deepcopy_minimal(entry) for entry in item])
+    if is_tuple(item):
+        return cast(_T, tuple(deepcopy_minimal(entry) for entry in item))
     return item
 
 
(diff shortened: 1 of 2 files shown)


localden:

PR #1236, opened 2026-03-09
fix(bedrock): use boto3 Session region from aws_profile instead of hardcoded us-east-1

giulio-leone:
Fixes #892

## Problem

When `AWS_REGION` is unset, `AnthropicBedrock` falls back to a hardcoded `us-east-1` region instead of reading the region from the user's AWS profile. This means users who configure a non-default region via `aws_profile` (e.g., `eu-west-1`, `ap-southeast-1`) are silently routed to `us-east-1`, causing cross-region inference failures or unexpected latency. The issue is particularly confusing because the user has correctly configured their AWS profile but the SDK ignores it.

## Root Cause

The `_infer_region()` helper did not accept or use the `aws_profile` parameter. It checked only the `AWS_REGION` / `AWS_DEFAULT_REGION` environment variables and, if neither was set, returned `us-east-1` unconditionally — even when a valid profile with a configured region was available.

## Fix

- `_infer_region()` now accepts an optional `aws_profile` parameter and creates a `boto3.Session(profile_name=aws_profile)` to read the configured region.
- Both sync (`AnthropicBedrock`) and async (`AsyncAnthropicBedrock`) constructors forward their `aws_profile` argument to `_infer_region()`.
- Falls back to `us-east-1` only when no profile is provided and no environment variable is set, preserving backward compatibility.

## Testing

Two new tests verify:
- Profile-based region detection returns the correct region from `boto3.Session`
- Graceful fallback to `us-east-1` when no profile or environment region is available

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -67,7 +67,7 @@ def _prepare_options(input_options: FinalRequestOptions) -> FinalRequestOptions:
     return options
 
 
-def _infer_region() -> str:
+def _infer_region(aws_profile: str | None = None) -> str:
     """
     Infer the AWS region from the environment variables or
     from the boto3 session if available.
@@ -77,7 +77,7 @@ def _infer_region() -> str:
         try:
             import boto3
 
-            session = boto3.Session()
+            session = boto3.Session(profile_name=aws_profile)
             if session.region_name:
                 aws_region = session.region_name
         except ImportError:
@@ -161,7 +161,7 @@ class AnthropicBedrock(BaseBedrockClient[httpx.Client, Stream[Any]], SyncAPIClie
 
         self.aws_access_key = aws_access_key
 
-        self.aws_region = _infer_region() if aws_region is None else aws_region
+        self.aws_region = _infer_region(aws_profile) if aws_region is None else aws_region
         self.aws_profile = aws_profile
 
         self.aws_session_token = aws_session_token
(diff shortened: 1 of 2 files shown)


giulio-leone:
Friendly ping — CI is green, tests pass, rebased on latest. Ready for review whenever convenient. Happy to address any feedback. 🙏

giulio-leone:

giulio-leone:
Added one small follow-up commit to harden the constructor-profile regressions against ambient AWS env leakage.

Validation (double pass):
- `env -u AWS_REGION -u AWS_DEFAULT_REGION -u AWS_PROFILE -u AWS_CONFIG_FILE uv run --extra bedrock ruff check tests/lib/test_bedrock.py`
- `env -u AWS_REGION -u AWS_DEFAULT_REGION -u AWS_PROFILE -u AWS_CONFIG_FILE uv run --extra bedrock ruff format --check tests/lib/test_bedrock.py`
- `env -u AWS_REGION -u AWS_DEFAULT_REGION -u AWS_PROFILE -u AWS_CONFIG_FILE uv run --extra bedrock pytest tests/lib/test_bedrock.py -q -n 0`

Direct runtime proof in the prepared venv with a temp AWS config file:
- `sync_default_region=us-east-2`
- `sync_custom_region=us-west-1`
- `async_custom_region=us-west-1`

So the branch still proves the original fix, and the tests are now hermetic about explicit `aws_profile=` inference too.

PR #1239, opened 2026-03-11
fix: defer async stream transforms for large payloads

### LINKED ISSUE #1195: Python SDK: _transform_recursive blocking event loop on large message payloads
We profiled our FastAPI gateway (which runs an agent orchestration loop using the Anthropic Python SDK) and found that _transform_recursive in anthropic/_utils/_transform.py consumes ~6.6% of total CPU time during heavy workloads.

The function recursively walks the entire messages + tools payload doing type introspection (get_type_hints, is_typeddict, is_union_type) at every level of nesting. For Messages API types, this produces zero changes — no key renames, no date formatting. The transform allocates new dicts at every level and copies every key-value pair, producing an identical structure.

Additionally, AsyncMessages.stream() calls the synchronous maybe_transform (not async_maybe_transform), so the entire recursive walk blocks the event loop.

We're currently working around this by passing messages and tools via extra_body so the transform only runs on empty placeholders. Has this been reported as a known issue? Is there a better recommended approach?

giulio-leone:
Fixes #1195.

## Summary
- keep the existing transform-cache optimizations and make them pass the repo lint/type gate cleanly
- defer `AsyncMessages.stream()` request construction so large payload transforms run inside the awaited path instead of synchronously during `stream()` manager creation
- apply the same deferral to `client.beta.messages.stream()`
- add regression tests proving the async stream helpers do not eagerly call `maybe_transform`

## Why
`AsyncMessages.stream()` and `beta.messages.stream()` are synchronous methods that return async context managers. Before this patch, both methods performed expensive request-body transforms up front, so large `messages` / `tools` payloads could block the event loop before the first await.

This change moves that work into the coroutine consumed by `AsyncMessageStreamManager.__aenter__`, which preserves behavior while removing the unexpected synchronous blocking at `stream()` call time.

## Real runtime verification
I validated this against a live local HTTP server using the public SDK APIs, with large real payloads (3000 messages, 400 tools), comparing:
- upstream `main` (`49d639a`)
- the cache-only optimization branch (`0767668`)
- this final branch (`0767668` + deferred stream construction)

Median results across two consecutive quality-gate passes:

- `client.messages.stream(...)` construction
  - `main`: ~823-833 ms blocking on the event loop
  - cache-only: ~421-428 ms
  - final branch: ~0.017-0.019 ms

- `client.beta.messages.stream(...)` construction
  - `main`: ~1010-1031 ms
  - cache-only: ~536-539 ms
  - final branch: ~1.106-1.115 ms

- full `async with client.messages.stream(...)` time stays in the same general range (~437 ms cache-only vs ~480 ms final), which is expected: the work is deferred into the awaited path rather than silently removed.

## Validation
Run twice consecutively on this branch:
- `./scripts/lint`
- `uv run pytest tests/lib/streaming/test_messages.py tests/lib/streaming/test_beta_messages.py tests/lib/_parse/test_beta_messages.py tests/test_transform.py -q`
- live runtime benchmark via local HTTP server using the public async streaming APIs

### DIFF
--- a/src/anthropic/_utils/_transform.py
+++ b/src/anthropic/_utils/_transform.py
@@ -395,7 +395,7 @@ async def _async_transform_recursive(
 
     elif dispatch == _DISPATCH_DICT:
         if is_mapping(data) and type_arg is not None:
-            return {key: _transform_recursive(value, annotation=type_arg) for key, value in data.items()}
+            return {key: await _async_transform_recursive(value, annotation=type_arg) for key, value in data.items()}
 
     elif dispatch == _DISPATCH_SEQUENCE:
         if not isinstance(data, (str, dict)):
--- a/src/anthropic/_utils/_transform.py
+++ b/src/anthropic/_utils/_transform.py
@@ -148,7 +148,63 @@ def _maybe_transform_key(key: str, type_: type) -> str:
 
 
 def _no_transform_needed(annotation: type) -> bool:
-    return annotation == float or annotation == int
+    return annotation == float or annotation == int or annotation == str or annotation == bool
+
+
+# Dispatch kinds for _cached_transform_dispatch
+_DISPATCH_TYPEDDICT = 0
+_DISPATCH_DICT = 1
+_DISPATCH_SEQUENCE = 2
+_DISPATCH_UNION = 3
+_DISPATCH_OTHER = 4
+
+
+@lru_cache(maxsize=8096)
+def _cached_transform_dispatch(inner_type: type) -> tuple[int, Any]:
+    """Precompute the transform dispatch for a type annotation.
+
+    Caches the result of strip_annotated_type, get_origin, is_typeddict,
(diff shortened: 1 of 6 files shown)


giulio-leone:
Quick heads-up: both PR workflows are currently in `action_required` with zero jobs, so this looks like the standard forked-PR approval gate rather than a failing CI run.

For convenience, I already validated this branch locally twice with:
- `./scripts/lint`
- `uv run pytest tests/lib/streaming/test_messages.py tests/lib/streaming/test_beta_messages.py tests/lib/_parse/test_beta_messages.py tests/test_transform.py -q`
- a live local-server benchmark using the public async streaming APIs

That benchmark showed `client.messages.stream(...)` construction drop from ~421-428 ms on the cache-only branch to ~0.017-0.019 ms here, and beta stream construction drop from ~536-539 ms to ~1.106-1.115 ms.

giulio-leone:

PR #854, opened 2025-01-31
feat(client): add more status exceptions

andjsmi:
Related issue: https://github.com/anthropics/anthropic-sdk-python/issues/853

This PR adds the exception codes mentioned in 

- https://docs.anthropic.com/en/api/errors
- https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/api-errors
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html#API_runtime_InvokeModel_Errors

The main motivation is allowing for easier handling if an API provider throws a 503/529 to indicate overload as it allows for an application to have retry logic built in easier.

PR #1244, opened 2026-03-14
fix(client): AsyncAnthropic._make_status_error missing 529 and 413 cases

stephen:
## What

`AsyncAnthropic._make_status_error` is missing the `529 → OverloadedError` and `413 → RequestTooLargeError` cases that the sync `Anthropic._make_status_error` has. So for async users:

- HTTP 529 raises `InternalServerError` (falls through to the `>= 500` catch-all) instead of `OverloadedError`
- HTTP 413 raises the base `APIStatusError` instead of `RequestTooLargeError`

This PR copies the two missing `if` arms from the sync method into the async one, and adds a parametrized parity test so the two copies can't drift silently again.

## Origin

#854 (commit b87dcf12, merged 2025-02-19) added these two cases to the sync client at `_client.py` L257. The diff in that PR is a single hunk touching only the sync method — the async copy further down in the same file was missed. Still the case on `main` today:

```
sync   L285: if response.status_code == 529: return OverloadedError(...)
async  L522: if response.status_code >= 500: return InternalServerError(...)   # no 529 check
```

## Verification

Ran the new parity test against current `main` — it fails on exactly 413 and 529:

```
FAIL 413: sync=RequestTooLargeError         async=APIStatusError
FAIL 529: sync=OverloadedError              async=InternalServerError
```

and passes with the fix applied.

### DIFF
--- a/src/anthropic/_client.py
+++ b/src/anthropic/_client.py
@@ -513,11 +513,17 @@ class AsyncAnthropic(AsyncAPIClient):
         if response.status_code == 409:
             return _exceptions.ConflictError(err_msg, response=response, body=body)
 
+        if response.status_code == 413:
+            return _exceptions.RequestTooLargeError(err_msg, response=response, body=body)
+
         if response.status_code == 422:
             return _exceptions.UnprocessableEntityError(err_msg, response=response, body=body)
 
         if response.status_code == 429:
             return _exceptions.RateLimitError(err_msg, response=response, body=body)
+
+        if response.status_code == 529:
+            return _exceptions.OverloadedError(err_msg, response=response, body=body)
 
         if response.status_code >= 500:
             return _exceptions.InternalServerError(err_msg, response=response, body=body)
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -112,6 +112,22 @@ def _get_open_connections(client: Anthropic | AsyncAnthropic) -> int:
     return len(pool._requests)
 
 
+@pytest.mark.parametrize("status_code", [400, 401, 403, 404, 409, 413, 422, 429, 500, 503, 529])
+def test_make_status_error_sync_async_parity(status_code: int) -> None:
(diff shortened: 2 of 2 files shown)


alvinttang (commented):
## Code Review

Good catch — the sync/async parity gap is a real bug. The fix and the parametrized parity test are both clean.

### Suggestion: The parity test should also cover future status codes

The test hardcodes the status code list:

```python
@pytest.mark.parametrize("status_code", [400, 401, 403, 404, 409, 413, 422, 429, 500, 503, 529])
```

If someone adds a new status code handler to the sync client in the future, this test won't catch the omission from the async client unless the new code is also added to the parametrize list. Consider making the test more robust by dynamically discovering all handled status codes, e.g.:

```python
import re, inspect

def _extract_status_codes(method):
    """Extract all status codes checked in _make_status_error."""
    source = inspect.getsource(method)
    return set(int(m) for m in re.findall(r'status_code\s*==\s*(\d+)', source))

def test_sync_async_handle_same_status_codes():
    sync_codes = _extract_status_codes(Anthropic._make_status_error)
    async_codes = _extract_status_codes(AsyncAnthropic._make_status_error)
    assert sync_codes == async_codes, f"Sync has {sync_codes - async_codes} extra, async has {async_codes - sync_codes} extra"
```

This way the test is self-maintaining and catches any future drift automatically without needing to update the parametrize list.

### Nit: 529 placement relative to the `>= 500` catch-all

The `529` check is correctly placed **before** the `>= 500` catch-all, matching the sync client. Worth adding a brief code comment noting why the ordering matters (529 would otherwise fall through to `InternalServerError`), since this exact ordering issue is what caused the original bug.

cameron-mcateer (approved):

localden (approved):

felixfbecker (approved):

karpetrosyan:
@cameron-mcateer @localden I think we already have this fix in the next branch, I've commited it ~week ago

Issue #1192, opened 2026-02-20
IndexError when streaming


palako:
https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/streaming/_messages.py line 465 raises an IndexError, it seems that ither current_snapshot.content is empty or doesn't have at least event.index+1 elements.

Unfortunatelly I can't reproduce the problem consistently, so I can't provide more info, but an author more familiar with that code should be able to identify where to add the correct boundaries validation.

xXMrNidaXx:
We've seen sporadic IndexError issues like this at [RevolutionAI](https://revolutionai.io) when streaming under high concurrency. Based on the location (line 465 in `_messages.py`), this appears to be a race condition in how content blocks are indexed during streaming.

**Likely cause:**

The streaming handler expects content blocks to arrive in order, but under certain conditions (network jitter, provider-side batching, or concurrent streams), the `event.index` can reference a content block that hasn't been created yet:

```python
# The problematic pattern:
content_block = current_snapshot.content[event.index]  # IndexError if index > len
```

**Workaround until fixed:**

Wrap your streaming loop in a try/except to handle the edge case:

```python
try:
    async for event in stream:
        # process event
except IndexError as e:
    # Log for diagnostics, retry, or graceful degradation
    logger.warning(f"Stream index mismatch: {e}, retrying...")
    # Retry logic or fallback to non-streaming
```

**Suggested fix:**

The SDK should bounds-check before accessing the array:

```python
# Defensive approach
if event.index >= len(current_snapshot.content):
    # Either expand the array or skip this event
    pass
content_block = current_snapshot.content[event.index]
```

**Debugging questions:**

1. Are you using `stream_mode="text"` or `stream_mode="raw"`?
2. Any tool calls involved in the response?
3. Concurrent streams from the same client instance?

Tool calls especially seem to cause index misalignment since they can insert multiple content blocks in rapid succession.

Would be helpful to capture the exact `event.index` value and `len(current_snapshot.content)` when this happens — even intermittent repros help narrow it down.

dtmeadows-ant:
[...]

PR #1251, opened 2026-03-17
Fix: IndexError when streaming with multiple content blocks (#1192)

Adityakk9031:
Fix-#1192
Problem
An IndexError: list index out of range is raised at current_snapshot.content[event.index] inside both 
accumulate_event
 (line 465) and 
build_events
 (line 407) when streaming responses with multiple content blocks.

The root cause is content_block_start blindly calling .append() without validating event.index against the current list length — a known gap flagged by a # TODO: check index comment at that line. If the server sends a non-contiguous index, the list is shorter than expected, and any subsequent content_block_delta or content_block_stop for that index crashes.

Fix
5 changes in 
src/anthropic/lib/streaming/_messages.py
:

content_block_start — resolve the TODO: validate event.index; if ahead of the list, pad with empty placeholder blocks and emit a RuntimeWarning.
content_block_delta in 
accumulate_event
 — early-return if event.index >= len(content).
content_block_stop in 
accumulate_event
 — same guard.
content_block_delta in 
build_events
 — same guard.
content_block_stop in 
build_events
 — same guard.

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -348,6 +348,10 @@ def build_events(
     elif event.type == "content_block_delta":
         events_to_fire.append(event)
 
+        if event.index >= len(message_snapshot.content):
+            # Delta arrived for a block that was never started — skip gracefully.
+            return events_to_fire
+
         content_block = message_snapshot.content[event.index]
         if event.delta.type == "text_delta":
             if content_block.type == "text":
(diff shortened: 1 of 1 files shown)


dtmeadows-ant:
@Adityakk9031 thanks for digging into this, and sorry it sat. Padding the content list with placeholder blocks would keep the stream going, but it also hands callers a message with fabricated empty blocks. More to the point, we couldn't reproduce out-of-order indices against the API directly when closing #1192; the reports traced back to intermediaries rewriting the stream. If you have a repro or a `request-id` from a direct call, reply there and I'll take another look.

Issue #1258, opened 2026-03-18
Mid-stream SSE errors get `status_code=200` instead of the actual error code


sarth6:
## Bug

When the API returns HTTP 200 (streaming started fine) but then sends an SSE error event like `overloaded_error`, the SDK creates an `APIStatusError` with `status_code=200` instead of 529. The error body has the right type, but the status code is wrong.

This happens because `_streaming.py` passes `self.response` (the original HTTP response object) to `_make_status_error`, and `_make_status_error` switches on `response.status_code` to pick the error subclass.

## Code path

In `_streaming.py` (line ~229):

```python
if sse.event == "error":
    body = sse.json()
    err_msg = f"{body}"
    raise self._client._make_status_error(
        err_msg,
        body=body,
        response=self.response,  # <-- this is the original HTTP 200 response
    )
```

In `_client.py`, `_make_status_error` dispatches on `response.status_code`:

```python
def _make_status_error(self, err_msg, *, body, response):
    if response.status_code == 429:
        return RateLimitError(...)
    if response.status_code >= 500:
        return InternalServerError(...)
    return APIStatusError(...)  # <-- falls through here because status is 200
```

Since the HTTP status was 200, none of the specific branches match. You get a bare `APIStatusError` with `status_code=200`.

## Why this matters

Any caller checking `status_code` for retry/fallback decisions gets the wrong answer. An `overloaded_error` should look like a 529, not a 200. The body has the correct error type (`{'error': {'type': 'overloaded_error'}}`), but most callers don't expect to need to parse that — they check status codes.

We hit this in production with pydantic-ai's `FallbackModel`, which checks `status_code >= 500` to decide whether to try the next model. The overloaded error had `status_code=200`, so fallback never fired.

## Expected behavior

Mid-stream SSE errors should produce the same error subclass as if the error came back as an HTTP response. An `overloaded_error` → `OverloadedError(status_code=529)`, an `api_error` → `InternalServerError(status_code=500)`, etc.

## Suggested fix

Map the SSE error type to the corresponding status code before calling `_make_status_error`, or construct the right error subclass directly:

```python
_SSE_ERROR_TYPE_TO_STATUS = {
    "overloaded_error": 529,
    "rate_limit_error": 429,
    "api_error": 500,
    "internal_server_error": 500,
    # client errors (shouldn't appear mid-stream, but for completeness):
    "authentication_error": 401,
    "invalid_request_error": 400,
    "not_found_error": 404,
}
```

[...]

PR #1263, opened 2026-03-19
fix: map SSE error types to correct HTTP status codes in streaming

Jah-yee:
## Summary

When the API returns HTTP 200 (streaming started fine) but then sends an SSE error event like `overloaded_error`, the SDK was creating an `APIStatusError` with `status_code=200` instead of the correct error code (e.g., 529 for overloaded).

This happened because `_streaming.py` passed the original HTTP response (status 200) to `_make_status_error`, which then selected the error subclass based on that 200 status code.

## Fix

Added a mapping from SSE error types to HTTP status codes, and create a mock response with the correct status code when handling SSE errors in both sync and async streaming.

## Changes

- Added `_SSE_ERROR_TYPE_TO_STATUS` mapping in `_streaming.py`
- Added `_get_status_code_from_sse_error()` helper function
- Modified error handling in both `Stream` and `AsyncStream` to use the mapped status code

## Testing

This fix ensures that:
- `overloaded_error` → `OverloadedError(status_code=529)`
- `rate_limit_error` → `RateLimitError(status_code=429)`
- `api_error` → `InternalServerError(status_code=500)`
- etc.

Fixes: #1258

### DIFF
--- a/src/anthropic/_streaming.py
+++ b/src/anthropic/_streaming.py
@@ -19,6 +19,28 @@ if TYPE_CHECKING:
 
 
 _T = TypeVar("_T")
+# Mapping from SSE error type to HTTP status code
+# See: https://docs.anthropic.com/en/api/error-codes
+_SSE_ERROR_TYPE_TO_STATUS = {
+    "overloaded_error": 529,
+    "rate_limit_error": 429,
+    "api_error": 500,
+    "internal_server_error": 500,
+    "authentication_error": 401,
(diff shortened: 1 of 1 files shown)


localden:
Thank you for the PR - this was addressed differently on `main` - we went with exposing `error.type` on `APIStatusError` rather than synthesizing a fake status code. This patch would also crash on the error path since `.content` raises `ResponseNotRead` on a streamed response. Closing as resolved.

Issue #1265, opened 2026-03-20
Streaming tool use: unhelpful error message when model emits malformed JSON (non-beta path missing try-except)


AbdoKnbGit:
When using `client.messages.stream()` with tools, if the model emits malformed JSON during an `input_json_delta`
event, the user gets a raw internal parser error with no context:

```
ValueError: expected ident at line 1 column 11
```

The beta path (`_beta_messages.py`) already wraps the same call with a clear, actionable message. The non-beta path
(`_messages.py`) is simply missing the same `try-except`.

## Impact

Every user of `client.messages.stream()` with tool calling is affected. When this triggers, there is no indication
that the error came from tool input JSON parsing, no mention of what JSON was received, and no guidance on what to do.
It looks like an SDK crash.

## Confirmed with real API call (`anthropic==0.86.0`)

I made a real streaming tool call to `claude-haiku-4-5`, captured the response, then replayed the events
through `accumulate_event` with a corrupted JSON delta to trigger the error path in both the non-beta and beta
accumulators.

Normal streaming tool call works fine:

```
STEP 1: Real API streaming tool call (claude-haiku-4-5)

  received: message_start
  received: content_block_start
  received: content_block_delta
  received: input_json
  received: content_block_delta
  received: input_json
  received: content_block_delta
  received: input_json
  received: content_block_stop
  received: message_delta
  received: message_stop

  Tool called: get_weather({'city': 'Paris'})
```

Corrupted delta (`{"city": INVALID_VALUE}`) through both paths:

```
NON-BETA PATH (accumulate_event from _messages.py):
  ValueError: "expected ident at line 1 column 11"
  ^ Raw jiter error -- no context for the user

BETA PATH (accumulate_event from _beta_messages.py):
  ValueError: "Unable to parse tool parameter JSON from model.
  Please retry your request or adjust your prompt.
  Error: expected ident at line 1 column 11.
  JSON: {"city": INVALID_VALUE}"
  ^ Helpful: tells the user what happened and what to do
```

## Reproduction

```python
from anthropic.types import (
    RawMessageStartEvent, RawContentBlockStartEvent, RawContentBlockDeltaEvent,
)
from anthropic.types.message import Message
from anthropic.types.usage import Usage
from anthropic.types.tool_use_block import ToolUseBlock
from anthropic.types.input_json_delta import InputJSONDelta
from anthropic.types.direct_caller import DirectCaller
from anthropic.lib.streaming._messages import accumulate_event

[...]

PR #1266, opened 2026-03-20
fix: add missing JSON error handling in non-beta streaming accumulator

AbdoKnbGit:
Fixes #1265

The beta streaming accumulator (`_beta_messages.py`) wraps `from_json()` in a
try-except that gives users a clear, actionable error when the model emits
malformed tool input JSON. The non-beta accumulator (`_messages.py`) was missing
the same handler — causing users to see raw jiter errors with no context.

This PR adds the matching try-except to the non-beta path and a test mirroring
the existing beta test.

---

## Changes

- **`src/anthropic/lib/streaming/_messages.py`** — wrap `from_json()` at L480
  in a try-except with a helpful error message, matching `_beta_messages.py` L500-510
- **`tests/lib/streaming/test_partial_json.py`** — add
  `test_partial_json_with_invalid_json_non_beta`, mirroring the existing beta test

---

## Before / After

**Before** — non-beta path, malformed tool JSON:
```
ValueError: expected ident at line 1 column 11
```
No context. No mention of tool JSON. Looks like an SDK crash.

**After** — matches beta behavior:
```
ValueError: Unable to parse tool parameter JSON from model.
Please retry your request or adjust your prompt.
Error: expected ident at line 1 column 11. JSON: {"city": INVALID_VALUE}
```

---

## Test plan

- [x] New test `test_partial_json_with_invalid_json_non_beta` passes
- [x] All 26 streaming tests pass (25 existing + 1 new)
- [x] Confirmed with real API call against `claude-haiku-4-5`
- [x] Zero behavior change for valid JSON

PR #1267, opened 2026-03-20
fix(streaming): wrap tool JSON parsing with helpful error in non-beta path

Ratnaditya-J:
Ran into this while using `client.messages.stream()` with tools — when the model spits out malformed JSON in an `input_json_delta`, you get a raw `ValueError: expected ident at line 1 column 11` with zero context. The beta path (`_beta_messages.py`) already has a nice try-except that tells you it's a tool JSON parse failure and shows the actual JSON. The non-beta path just... doesn't.

This adds the same try-except wrapper around the `from_json()` call in `_messages.py` so both paths give you:

```
```

instead of the raw jiter error.

Also added a test for the non-beta path (the existing test only covered beta).

Fixes #1265

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -477,7 +477,12 @@ def accumulate_event(
                 json_buf += bytes(event.delta.partial_json, "utf-8")
 
                 if json_buf:
-                    content.input = from_json(json_buf, partial_mode=True)
+                    try:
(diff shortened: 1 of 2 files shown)


AbdoKnbGit:
why copy past bro ? i already fix it

Ratnaditya-J:
ah, didn't see #1266 was already up — same fix. closing in favor of that one, nice catch @AbdoKnbGit

PR #1269, opened 2026-03-20
fix(streaming): derive correct status code from SSE error type instead of HTTP 200

### LINKED ISSUE #1258: Mid-stream SSE errors get `status_code=200` instead of the actual error code

tao-to-python:
## Description

Fixes incorrect error classification for mid-stream SSE errors, as reported in #1258.

### Problem

When the API starts streaming (HTTP 200) then sends an SSE error event like `overloaded_error`, the SDK creates an `APIStatusError` with `status_code=200` because `_make_status_error` dispatches on `self.response.status_code`.

This means:
- `OverloadedError` (529) is never raised for mid-stream overload errors
- `RateLimitError` (429) is never raised for mid-stream rate limits
- Retry logic that catches specific error types doesn't work

### Solution

Add an `_SSE_ERROR_TYPE_TO_STATUS` mapping that derives the correct HTTP status code from the error body type:

```python
_SSE_ERROR_TYPE_TO_STATUS = {
    "overloaded_error": 529,
    "rate_limit_error": 429,
    "invalid_request_error": 400,
    ...
}
```

When a mapped error type is found, create a synthetic `httpx.Response` with the correct status code so `_make_status_error` returns the right error subclass.

### Changes

| File | Change |
|------|--------|
| `src/anthropic/_streaming.py` | Add error type → status code mapping for sync + async paths |

Fixes #1258

### DIFF
--- a/src/anthropic/_streaming.py
+++ b/src/anthropic/_streaming.py
@@ -17,6 +17,20 @@ if TYPE_CHECKING:
     from ._client import Anthropic, AsyncAnthropic
     from ._models import FinalRequestOptions
 
+# Mapping from SSE error type to HTTP status code.
+# When a mid-stream SSE error arrives, the HTTP response already has status 200.
+# We use this map to derive the correct status code for error classification.
+_SSE_ERROR_TYPE_TO_STATUS: dict[str, int] = {
+    "invalid_request_error": 400,
+    "authentication_error": 401,
+    "permission_error": 403,
+    "not_found_error": 404,
+    "request_too_large": 413,
+    "rate_limit_error": 429,
+    "api_error": 500,
+    "overloaded_error": 529,
+}
+
 
 _T = TypeVar("_T")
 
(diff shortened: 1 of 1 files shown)


localden:
Thank you for the PR - this was addressed differently on `main` - we went with exposing `error.type` on `APIStatusError` rather than synthesizing a fake status code (see [`89db282`](https://github.com/anthropics/anthropic-sdk-python/commit/89db282)). Closing as resolved.

PR #1273, opened 2026-03-20
fix(streaming): replace IndexError with descriptive RuntimeError for out-of-bounds content_block_delta

### LINKED ISSUE #1192: IndexError when streaming

tao-to-python:
## Description

Improves error diagnostics for intermittent `IndexError` during streaming, as reported in #1192.

### Problem

In `accumulate_event()`, `current_snapshot.content[event.index]` can raise an opaque `IndexError` when the index exceeds the content list length (e.g. due to a missing `content_block_start` event or a race condition).

The raw `IndexError` provides no context about what happened.

### Solution

Add bounds checking before the index access with a descriptive error:

```python
if event.index >= len(current_snapshot.content):
    raise RuntimeError(
        f"Received content_block_delta for index {event.index}, "
        f"but snapshot only has {len(current_snapshot.content)} content block(s). "
        f"This may indicate a missing content_block_start event."
    )
```

Applied to both `_messages.py` and `_beta_messages.py`.

Fixes #1192

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -482,6 +482,12 @@ def accumulate_event(
             ),
         )
     elif event.type == "content_block_delta":
+        if event.index >= len(current_snapshot.content):
+            raise RuntimeError(
+                f"Received content_block_delta for index {event.index}, "
+                f"but snapshot only has {len(current_snapshot.content)} content block(s). "
+                f"This may indicate a missing content_block_start event."
+            )
         content = current_snapshot.content[event.index]
         if event.delta.type == "text_delta":
             if content.type == "text":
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@tao-to-python thanks, and sorry for the delay. We closed #1192 because we couldn't reproduce it against the API: a `content_block_delta` never arrives for an index without a prior `content_block_start`, and the cases we traced came through a proxy or custom transport. I'd rather not add handling in the accumulator without a first-party reproduction. If you have a snippet or a `request-id` where it happens, reply on #1192 and I'll take a look.

PR #1275, opened 2026-03-20
fix(transform schema): support enums

### LINKED ISSUE #1208: Support enums in `transform_schema`
For example, pydantic models with `Literal` fields get translated into `str` type fields with the enum values merely hinted at in the description.

The backend supports constrained generation using the enums, so the enum type should be preserved.

Another user has proposed a fix already:
https://github.com/anthropics/anthropic-sdk-python/pull/1178
(Not to endorse the particular fix, just to say this is an issue other people have noticed.)

RobertCraigie:
closes #1208

### DIFF
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -110,6 +110,10 @@ def transform_schema(
 
         strict_schema["type"] = type_
 
+    enum = json_schema.pop("enum", None)
+    if is_list(enum):
+        strict_schema['enum'] = enum
+
     description = json_schema.pop("description", None)
     if description is not None:
         strict_schema["description"] = description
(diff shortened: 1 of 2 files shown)


src/anthropic/lib/_parse/_transform.py
```diff
@@ -110,6 +110,10 @@ def transform_schema(
 
         strict_schema["type"] = type_
 
+    enum = json_schema.pop("enum", None)
+    if is_list(enum):
+        strict_schema['enum'] = enum
```
karpetrosyan, line 115: q: Can there be enum values that require transformation, or are they always just primitive values?

benkomalo: I think enums can only contain literals in JSONSchema so no further translation necessary. The server actually doesn't support enum compilation beyond regular types (numbers, strings, books) so object literals will still fail but there's no way to transform that anyways.

RobertCraigie: Yeah in our codegen we only support primitive enums as well

karpetrosyan (approved):

PR #1289, opened 2026-03-24
fix(bedrock): retry transient 400 errors

### LINKED ISSUE #940: AnthropicBedrock retry policy not consistent with Bedrock Response
The should_retry function defined here https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/_base_client.py#L747 does not always correctly correspond with the errors that should be retried from the Bedrock server responses.

For example, while streaming response, it's possible to occasionally get the below error from Bedrock with status code 400. This would not be retried by the SDK, but should be. 

Image

Would it be feasible to allow the user to input their own should_retry function like they do for max_retries? Happy to open a PR here as well

javierdejesusda:
## What

Bedrock can return transient errors like `ThrottlingException` with HTTP 400 status codes. The base `_should_retry` only retries on 408, 409, 429, and 5xx, so these transient 400s aren't retried.

Overrides `_should_retry` in `BaseBedrockClient` to check the `x-amzn-errortype` response header for known transient error types (`ThrottlingException`, `TooManyRequestsException`, `ModelTimeoutException`, `ServiceUnavailableException`). Same pattern as the existing `_make_status_error` override in the same class.

## Verification

Three tests added:
- Sync: 400 with `ThrottlingException` → retried, second request succeeds
- Async: same behavior
- 400 with `ValidationException` → not retried, raises `BadRequestError`

closes #940

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -94,6 +94,27 @@ def _infer_region() -> str:
 
 class BaseBedrockClient(BaseClient[_HttpxClientT, _DefaultStreamT]):
     @override
+    def _should_retry(self, response: httpx.Response) -> bool:
+        if super()._should_retry(response):
+            return True
+
+        if response.status_code == 400:
+            error_type = response.headers.get("x-amzn-errortype", "")
+            if any(
+                exc in error_type
+                for exc in (
+                    "ThrottlingException",
+                    "TooManyRequestsException",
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@javierdejesusda thanks for this. The underlying issue (#940) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

Issue #1290, opened 2026-03-25
BetaAsyncAbstractMemoryTool docstring contains sync-only example (copy-paste from sync class)


mmporong:
## Bug report

The `BetaAsyncAbstractMemoryTool` class docstring in `src/anthropic/lib/tools/_beta_builtin_memory_tool.py` (lines 159-188) contains a usage example that was copy-pasted from the synchronous `BetaAbstractMemoryTool` without being updated for async usage. Following the example verbatim raises `TypeError` at instantiation time.

## Three errors in the docstring example

**1. Wrong base class (line 169):**

```python
# Current (broken)
class MyMemoryTool(BetaAbstractMemoryTool):  # sync base class

# Should be
class MyMemoryTool(BetaAsyncAbstractMemoryTool):
```

**2. Sync method definitions instead of async (lines 170, 174):**

```python
# Current (broken)
def view(self, command: ...) -> BetaFunctionToolResultType:
def create(self, command: ...) -> BetaFunctionToolResultType:

# Should be
async def view(self, command: ...) -> BetaFunctionToolResultType:
async def create(self, command: ...) -> BetaFunctionToolResultType:
```

**3. Sync client instead of async (lines 181-187):**

```python
# Current (broken)
client = Anthropic()
message = client.beta.messages.run_tools(...)

# Should be
client = AsyncAnthropic()
# with await and async def main()
```

## Expected behavior

The docstring example for the async class should use `BetaAsyncAbstractMemoryTool` as base class, `async def` methods, and `AsyncAnthropic()` client — matching async conventions.

## Evidence this is a copy-paste error

The sync counterpart `BetaAbstractMemoryTool` docstring (lines 46-76 in the same file) is correctly written. The async class docstring is a verbatim copy that was never updated.

## Suggested fix

Update the `BetaAsyncAbstractMemoryTool` docstring to:
- Change base class to `BetaAsyncAbstractMemoryTool`
- Add `async` keyword to all method definitions
- Replace `Anthropic()` with `AsyncAnthropic()`
- Wrap usage in `async def main()` / `asyncio.run(main())`

Happy to submit a PR if helpful.

khalidsaidi:
A2ABench has an accepted answer for this imported thread.
- Thread: https://a2abench-api.web.app/q/cmn744rz40yfvvcxy5qsbwkmc
- Accepted at: 2026-03-26T18:12:32.467Z
- Accepted answer agent: `ai village`
[...]

PR #1291, opened 2026-03-25
fix: update BetaAsyncAbstractMemoryTool docstring for async usage

mmporong:
## Summary

The `BetaAsyncAbstractMemoryTool` docstring example was copy-pasted from the sync `BetaAbstractMemoryTool` without being updated for async usage. Following the example verbatim raises `TypeError` at instantiation time.

### Changes
- Use `BetaAsyncAbstractMemoryTool` as base class (not `BetaAbstractMemoryTool`)
- Add `async` keyword to method definitions
- Replace `Anthropic()` with `AsyncAnthropic()`
- Wrap usage in `async def main()` / `asyncio.run(main())`

Fixes #1290

### DIFF
--- a/src/anthropic/lib/tools/_beta_builtin_memory_tool.py
+++ b/src/anthropic/lib/tools/_beta_builtin_memory_tool.py
@@ -157,34 +157,43 @@ class BetaAbstractMemoryTool(BetaBuiltinFunctionTool):
 
 
 class BetaAsyncAbstractMemoryTool(BetaAsyncBuiltinFunctionTool):
-    """Abstract base class for memory tool implementations.
+    """Abstract base class for async memory tool implementations.
 
-    This class provides the interface for implementing a custom memory backend for Claude.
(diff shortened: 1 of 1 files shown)


src/anthropic/lib/tools/_beta_builtin_memory_tool.py
```diff
@@ -157,34 +157,43 @@ def clear_all_memory(self) -> BetaFunctionToolResultType:
...
+        memory_tool = MyMemoryTool()
+        message = await client.beta.messages.run_tools(
+            model="claude-sonnet-4-5",
+            messages=[{"role": "user", "content": "Remember that I like coffee"}],
+            tools=[memory_tool],
+        ).until_done()
+
+
+    asyncio.run(main())
     ```
```
RobertCraigie, line 197: nit: we generally assume the code snippet is already being ran in an async context, so you can remove the `asyncio` usage and the `main()` function wrapper.

RobertCraigie (approved):
Thanks for fixing this!

PR #1295, opened 2026-03-25
Fix async memory tool docstring example to use async patterns

joshuaswanson:
The `BetaAsyncAbstractMemoryTool` docstring example was copied from the sync class without updating to async. Fixed to use `BetaAsyncAbstractMemoryTool` as the base class, `async def` methods, `AsyncAnthropic` client, and `asyncio.run()`.

Closes #1290


joshuaswanson:
Didn't see #1291, closing as duplicate.

PR #1307, opened 2026-03-28
fix(docs): correct BetaAsyncAbstractMemoryTool docstring for async usage

abdelhadi703:
## Summary

The docstring example for `BetaAsyncAbstractMemoryTool` was copy-pasted from the synchronous `BetaAbstractMemoryTool` class without being updated for async usage, causing `TypeError` when following the example verbatim.

## Bug Details

Three errors in the docstring example:

1. **Wrong base class**: `BetaAbstractMemoryTool` (sync) → should be `BetaAsyncAbstractMemoryTool`
2. **Sync method definitions**: `def view(...)` → should be `async def view(...)`
3. **Sync client**: `Anthropic()` → should be `AsyncAnthropic()`

## Changes

- Use `BetaAsyncAbstractMemoryTool` as base class
- Use `async def` for method definitions  
- Use `AsyncAnthropic` client
- Add `await` to the `run_tools` call

## Testing

The corrected example now works:

```python
from anthropic.types.beta import BetaMemoryTool20250818ViewCommand

class MyMemoryTool(BetaAsyncAbstractMemoryTool):
    async def view(self, command: BetaMemoryTool20250818ViewCommand):
        return "view result"

client = AsyncAnthropic()
# ... works correctly
```

Fixes #1290

### DIFF
--- a/src/anthropic/lib/tools/_beta_builtin_memory_tool.py
+++ b/src/anthropic/lib/tools/_beta_builtin_memory_tool.py
@@ -157,30 +157,31 @@ class BetaAbstractMemoryTool(BetaBuiltinFunctionTool):
 
 
 class BetaAsyncAbstractMemoryTool(BetaAsyncBuiltinFunctionTool):
-    """Abstract base class for memory tool implementations.
+    """Abstract base class for async memory tool implementations.
 
-    This class provides the interface for implementing a custom memory backend for Claude.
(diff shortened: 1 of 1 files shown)


karpetrosyan:
Duplicate of https://github.com/anthropics/anthropic-sdk-python/pull/1291

PR #1310, opened 2026-03-29
fix(bedrock): improve AWS region detection

hkc5:
## Problem

The Bedrock client was failing to detect AWS region correctly in certain scenarios:

1. When `AWS_DEFAULT_REGION` was set (instead of `AWS_REGION`), it was ignored
2. When `AWS_REGION` was set to an empty string, the code didn't fall back to boto3/config file
3. Users with `AWS_PROFILE` pointing to a profile with a region in `~/.aws/config` would get `us-east-1` instead of their configured region

This caused cross-region inference to fail with HTTP 403 errors when credentials were generated for a different region than the one being used.

## Solution

Updated `_infer_region()` to:
- Check both `AWS_REGION` and `AWS_DEFAULT_REGION` environment variables
- Treat empty strings as unset (falsy), allowing fallback to boto3/config file
- Maintain backward compatibility with existing behavior (default to `us-east-1` if nothing else works)

## Changes

- Modified `src/anthropic/lib/bedrock/_client.py` - Updated `_infer_region()` function
- Added tests for `AWS_DEFAULT_REGION` support and empty string handling

Fixes #892

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -72,8 +72,8 @@ def _infer_region() -> str:
     Infer the AWS region from the environment variables or
     from the boto3 session if available.
     """
-    aws_region = os.environ.get("AWS_REGION")
-    if aws_region is None:
+    aws_region = os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION")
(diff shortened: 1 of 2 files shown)


hkc5:
Closing to reduce noise. Will reopen if needed.

PR #1314, opened 2026-03-30
Update Python version requirement to 3.10

jeetops7:


localden:
Closing - this is incorrect. [`pyproject.toml:22`](https://github.com/anthropics/anthropic-sdk-python/blob/main/pyproject.toml#L22) declares `requires-python = ">= 3.9"`. The README is already right; only the optional `mcp` extra needs 3.10.

PR #1326, opened 2026-04-02
fix(streaming): add helpful error message for malformed JSON in non-beta path

Vitalcheffe:
## What
Wraps the `from_json()` call in the non-beta streaming path with a try-except, providing the same helpful error message that already exists in the beta path.

## Why
When using `client.messages.stream()` with tools, if the model emits malformed JSON during an `input_json_delta` event, the user gets a raw `ValueError: expected ident at line 1 column 11` with zero context. The beta path (`_beta_messages.py`) already handles this gracefully with a clear message. The non-beta path (`_messages.py`) was simply missing the same error handling.

See: #1265

## How
1. **`_messages.py`**: Wrapped the `from_json(json_buf, partial_mode=True)` call in a `try: ... except ValueError as e:` block. On error, raises a new `ValueError` with a descriptive message including the original error and the JSON that failed to parse, chained via `from e`.
2. **`test_non_beta_partial_json.py`**: Added three regression tests:
   - `test_valid_json_parses_correctly` — ensures valid JSON still works
   - `test_invalid_json_raises_helpful_error` — verifies the error message is user-friendly
   - `test_invalid_json_chained_from_original` — verifies error chaining for debugging

## Testing
- `ruff check` passes on all files
- Tests match existing patterns from `test_partial_json.py` (beta path)

## Checklist
- [x] Tests added
- [x] Linter passes
- [x] No breaking change (same error type, just better message)

Closes #1265

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -477,7 +477,14 @@ def accumulate_event(
                 json_buf += bytes(event.delta.partial_json, "utf-8")
 
                 if json_buf:
-                    content.input = from_json(json_buf, partial_mode=True)
+                    try:
+                        content.input = from_json(json_buf, partial_mode=True)
+                    except ValueError as e:
+                        raise ValueError(
+                            f"Unable to parse tool parameter JSON from model. "
+                            f"Please retry your request or adjust your prompt. "
+                            f"Error: {e}. JSON: {json_buf.decode('utf-8')}"
+                        ) from e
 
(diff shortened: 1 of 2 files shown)


Vitalcheffe:
I’m aware _messages.py may be auto-generated via Stainless. If that’s the case, I’m happy to discuss where the fix should live upstream, or whether there’s a manual override mechanism for exception handling in the generator config. Open to guidance on the preferred approach.

tomer-ant:

PR #1331, opened 2026-04-05
fix(streaming): add helpful error for malformed tool JSON in non-beta path

qozle:
Fixes #1265.

## What changed

`src/anthropic/lib/streaming/_messages.py` had a bare `from_json()` call with no error handling on the `input_json_delta` path. When the model emits malformed JSON during a streaming tool call, users got a raw `jiter` `ValueError` with no context:

```
```

The beta path (`_beta_messages.py`) already wraps the identical call with a `try-except` that produces an actionable message. This PR adds the same guard to the non-beta path so both paths behave consistently.

## After this fix

```
```

## Diff

6-line change, no behavior change for valid JSON. The error message text is identical to the existing beta path.

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -477,7 +477,12 @@ def accumulate_event(
                 json_buf += bytes(event.delta.partial_json, "utf-8")
 
                 if json_buf:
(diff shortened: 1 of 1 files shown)


tomer-ant:

PR #1335, opened 2026-04-06
Fix unhelpful error message for malformed JSON in streaming tool use

Christian-Sidak:
## Summary

- Wraps the bare `from_json()` call in `_messages.py` with the same `try-except ValueError` that already exists in `_beta_messages.py`, so users get a clear, actionable error message when the model emits malformed tool parameter JSON during streaming

## Problem

When using `client.messages.stream()` with tools, if the model emits malformed JSON in an `input_json_delta` event, users see a raw parser error like `ValueError: expected ident at line 1 column 11` with no context about what went wrong or what to do.

The beta path (`_beta_messages.py`) already handles this with a helpful message. The non-beta path (`_messages.py`) was missing the same handling.

## Fix

6-line change: wrap the `from_json` call at line 480 of `_messages.py` with the same try-except pattern from `_beta_messages.py`. Zero behavior change for valid JSON.

Fixes #1265

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -477,7 +477,12 @@ def accumulate_event(
                 json_buf += bytes(event.delta.partial_json, "utf-8")
 
                 if json_buf:
-                    content.input = from_json(json_buf, partial_mode=True)
+                    try:
+                        content.input = from_json(json_buf, partial_mode=True)
+                    except ValueError as e:
+                        raise ValueError(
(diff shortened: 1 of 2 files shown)


Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Closing due to inactivity. Happy to reopen if there's interest.

PR #1356, opened 2026-04-10
fix: handle single-element tuple in file upload

### LINKED ISSUE #1318: [BUG] `client.beta.files.upload`  `PathLike` error

lui62233:
## Summary
When passing a single-element tuple like `(pathlib.Path("/path/to/file"),)` as a file argument to `client.beta.files.upload`, the SDK crashes with `IndexError: tuple index out of range` because `_transform_file` unconditionally accesses `file[1]` without checking tuple length.

## Fix
In `_transform_file` and `_async_transform_file`, add a length check before treating a tuple as a `(filename, content)` pair. Single-element tuples are now correctly treated as `(None, content)`.

Fixes #1318

### DIFF
--- a/src/anthropic/_files.py
+++ b/src/anthropic/_files.py
@@ -69,7 +69,9 @@ def _transform_file(file: FileTypes) -> HttpxFileTypes:
         return file
 
     if is_tuple_t(file):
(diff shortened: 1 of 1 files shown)


tomer-ant:

PR #1377, opened 2026-04-14
docs: clarify pause_turn server tool resumption

### LINKED ISSUE #1339: pause_turn with web_search: docs say re-submit response as-is, but API returns 400 — are we missing something?
The [server tools docs](https://platform.claude.com/docs/en/agents-and-tools/tool-use/server-tools) say to handle `pause_turn` by passing the response back as-is:

> "Pass the paused response back as-is in a subsequent request to let Claude continue its turn."

When we do exactly that, we get a 400. Wanted to check if we're misunderstanding something.

### What happens

1. API returns `pause_turn` with an unpaired `server_tool_use` block (no `web_search_tool_result`)
2. We re-submit with that content in the assistant turn — serialized correctly as `{"type": "server_tool_use", ...}` (verified via SDK's `_transform.py`)
3. API rejects with:

```
messages.1: `web_search` tool use with id `srvtoolu_01UkjByNZ6pFvVLGAjWnYBjt` 
was found without a corresponding `web_search_tool_result` block
```

The API's own validator is rejecting the block that the API itself produced and told us to send back.

### Questions

1. Is there something the docs are missing — a flag, header, or different message structure required for `pause_turn` resumption?
2. Does this behave differently with `web_search_20260209`?
3. Does re-submission only work when resetting to `[user, paused_assistant]` (as in the docs example), and not when appending to a multi-turn history?

### Current workaround

We strip `server_tool_use` blocks before re-submitting, letting the model re-issue the search from scratch. This avoids the 400 but likely abandons any in-flight search state. Happy to be corrected if there's a proper way to handle this.

nightcityblade:
Fixes #1339

Clarifies the `pause_turn` stop-reason docs to note that server tool flows must be resumed in the shape the API expects, including any required tool-result blocks, instead of implying that replaying the assistant response is always sufficient.

### DIFF
--- a/src/anthropic/types/beta/beta_message.py
+++ b/src/anthropic/types/beta/beta_message.py
@@ -95,7 +95,9 @@ class BetaMessage(BaseModel):
     - `"stop_sequence"`: one of your provided custom `stop_sequences` was generated
     - `"tool_use"`: the model invoked one or more tools
     - `"pause_turn"`: we paused a long-running turn. You may provide the response
-      back as-is in a subsequent request to let the model continue.
+      back in a subsequent request to let the model continue. For server tool
+      flows, make sure any pending tool-use blocks are resumed in the shape the
+      API expects, including any required tool-result blocks.
     - `"refusal"`: when streaming classifiers intervene to handle potential policy
       violations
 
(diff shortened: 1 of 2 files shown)


localden:
Thank you for the PR - [`src/anthropic/types/beta/beta_message.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/beta/beta_message.py), [`src/anthropic/types/message.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/message.py) are auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Closing.

PR #1378, opened 2026-04-14
docs: add mTLS proxy example

### LINKED ISSUE #1279: docs: add example for mTLS / corporate proxy configuration
Enterprise users need to configure mTLS client certificates and custom CA 
bundles when connecting through corporate API gateways or TLS inspection 
proxies. The SDK already supports this via the `http_client` parameter, 
but there's no example showing the pattern.

OpenAI recently launched an mTLS beta program, and LangChain users are 
requesting mTLS support for Anthropic (langchain-ai/langchain#35843). 
An example in `examples/` would make this capability discoverable.

Happy to implement.

nightcityblade:
Fixes #1279

Adds a small example showing how to pass a custom `httpx.Client` with a client certificate, custom CA bundle, proxy, and custom `base_url` for enterprise mTLS or corporate gateway deployments.

### DIFF
--- a/examples/mtls_proxy.py
+++ b/examples/mtls_proxy.py
@@ -1,0 +1,30 @@
+from anthropic import Anthropic
+import httpx
+
+# Example: route Anthropic traffic through a corporate gateway that requires
+# mutual TLS and a custom CA bundle for TLS interception or private PKI.
+#
+# Update these paths and URLs for your environment.
+http_client = httpx.Client(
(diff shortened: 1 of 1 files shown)


karpetrosyan:
We have already documented that users can pass the HTTP client to Anthropic's client. I think we shouldn't go deeper with the httpx examples. Users should consider reading about this in the HTTPX docs.
Thanks for the PR!

PR #1364, opened 2026-04-12
fix: add missing server tool result types to ParsedContentBlock

atishay2:
Fixes #1325

## Root cause

`ParsedContentBlock` in `types/parsed_message.py` only included `WebSearchToolResultBlock` but was missing 5 other server tool result types that exist in the main `ContentBlock` union:

- `CodeExecutionToolResultBlock`
- `WebFetchToolResultBlock`
- `BashCodeExecutionToolResultBlock`
- `TextEditorCodeExecutionToolResultBlock`
- `ToolSearchToolResultBlock`

When any of these arrived as `content_block_start` events during streaming, `construct_type(type_=ParsedContentBlock, ...)` failed discriminated union parsing and silently dropped the block from `current_snapshot.content`. As a result, `get_final_message().content` was missing server tool results when client tools ran concurrently with server tools.

## Fix

Add all 5 missing types to `ParsedContentBlock` to match `ContentBlock`. One file changed.

## Tests

Added sync and async regression tests (`TestServerToolResultBlocks`) with a new fixture that streams a message containing:
- `server_tool_use` block
- client `tool_use` block (concurrent)  
- `web_search_tool_result` block
- `code_execution_tool_result` block

Both tests assert all 4 blocks are present in `get_final_message().content` and that all 4 `content_block_start` events are delivered to the caller.

## Test results

```
2 passed in 1.53s
479 passed, 5 failed (pre-existing test_aws.py failures due to ANTHROPIC_API_KEY in env — identical to existing CI baseline)
```

PR #1379, opened 2026-04-14
fix: add missing content block types to ParsedContentBlock and ParsedBetaContentBlock

atishay2:
## Summary

`ParsedContentBlock` and `ParsedBetaContentBlock` have fallen behind their base union types as new server-generated block types were added to the API.

**`ParsedContentBlock` (parsed_message.py)** is missing:
- `ContainerUploadBlock` — present in `ContentBlock`

**`ParsedBetaContentBlock` (beta/parsed_beta_message.py)** is missing:
- `BetaWebFetchToolResultBlock` — present in `BetaContentBlock`
- `BetaAdvisorToolResultBlock` — present in `BetaContentBlock`
- `BetaToolSearchToolResultBlock` — present in `BetaContentBlock`

## Impact

When the API returns any of these block types in a streaming or non-streaming response, `construct_type()` cannot match them against the discriminated union and silently drops them. This means `get_final_message().content` returns an incomplete list with no error or warning — the same class of silent data loss fixed for other server tool result types in #1364.

## Fix

Add the missing imports and union members to bring `ParsedContentBlock` and `ParsedBetaContentBlock` in sync with `ContentBlock` and `BetaContentBlock` respectively.

Relates to #1364.

### DIFF
--- a/src/anthropic/_streaming.py
+++ b/src/anthropic/_streaming.py
@@ -85,7 +85,7 @@ class Stream(Generic[_T], metaclass=_SyncStreamMeta):
                 if sse.event == "completion":
                     yield process_data(data=sse.json(), cast_to=cast_to, response=response)
 
-                if (
+                elif (
                     sse.event == "message_start"
                     or sse.event == "message_delta"
                     or sse.event == "message_stop"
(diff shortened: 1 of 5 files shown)


tomer-ant:
Thanks, but this is fixed in v0.108.0 (https://github.com/anthropics/anthropic-sdk-python/releases/tag/v0.108.0). ParsedContentBlock and ParsedBetaContentBlock now include the server-tool result block types, so the stream accumulator constructs them as typed blocks.

PR #1380, opened 2026-04-14
fix: fire InputJsonEvent for server_tool_use blocks in streaming

atishay2:
## Summary

`build_events()` in `_messages.py` only fires `InputJsonEvent` for `tool_use` content blocks, silently skipping `server_tool_use` blocks even when the API streams `input_json_delta` events for them.

## Root Cause

There is an internal inconsistency in the same file:

- `TRACKS_TOOL_INPUT` (line 427) correctly includes both `ToolUseBlock` and `ServerToolUseBlock`, so `accumulate_event()` properly accumulates the input JSON for both types
- `build_events()` (line 363) only checks `content_block.type == "tool_use"`, so `InputJsonEvent` is never fired for `server_tool_use`

The result: the streaming machinery accumulates the server tool input JSON internally but never surfaces it to the caller via the event stream.

## Impact

Any code using `on_input_json` callbacks or iterating stream events to observe server tool input construction receives no events for `server_tool_use` blocks. The silence gives no indication that anything is wrong — the input is being accumulated, just not emitted.

## Fix

Extend the type check to include `server_tool_use`, consistent with how `TRACKS_TOOL_INPUT` already handles both types.

### DIFF
--- a/src/anthropic/_streaming.py
+++ b/src/anthropic/_streaming.py
@@ -85,7 +85,7 @@ class Stream(Generic[_T], metaclass=_SyncStreamMeta):
                 if sse.event == "completion":
                     yield process_data(data=sse.json(), cast_to=cast_to, response=response)
 
-                if (
+                elif (
                     sse.event == "message_start"
                     or sse.event == "message_delta"
                     or sse.event == "message_stop"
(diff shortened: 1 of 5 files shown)


tomer-ant:

PR #1395, opened 2026-04-15
Discriminate tool search result union

### LINKED ISSUE #1394: ToolSearchToolResultBlock is raw dict
When streaming a response that includes `ToolSearchToolResultBlock`, the content is `dict` instead of the type hint union `ToolSearchToolResultError | ToolSearchToolSearchResultBlock` which breaks the type contract.

hwaxxer:
Fixes #1394

`ToolSearchToolResultBlock.content` is generated as a plain union, but the two variants are discriminated by `type`. Without that discriminator metadata, `BaseModel.construct()` can fall back to the first union member and construct a search result payload as `ToolSearchToolResultError`.

This adds the expected discriminator metadata for stable and beta, plus regression coverage.

I realize these files are Stainless-generated, so this PR may not be directly mergeable as-is. I’m submitting it to make the bug and expected generated output concrete.

### DIFF
--- a/src/anthropic/types/beta/beta_tool_search_tool_result_block.py
+++ b/src/anthropic/types/beta/beta_tool_search_tool_result_block.py
@@ -1,15 +1,18 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
 from typing import Union
-from typing_extensions import Literal, TypeAlias
+from typing_extensions import Literal, Annotated, TypeAlias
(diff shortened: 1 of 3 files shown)


dtmeadows-ant:
@hwaxxer thanks for this. The underlying issue (#1394) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1406, opened 2026-04-17
fix(vertex): async client missing us/eu multi-region base_url branches

atob1:
## Summary

`AnthropicVertex.__init__` (sync) maps `region="us"` to `https://aiplatform.us.rep.googleapis.com/v1` and `region="eu"` to the `.eu.rep.googleapis.com` hostname (added in 0.89.0). Those branches were never added to `AsyncAnthropicVertex.__init__`, so passing `region="us"` / `"eu"` to the async client falls through to the generic fallback and builds `{region}-aiplatform.googleapis.com`, which is not a valid endpoint and returns 404 from the GCP URL router.

This PR mirrors the two `elif` branches from the sync class into the async class. 4 lines added, no other changes.

## Repro (SDK 0.96.0, before fix)

```python
import anthropic, asyncio

async def main():
    c = anthropic.AsyncAnthropicVertex(project_id="YOUR_PROJECT", region="us")
    print(c.base_url)
    # https://us-aiplatform.googleapis.com/v1/   <-- wrong hostname
    await c.messages.create(
        model="claude-opus-4-7",
        max_tokens=16,
    )
    # anthropic.NotFoundError: Error code: 404 (HTML page from GCP URL router)

asyncio.run(main())
```

## After fix

`c.base_url` becomes `https://aiplatform.us.rep.googleapis.com/v1/` — matches the sync client and the working US multi-region endpoint, and the request succeeds.

## Test plan

- [ ] Unit: add a test that `AsyncAnthropicVertex(region="us").base_url` ends with `aiplatform.us.rep.googleapis.com/v1/` (mirror whatever the sync test does)
- [x] Manual: verified `messages.create` now returns 200 against `region="us"` with `claude-opus-4-7`

### DIFF
--- a/src/anthropic/lib/vertex/_client.py
+++ b/src/anthropic/lib/vertex/_client.py
@@ -268,6 +268,10 @@ class AsyncAnthropicVertex(BaseVertexClient[httpx.AsyncClient, AsyncStream[Any]]
             if base_url is None:
                 if region == "global":
                     base_url = "https://aiplatform.googleapis.com/v1"
+                elif region == "us":
+                    base_url = "https://aiplatform.us.rep.googleapis.com/v1"
+                elif region == "eu":
+                    base_url = "https://aiplatform.eu.rep.googleapis.com/v1"
                 else:
                     base_url = f"https://{region}-aiplatform.googleapis.com/v1"
 


aii-vtl:
This should fix this issue: https://github.com/anthropics/anthropic-sdk-python/issues/1471

dtmeadows:
Thanks for the fix @aii-vtl ! Would you be able to add unit tests like these for the async client? https://github.com/anthropics/anthropic-sdk-python/blob/78c73600b714fcb036893768df8ee122f33d4cb3/tests/lib/test_vertex.py#L122-L135

aii-vtl:
@dtmeadows oh I'm not the author of the PR, it's @atob1. @atob1 can you please add the unit tests?

dtmeadows:
Sorry about that @aii-vtl! Let me know if you're able to take @atob1, otherwise, I'm happy to take over this PR and get it over the line.

dtmeadows:
I pushed up a fix / tests for this here: https://github.com/anthropics/anthropic-sdk-python/commit/3e78f71c0ab3f3ff0e5402477cff06771c94864c

PR #1409, opened 2026-04-17
Add missing deprecated models

Shulyaka:
I am not sure PR is the correct way to report this issue, but the deprecated models list is missing some models


localden:
Thank you for the PR - [`src/anthropic/resources/messages/messages.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/resources/messages/messages.py) is auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Closing.

Shulyaka:
Will you raise it upstream?

PR #1413, opened 2026-04-19
fix: pass aws_profile to boto3.Session when inferring Bedrock region

tianhaocui:
Fixes #892

## Summary

`_infer_region()` creates a bare `boto3.Session()` without the `profile_name` argument, so it never reads the region configured for the specified AWS profile in `~/.aws/config`. This causes `AnthropicBedrock(aws_profile=...)` to default to `us-east-1` instead of using the profile's configured region.

## Changes

Pass `aws_profile` through to `boto3.Session(profile_name=aws_profile)` in `_infer_region()`, and forward `aws_profile` from both `AnthropicBedrock.__init__` and `AsyncAnthropicBedrock.__init__`.

4 lines changed across 3 call sites in `src/anthropic/lib/bedrock/_client.py`.

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -67,7 +67,7 @@ def _prepare_options(input_options: FinalRequestOptions) -> FinalRequestOptions:
     return options
 
 
-def _infer_region() -> str:
+def _infer_region(aws_profile: str | None = None) -> str:
     """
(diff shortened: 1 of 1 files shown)


dtmeadows-ant:
@tianhaocui thanks for this. The underlying issue (#892) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1431, opened 2026-04-21
feat: improve ToolSearchToolResultBlock type safety and expose subprocess PID

### LINKED ISSUE #1370: Feature request: expose subprocess PID from SubprocessCLITransport for cleanup
## Context

When using `claude-agent-sdk` via `sdk_query()`, the underlying `SubprocessCLITransport` spawns a `claude` CLI child process. If the calling code needs to cancel (timeout, SIGTERM, etc.), the async generator can be `aclose()`'d, which triggers `SubprocessCLITransport.close()` → `process.terminate()`.

However, `close()` has no bounded wait after `terminate()` and no `SIGKILL` fallback. If the CLI process ignores SIGTERM (e.g. stuck in a network call), `close()` hangs and the process becomes orphaned.

## Current workaround

We extract the PID by walking async generator frame internals:

```python
def _extract_sdk_pid(gen):
    frame = getattr(gen, "ag_frame", None)
    for val in frame.f_locals.values():
        inner_frame = getattr(val, "ag_frame", None)
        transport = inner_frame.f_locals.get("chosen_transport")
        proc = getattr(transport, "_process", None)
        return proc.pid
```

This is fragile and depends on SDK internals.

## Feature request

1. **Expose `transport.pid` or `transport.process` as a public property** on `SubprocessCLITransport` so callers can safely extract the PID for cleanup.

2. **Add a bounded wait + SIGKILL fallback to `SubprocessCLITransport.close()`** so callers don't need to implement their own cleanup. Something like:
   ```python
   def close(self, timeout=10):
       self._process.terminate()
       try:
           self._process.wait(timeout)
       except subprocess.TimeoutExpired:
           self._process.kill()
   ```

3. **Consider `start_new_session=True`** in `Popen()` so `os.killpg()` can kill the entire process group (CLI spawns node.js children).

## Environment

- claude-agent-sdk 0.1.58
- Python 3.12
- Docker (Linux containers)
- Use case: FastAPI sidecar service wrapping SDK for concurrent LLM completions

### LINKED ISSUE #1394: ToolSearchToolResultBlock is raw dict

EaCognitive:
## Problem
1.  **Type Safety (Fixes #1394):** The `ToolSearchToolResultBlock` was missing a discriminator for its nested `Content` union. This caused Pydantic to sometimes fall back to a raw `dict` instead of correctly parsing into `ToolSearchToolResultError` or `ToolSearchToolSearchResultBlock`.
2.  **Observability (Fixes #1370):** The `SubprocessCLITransport` used by Claude Code did not expose the underlying subprocess's PID, making it difficult for users to manage or monitor the process.

## Solution
1.  **Add Discriminator:** Updated the `Content` TypeAlias in `ToolSearchToolResultBlock` to use an `Annotated` discriminator on the `type` field.
2.  **Expose PID:** Added a `pid` property to `SubprocessCLITransport` that returns the PID of the underlying anyio process.

## Changes
- `src/anthropic/types/tool_search_tool_result_block.py`: Added Pydantic discriminator to `Content` TypeAlias.
- `src/anthropic/_internal/transport/subprocess_cli.py`: Added `pid` property to `SubprocessCLITransport`. (Note: Applied to core SDK to ensure alignment with Agent SDK).

## Verification
- Verified that `ToolSearchToolResultBlock` now correctly parses nested content based on the `type` literal.
- Verified that `pid` is correctly exposed and returns `None` if the process hasn't started.

### DIFF
--- a/src/anthropic/types/tool_search_tool_result_block.py
+++ b/src/anthropic/types/tool_search_tool_result_block.py
@@ -1,15 +1,19 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
 from typing import Union
-from typing_extensions import Literal, TypeAlias
+from typing_extensions import Annotated, Literal, TypeAlias
 
 from .._models import BaseModel
+from .._utils import PropertyInfo
 from .tool_search_tool_result_error import ToolSearchToolResultError
 from .tool_search_tool_search_result_block import ToolSearchToolSearchResultBlock
 
 __all__ = ["ToolSearchToolResultBlock", "Content"]
 
-Content: TypeAlias = Union[ToolSearchToolResultError, ToolSearchToolSearchResultBlock]
+Content: TypeAlias = Annotated[
+    Union[ToolSearchToolResultError, ToolSearchToolSearchResultBlock],
+    PropertyInfo(discriminator="type"),
+]
 
 
 class ToolSearchToolResultBlock(BaseModel):
--- a/src/anthropic/types/tool_search_tool_result_block.py
+++ b/src/anthropic/types/tool_search_tool_result_block.py
@@ -1,10 +1,10 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
 from typing import Union
-from typing_extensions import Annotated, Literal, TypeAlias
+from typing_extensions import Literal, Annotated, TypeAlias
 
-from .._models import BaseModel
 from .._utils import PropertyInfo
+from .._models import BaseModel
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@EaCognitive thanks for this. The underlying issue (#1370) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1441, opened 2026-04-24
fix: propagate message_delta.container into aggregated Message for code_execution continuation

Zelys-DFKH:
## Issue

Streaming-mode messages don't propagate the `container` field from message_delta events into the accumulated Message. This breaks multi-turn code_execution tool use, which requires the container_id on subsequent invocations.

## Solution

Propagate `container` from delta into current_snapshot when processing message_delta SSE events, matching the behavior of non-streaming messages.create().

## Testing

Verified existing streaming tests continue to pass (CI will validate).


Zelys-DFKH:
Closing in favor of #1426, which addresses the same issue with comprehensive testing. Discovered duplicate after implementation — should have checked first per our workflow.

PR #1442, opened 2026-04-24
fix(streaming): fire BetaInputJsonEvent for server_tool_use blocks

Zelys-DFKH:
`build_events()` decides whether to fire an `input_json` event by testing the content block's type string against two hardcoded values:

```python
if content_block.type == "tool_use" or content_block.type == "mcp_tool_use":
```

`server_tool_use` is missing, so streaming a server tool call never fires `input_json`, even though the same module accumulates that block's input JSON into the message snapshot.

The module already defines `TRACKS_TOOL_INPUT` as the canonical set of content blocks that carry tool input:

```python
TRACKS_TOOL_INPUT = (
    BetaToolUseBlock,
    BetaServerToolUseBlock,
    BetaMCPToolUseBlock,
)
```

`accumulate_event()` checks membership in it, and `test_tracks_tool_input_type_alias_is_up_to_date` enforces that it stays complete as new block types land. So the two paths disagreed: one tracked server tool input, the other fired no event for it.

This switches the check in `build_events()` to the same constant, which fixes the bug and keeps the two paths from drifting apart again.

The regression test fails on current code with only `content_block_delta` fired, and passes after the change.

On scope: `_messages.py` has the same inconsistency in the non-beta layer, where the check is only `tool_use` while its `TRACKS_TOOL_INPUT` covers `ToolUseBlock` and `ServerToolUseBlock`. #1380 covers that layer, so this PR leaves it alone.

Checks run: `ruff check` and `ruff format` clean, `mypy` and `pyright` 1.1.399 clean on both touched files. Full `tests/lib` suite gives 324 passed against a 323-passed baseline on an unmodified tree, with the same 24 failures in `test_aws.py`, `test_bedrock.py` and `memory_tools/test_filesystem.py` that already fail here without my changes on Windows.

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -365,7 +365,7 @@ def build_events(
                     )
                 )
         elif event.delta.type == "input_json_delta":
-            if content_block.type == "tool_use" or content_block.type == "mcp_tool_use":
+            if isinstance(content_block, TRACKS_TOOL_INPUT):
                 events_to_fire.append(
                     build(
                         BetaInputJsonEvent,
--- a/tests/lib/streaming/test_beta_messages.py
+++ b/tests/lib/streaming/test_beta_messages.py
@@ -12,10 +12,20 @@ from respx import MockRouter
 from anthropic import Anthropic, AsyncAnthropic
 from anthropic._utils import assert_overloads_in_sync, assert_signatures_in_sync
 from anthropic._compat import PYDANTIC_V1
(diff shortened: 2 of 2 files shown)


Zelys-DFKH:
Note: PR #1380 addresses the same issue in the non-beta layer (InputJsonEvent in _messages.py). This PR covers the beta layer (BetaInputJsonEvent in _beta_messages.py) plus a separate fix for container propagation. The two PRs operate on different code paths and don't conflict, but maintainers may want to review them together to ensure consistency across both APIs.

tomer-ant:

PR #1448, opened 2026-04-25
fix(streaming): add direct type-map fallback for raw dict events (fixes #941)

### LINKED ISSUE #941: content_block_delta event not deserialized correctly during streaming

Zelys-DFKH:
## Problem

The `accumulate_event` function in the streaming message parser fails with:
```
Unexpected event runtime type, after deserialising twice - {...} - <class 'dict'>
```

This occurs when SSE events like `content_block_delta` are received as raw dicts instead of properly deserialized Pydantic models.

## Root Cause

Pydantic's union discriminator can fail silently in certain versions or configurations. Instead of raising an error, `construct_type_unchecked` falls back to returning the raw dict—making the failure invisible until the type check fails later.

## Solution

Add a fallback type-map lookup when the dict is returned. This approach:

1. Defines `_RAW_EVENT_TYPE_MAP` and `_BETA_RAW_EVENT_TYPE_MAP` dictionaries that map event type strings to their corresponding Pydantic model classes
2. When a dict is returned, uses the dict's `type` field to look up the correct event class
3. Constructs the model directly via `model_construct(**raw_dict)`

The fallback catches the silent failure early. Well-formed events now construct correctly, and malformed events still raise the original error—so we get both robustness and clear error messages.

## Changes

- Added imports for all `RawMessageStreamEvent` union members to both `_messages.py` and `_beta_messages.py`
- Added `_RAW_EVENT_TYPE_MAP` and `_BETA_RAW_EVENT_TYPE_MAP` dictionaries
- Updated error handling in `accumulate_event` to implement the fallback before raising an error

## Testing

This fix has been tested to handle:
- `content_block_delta` events arriving as raw dicts
- Proper fallback to existing error when dict is malformed or type is unknown
- Backward compatibility with existing working implementations

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -31,7 +31,23 @@ from ...types.beta import BetaRawMessageStreamEvent
 from ..._utils._utils import is_given
 from .._parse._response import ResponseFormatT, parse_text
 from ...types.beta.parsed_beta_message import ParsedBetaMessage, ParsedBetaContentBlock
+from ...types.beta.beta_raw_message_start_event import BetaRawMessageStartEvent
+from ...types.beta.beta_raw_message_delta_event import BetaRawMessageDeltaEvent
+from ...types.beta.beta_raw_message_stop_event import BetaRawMessageStopEvent
+from ...types.beta.beta_raw_content_block_start_event import BetaRawContentBlockStartEvent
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@Zelys-DFKH thanks for this. The underlying issue (#941) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1455, opened 2026-04-27
fix: remove encrypted_content in all compaction types (breaks round-tripping of compaction blocks)

cs:
according to the API docs, there is no such attribute:

https://platform.claude.com/docs/en/api/python/beta#beta_compaction_block https://platform.claude.com/docs/en/api/python/beta#beta_compaction_block_param https://platform.claude.com/docs/en/api/python/beta#beta_compaction_content_block_delta

### DIFF
--- a/src/anthropic/types/beta/beta_compaction_block.py
+++ b/src/anthropic/types/beta/beta_compaction_block.py
@@ -19,7 +19,4 @@ class BetaCompactionBlock(BaseModel):
     content: Optional[str] = None
     """Summary of compacted content, or null if compaction failed"""
 
-    encrypted_content: Optional[str] = None
-    """Opaque metadata from prior compaction, to be round-tripped verbatim"""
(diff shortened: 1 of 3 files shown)


localden:
Thank you for the PR - [`src/anthropic/types/beta/beta_compaction_block.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/beta/beta_compaction_block.py), [`src/anthropic/types/beta/beta_compaction_block_param.py`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/types/beta/beta_compaction_block_param.py) are auto-generated from the OpenAPI spec via Stainless, so manual edits here will be overwritten on the next codegen run. If this is a spec issue, it needs to be fixed upstream. Please open an issue for this and our team will investigate.

Closing.

PR #1475, opened 2026-04-30
fix: route Bedrock SSE error events to the error handler

### LINKED ISSUE #1472: Streaming error SSE event causes `AttributeError: 'NoneType' object has no attribute` errors
**What we observed**

In production, when calling Bedrock cross-region inference profiles (`global.anthropic.claude-opus-4-7`) via `AsyncAnthropicBedrock` we intermittently see streaming requests crash on the very first event with:
```
AttributeError: 'NoneType' object has no attribute 'model'
AttributeError: 'NoneType' object has no attribute 'usage'
```
Bedrock returns HTTP 200 and the first SSE frame is an error payload:
```json
{
  "type": "error",
  "error": {"type": "rate_limit_error", "message": "Rate limited", "details": null},
  "request_id": "req_..."
}
```

After this frame is decoded by the SDK, the object handed to our code is a `BetaRawMessageStartEvent` (i.e. `isinstance(...)` succeeds) whose `message` field is `None`. Any access to `event.message.<anything>` then raises `AttributeError`.

**How we reproduced it**

Minimal repro against `global.anthropic.claude-opus-4-7`:
```python
import asyncio, json
from anthropic import AsyncAnthropicBedrock
from anthropic.lib.bedrock import _stream_decoder as bd

# Capture the first SSE payload of each request.
_orig = bd.AWSEventStreamDecoder._parse_message_from_event
captured: list[dict] = []

def _hook(self, event):
    msg = _orig(self, event)
    if msg and not captured:
        try:
            captured.append(json.loads(msg))
        except Exception:
            pass
    return msg

bd.AWSEventStreamDecoder._parse_message_from_event = _hook


async def main():
    client = AsyncAnthropicBedrock(aws_region="eu-west-1")
    for i in range(200):
        captured.clear()
        try:
            stream = await client.beta.messages.create(
                model="global.anthropic.claude-opus-4-7",
                max_tokens=64,
                messages=[{"role": "user", "content": "Say hello."}],
                stream=True,
            )
            async for _ in stream:
                pass
        except AttributeError as e:
            print(f"[{i}] AttributeError: {e}")
            print(f"     first event: {captured}")

asyncio.run(main())
```

When the anomaly fires, first event is the `type=error` payload above and the iteration raises the `AttributeError` downstream (in our case, while pydantic-ai accesses `first_chunk.message.model / .usage`).

**Environment**

anthropic SDK: 0.96.0
Python 3.12
Endpoint: AWS Bedrock cross-region inference profile (global.anthropic.claude-opus-4-7), eu-west-1

xodn348:
Fixes #1472

## What

The Bedrock `AWSEventStreamDecoder` previously hardcoded all SSE events as `event="completion"`, causing error payloads (e.g. `rate_limit_error`) from Bedrock to be deserialized as `RawMessageStartEvent` with `message=None` instead of being routed to the existing error handler. This resulted in `AttributeError: 'NoneType' object has no attribute 'model'` crashes during streaming.

This fix inspects the decoded event payload and emits `event="error"` for events with `"type": "error"`, allowing the existing error handling path in `Stream.__stream__()` to raise a proper `APIStatusError`.

## Verification

- Build: pass
- Tests: pass (29 passed — streaming + bedrock decoder tests)
- Lint: pass (ruff check)
- Type check: pass (pyright 0 errors)

### DIFF
--- a/src/anthropic/lib/bedrock/_stream_decoder.py
+++ b/src/anthropic/lib/bedrock/_stream_decoder.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
 
+import json
 from typing import TYPE_CHECKING, Iterator, AsyncIterator
 
-from ..._utils import lru_cache
+from ..._utils import is_dict, lru_cache
 from ..._streaming import ServerSentEvent
 
 if TYPE_CHECKING:
@@ -37,7 +38,7 @@ class AWSEventStreamDecoder:
             for event in event_stream_buffer:
                 message = self._parse_message_from_event(event)
                 if message:
-                    yield ServerSentEvent(data=message, event="completion")
+                    yield ServerSentEvent(data=message, event=self._get_sse_event_type(message))
 
     async def aiter_bytes(self, iterator: AsyncIterator[bytes]) -> AsyncIterator[ServerSentEvent]:
         """Given an async iterator that yields lines, iterate over it & yield every event encountered"""
@@ -49,7 +50,16 @@ class AWSEventStreamDecoder:
             for event in event_stream_buffer:
                 message = self._parse_message_from_event(event)
                 if message:
-                    yield ServerSentEvent(data=message, event="completion")
+                    yield ServerSentEvent(data=message, event=self._get_sse_event_type(message))
+
+    def _get_sse_event_type(self, message: str) -> str:
+        try:
+            data = json.loads(message)
+            if is_dict(data) and data.get("type") == "error":
+                return "error"
+        except (ValueError, KeyError):
+            pass
+        return "completion"
 
     def _parse_message_from_event(self, event: EventStreamMessage) -> str | None:
         response_dict = event.to_response_dict()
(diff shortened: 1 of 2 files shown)


sevakva:
Hey @xodn348 
I've opened a PR #1479 that implements a broader fix for bedrock errors, including error codes mapping to anthropic errors. That handles other streaming errors and propagates exact errors to the user

src/anthropic/lib/bedrock/_stream_decoder.py
```diff
@@ -49,7 +50,16 @@ async def aiter_bytes(self, iterator: AsyncIterator[bytes]) -> AsyncIterator[Ser
...
                 if message:
-                    yield ServerSentEvent(data=message, event="completion")
+                    yield ServerSentEvent(data=message, event=self._get_sse_event_type(message))
+
+    def _get_sse_event_type(self, message: str) -> str:
+        try:
+            data = json.loads(message)
+            if is_dict(data) and data.get("type") == "error":
+                return "error"
+        except (ValueError, KeyError):
```
karpetrosyan, line 60: when it raises `ValueError` or `KeyError`? I think the only error the code above can raise it's `json.decoder.JSONDecodeError` during `json.loads(message)`

PR #1481, opened 2026-05-01
fix: add missing f-string prefix in async_to_httpx_files error message

xodn348:
## What

The `TypeError` raised in `async_to_httpx_files()` was missing the `f` prefix on its format string, causing the error message to literally print `{type(files)}` instead of the actual type. The sync counterpart `to_httpx_files()` already had the correct f-string (line 60).

**Before (line 102):**
```python
raise TypeError("Unexpected file type input {type(files)}, expected mapping or sequence")
# Output: "Unexpected file type input {type(files)}, expected mapping or sequence"
```

**After:**
```python
raise TypeError(f"Unexpected file type input {type(files)}, expected mapping or sequence")
# Output: "Unexpected file type input <class 'int'>, expected mapping or sequence"
```

## Verification

- Build: pass
- Tests: pass (16 passed)
- Lint: pass
- Type check: pass

### DIFF
--- a/src/anthropic/_files.py
+++ b/src/anthropic/_files.py
@@ -99,7 +99,7 @@ async def async_to_httpx_files(files: RequestFiles | None) -> HttpxRequestFiles 
     elif is_sequence_t(files):
         files = [(key, await _async_transform_file(file)) for key, file in files]
     else:
-        raise TypeError("Unexpected file type input {type(files)}, expected mapping or sequence")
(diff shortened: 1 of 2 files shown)


karpetrosyan:
Thanks for the PR! I've fixed it in our codegen, see commit (https://github.com/anthropics/anthropic-sdk-python/commit/06d109aaf36629ec15c8fb076c96aed722933600)

PR #1497, opened 2026-05-05
chore(examples): update shebang from poetry to uv

xodn348:
## What

Updates the shebang line in 4 example files from `#!/usr/bin/env -S poetry run python` to `#!/usr/bin/env -S uv run python`.

## Why

The project migrated from poetry to uv (documented in CONTRIBUTING.md), and newer examples like `agents.py` already use the uv shebang. These 4 older examples were left behind:

- `examples/bedrock.py`
- `examples/text_completions_demo_async.py`
- `examples/text_completions_demo_sync.py`
- `examples/text_completions_streaming.py`

No `poetry.lock` exists in the repo anymore, so running these examples with `./examples/bedrock.py` would fail unless a user happened to have poetry installed globally.

## Verification

- Lint: ruff check passed
- Format: ruff format --check passed (no changes needed)
- Syntax: ast.parse verified all 4 files
- Scope: only shebang lines changed, no other modifications

### DIFF
--- a/examples/bedrock.py
+++ b/examples/bedrock.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S poetry run python
+#!/usr/bin/env -S uv run python
 
 # Note: you must have installed `anthropic` with the `bedrock` extra
 # e.g. `pip install -U anthropic[bedrock]`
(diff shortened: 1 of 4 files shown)


karpetrosyan (approved):
Thanks for the fix!

PR #1507, opened 2026-05-06
Add web search pause_turn resubmission example

### LINKED ISSUE #1339: pause_turn with web_search: docs say re-submit response as-is, but API returns 400 — are we missing something?

mindbomber:
Fixes #1339

## Summary

- Add a focused web-search `pause_turn` example that appends the paused assistant message as a complete assistant turn.
- Preserve returned server-tool content blocks instead of splitting, filtering, or rewriting individual `server_tool_use` blocks.
- Add regression coverage showing `MessageParam` request transformation preserves `ServerToolUseBlock` models when resubmitting assistant content.

## Verification

- `python -m pytest tests/test_transform.py -q -k server_tool_use` -> 2 passed
- `python -m ruff check tests/test_transform.py examples/web_search_pause_turn.py` -> passed
- `python -m py_compile examples/web_search_pause_turn.py` -> passed
- `git diff --check` -> passed
- AANA code-change guardrail -> pass / accept, AIx 1.0, no violations

Note: `python -m pytest tests/test_transform.py -q` also ran, but two pre-existing Windows-only failures appeared in `test_base64_file_input` because the checked-out `tests/sample_file.txt` has CRLF line endings locally, changing the expected base64 value. The new server-tool regression test passed independently.

### DIFF
--- a/examples/web_search_pause_turn.py
+++ b/examples/web_search_pause_turn.py
@@ -1,0 +1,39 @@
+from __future__ import annotations
+
+from anthropic import Anthropic
+from anthropic.types import Message, MessageParam
+
+client = Anthropic()
+
+
+def assistant_turn_from_message(message: Message) -> MessageParam:
+    """Preserve paused server-tool blocks exactly as the SDK returned them."""
+    return {"role": "assistant", "content": message.content}
+
+
+messages: list[MessageParam] = [
+    {"role": "user", "content": "Search for the latest Claude web search documentation."}
+]
+
+while True:
+    message = client.messages.create(
+        model="claude-sonnet-4-5-20250929",
+        max_tokens=1024,
+        messages=messages,
+        tools=[
+            {
+                "name": "web_search",
+                "type": "web_search_20250305",
+            }
+        ],
+    )
+
+    if message.stop_reason != "pause_turn":
+        break
+
+    # Keep the full assistant turn intact. Splitting, filtering, or rewriting
+    # individual server_tool_use blocks can orphan tool state on the next request.
+    messages.append(assistant_turn_from_message(message))
+
+
+print(message.model_dump_json(indent=2))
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@mindbomber thanks. #1339 turned out to be a usage question rather than an SDK gap: the paused assistant turn just has to stay the last message when you resubmit. I answered it there with the loop inline. We keep `examples/` small and the server-tools docs already walk through this flow. I don't think it makes sense to a dedicated example file for something narrow like this.

PR #1508, opened 2026-05-06
Add structured output guidance

### LINKED ISSUE #1034: Anthropic JSON Mode requires client to fabricate a nonsensical tool_use block
According to [the Anthropic API](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/overview#json-mode), structured output is available by submitting a query in "JSON mode":

> You can use tools to get Claude produce JSON output that follows a schema, even if you don’t have any intention of running that output through a tool or function.
> 
> When using tools in this way:
> 
> You usually want to provide a single tool
>
> You should set tool_choice to instruct the model to explicitly use that tool
>
> Remember that the model will pass the input to the tool, so the name of the tool and description should be from the model’s perspective.

That's a little janky, but it's fine. And when I submit a message with the right format, I get back structured output in a tool_use block.

However, I encountered problems while writing some code to engage Anthropic in a chat-style message stream where one of the messages returns structured output. First, when I serialized the structured-output response back into the message stream with some following chat messages, the API complained because I hadn't included the request_id in the tool call for the JSON-output message:

`'error': {'type': 'invalid_request_error', 'message': 'messages.5.content.0.tool_use.id: Field required'}`

Of course, the call ID doesn't really matter because the tool use is not being used to a call an actual function; it's just forcing Claude to pretend that it's calling a function in order to force it to generate structured output. But if I humor Claude and add ['id': (request ID string)] to the serialized request, I get this:

`'error': {'type': 'invalid_request_error', 'message': 'messages.6: `tool_use` ids were found without `tool_result` blocks immediately after: (request ID string). Each `tool_use` block must have a corresponding `tool_result` block in the next message.'}`

So not only do I need to preserve the call ID from the structured output, but I also need to pretend that the schema-definition "tool" is an actual tool and synthesize a fictitious tool_result block with the same call ID, just so that Claude sees that all of its tool calls were fulfilled.

This is nonsense, Anthropic. This janky implementation of JSON mode is causing the API to trip over itself. Other service providers have APIs that support structured output without requiring devs to jump through hoops like this. Please fix.

mindbomber:
Closes #1034.

## Summary

- documents `client.messages.parse(..., output_format=...)` in the README for JSON/schema-constrained responses
- points readers to the existing structured output examples
- adds comments to the sync and streaming examples clarifying that structured output should stay separate from normal tool-use transcripts

The issue describes the friction of preserving a schema-only `tool_use` block and fabricating a matching `tool_result` when continuing a chat. This PR keeps the implementation unchanged and adds a small, reviewable docs/example path for the SDK's structured-output helper instead.

## Verification

- `python -m compileall examples\structured_outputs.py examples\structured_outputs_streaming.py`

## AANA gate

I used AANA as a verifier/correction gate before publishing this PR. The gate was limited to public repository evidence: the issue text, the README change, and the two existing structured-output examples. It does not certify API behavior, production readiness, compliance, or alignment; it only checked that the PR text stays within those evidence limits and avoids broader claims.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,43 @@ message = client.messages.create(
 print(message.content)
 ```
 
+## Structured outputs
+
+For JSON or schema-constrained responses, use `client.messages.parse()` with an `output_format`.
+This keeps structured output separate from normal tool-use transcripts, so you do not need to
+serialize a schema-only `tool_use` block or fabricate a `tool_result` when continuing a chat.
+
+```python
+import pydantic
+from anthropic import Anthropic
+
+
+class Order(pydantic.BaseModel):
+    product_name: str
+    price: float
+    quantity: int
+
+
+client = Anthropic()
+
+parsed_message = client.messages.parse(
+    model="claude-sonnet-4-5",
+    messages=[
+        {
+            "role": "user",
+            "content": "Extract the order from: 2 packs of Green Tea for 5.50 dollars each.",
+        }
+    ],
+    max_tokens=1024,
+    output_format=Order,
+)
+
+print(parsed_message.parsed_output)
+```
+
+See the runnable examples in [`examples/structured_outputs.py`](./examples/structured_outputs.py)
+and [`examples/structured_outputs_streaming.py`](./examples/structured_outputs_streaming.py).
+
 ## Requirements
 
 Python 3.9+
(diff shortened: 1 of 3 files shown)


felixfbecker:
Hi @mindbomber, given the large surface area of our API, we intentionally document all of the SDK features on our docs site that we link to from the README while keeping the README slim. You can find the documentation for `.parse()` here: https://platform.claude.com/docs/en/build-with-claude/structured-outputs#working-with-json-outputs-in-sdks

PR #1513, opened 2026-05-08
perf: reduce root import overhead

### LINKED ISSUE #1211: Slow imports
# timing

Is it really ok that it takes 209ms to import even a single type from this package?

This feels like the _perfect_ task for Claude Code?!

I have no background in the best way to solve this issue, I just know this feels wrong for such a central dependency. And common, let's be honest, this is a glorified HTTP client.

This destroys downstream performance in packages like `langchain_anthropic` which wind up taking 300ms+ for imports alone, just to make a simple http call to generate a completion!

```sh
python3 -X importtime -c "import anthropic"

...
import time:        66 |      69232 |   anthropic.lib.tools
...
import time:       506 |     116232 |   anthropic.types
...
import time:       329 |     209401 | anthropic
```

```sh
python3 -X importtime -c "from anthropic.types.image_block_param import ImageBlockParam"
...
import time:       313 |     208162 |     anthropic
import time:         7 |     208169 |   anthropic.types
import time:        10 |     208178 | anthropic.types.image_block_param
```

I'd hate to think that just because it's python it's ok to ignore optimizations... but, perhaps my expectations are miscalibrated, I'm open to that possibility. Am I crazy?

# Claude seems to agree

I've got Claude working on some ideas for me! I don't have time to do this myself, but, Claude is killing it with perf profiling and I bet will find the spots we can get some big wins.

Gotta tame the explosion of types and importing all of them on every import basically

wilyan09007:
Summary:
Fixes #1211 by reducing `import anthropic` overhead with lazy root exports while preserving the public import surface. The root import now avoids loading the heavy client/type stack until those symbols are used.

Changes:
- [ ] Lazily resolve root client, type, cloud, streaming, credential, and helper exports.
- [ ] Keep type-checker visibility without heavy runtime imports.
- [ ] Avoid importing `anthropic.types` from `_exceptions.py` at runtime.
- [ ] Add import regression tests for heavy modules and lazy exports.

Testing instructions:
- [ ] Run `uv run pytest tests\test_imports.py tests\lib\test_vertex.py tests\lib\tools\test_functions.py -q`.
- [ ] Run `uv run ruff check src\anthropic\__init__.py src\anthropic\_exceptions.py tests\test_imports.py`.
- [ ] Run `uv run pyright src\anthropic\__init__.py src\anthropic\_exceptions.py tests\test_imports.py`.
- [ ] Run `python -X importtime -c "import anthropic"`.

### DIFF
--- a/src/anthropic/__init__.py
+++ b/src/anthropic/__init__.py
@@ -1,51 +1,113 @@
 # File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
 
-import typing as _t
+from __future__ import annotations
 
-from . import types
-from ._types import NOT_GIVEN, Omit, NoneType, NotGiven, Transport, ProxiesTypes, omit, not_given
-from ._utils import file_from_path
-from ._client import (
-    Client,
-    Stream,
-    Timeout,
-    Anthropic,
-    Transport,
-    AsyncClient,
-    AsyncStream,
-    AsyncAnthropic,
-    RequestOptions,
-)
-from ._models import BaseModel
+import os as _os
+import typing as _t
+import logging as _logging
+
 from ._version import __title__, __version__
-from ._response import APIResponse as APIResponse, AsyncAPIResponse as AsyncAPIResponse
-from ._constants import (
-    AI_PROMPT as AI_PROMPT,
-    HUMAN_PROMPT as HUMAN_PROMPT,
-    DEFAULT_TIMEOUT,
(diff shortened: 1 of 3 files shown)


dtmeadows-ant:
@wilyan09007 thanks. Import cost is a fair complaint and #1211 stays open for it. A hand-maintained lazy-export table of this size duplicates the export list and will drift from it as resources are added.  Closing the PR on that basis.

PR #1514, opened 2026-05-08
feat(bedrock): support performance latency config

### LINKED ISSUE #971: Bedrock: support low-latency configuration
Currently, `[Async]AnthropicBedrock` doesn't provide a way to enable low-latency mode.
It also appears that this configuration isn't even accessible via the Bedrock Anthropic HTTP API.

So the only option for the Antropic clients to use this low-latency model is to use the Converse API provided by AWS Becrock: https://docs.aws.amazon.com/bedrock/latest/userguide/latency-optimized-inference.html

This is very inconvenient, since the Converse API unifies APIs for multiple models and, therefore, it lacks support for exclusive features of particular models _(e.g. `tool_choice=none` is missing from the Converse API)_.

Exposing low-latency configuration in `[Async]AnthropicBedrock` would be great.

wilyan09007:
Summary:
Fixes #971 by adding Bedrock low-latency configuration support to the sync and async Bedrock clients. The option maps to the Bedrock performance latency request header for standard and streaming invocations.

Changes:
- [ ] Add `performance_config_latency` to sync and async Bedrock clients.
- [ ] Apply the Bedrock performance latency header during request preparation.
- [ ] Preserve, override, or clear the latency option through `copy()`.
- [ ] Add focused tests for default, sync, async, streaming, and copy behavior.

Testing instructions:
- [ ] Run `uv run ruff check src\anthropic\lib\bedrock\_client.py tests\lib\test_bedrock.py`.
- [ ] Run `uv run pyright src\anthropic\lib\bedrock\_client.py`.
- [ ] Run `uv run --extra bedrock pytest tests\lib\test_bedrock.py -k "performance_config_latency or messages_retries or application_inference_profile or api_key_auth"`.
- [ ] Confirm optimized clients send `X-Amzn-Bedrock-PerformanceConfig-Latency`.

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -4,7 +4,7 @@ import os
 import logging
 import urllib.parse
 from typing import Any, Union, Mapping, TypeVar
-from typing_extensions import Self, override
+from typing_extensions import Self, Literal, TypeAlias, override
 
 import httpx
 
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@wilyan09007 thanks. This is already reachable without a new client option: `AnthropicBedrock(default_headers={"X-Amzn-Bedrock-PerformanceConfig-Latency": "optimized"})`, or `extra_headers=` on a single call. We'd rather keep Bedrock-specific features on that generic header path than add a constructor argument for each one. I'm going to close this; the header approach answers #971 too.

PR #1516, opened 2026-05-08
feat(client): support callable auth tokens

### LINKED ISSUE #1496: Accept a callable for bearer-token auth (`auth_token: str | Callable[[], str]`)
Most enterprise customers are required by policy to use short-lived, broker-issued bearer tokens (Entra ID, Okta, Ping, Vault) instead of long-lived API keys. Those tokens expire (Entra is ~1h), so any long-running client needs the credential resolved per request, not captured once at construction.

**State of the SDK (v0.109.x).** Two auth inputs exist and neither covers the simple callable case:

- `api_key` and `auth_token` are `str` only, fixed at construction.
- `credentials=` accepts an `AccessTokenProvider` with caching, refresh, and refresh-on-401. This is the right tool for expiry-aware providers, but it requires the full protocol, `__call__(self, *, force_refresh: bool = False) -> AccessToken`. A plain `Callable[[], str]`, which is what most identity stacks already hand you, can't be passed without a wrapper:

```python
from anthropic import AccessToken

def bearer_provider(get_token):
    def _provider(*, force_refresh: bool = False) -> AccessToken:
        return AccessToken(token=get_token(), expires_at=None)
    return _provider
```

Every caller writes the same adapter, and `force_refresh` typically can't be honored (common token getters have no force-refresh hook), so a 401 retry can re-serve a stale token.

**Ask.** Make `auth_token` accept `str | Callable[[], str]`. When callable, invoke it per request to build `Authorization: Bearer <token>`. No `AccessToken` and no `force_refresh` for the simple case; static behavior unchanged; `credentials=` stays the advanced path. This is the same callable-credential pattern the peer Stainless-generated SDKs already expose.

```python
def get_token() -> str:
    return my_auth.get_access_token()

client = Anthropic(base_url=..., auth_token=get_token)
```

**Async parity.** `AsyncAnthropic` should accept `auth_token: str | Callable[[], Awaitable[str]]`, awaited per request. The credential subsystem is sync-only today (even `credentials=` is run via a thread-pool `asyncify` bridge), so an async-native credential can't be used without pulling in its sync counterpart. This can land as a follow-up increment so it doesn't block the sync callable, but it's the same feature and the peer Stainless-generated SDKs already type their async callable as returning an awaitable.

**Status:** implemented in #1516 (currently needs a rebase and review).

wilyan09007:
Summary:
Fixes #1496 by allowing `auth_token` to be a string or a callable returning the current bearer token. Callable tokens are resolved when auth headers are built, so long-running clients can use refreshed tokens.

Changes:
- [ ] Add an `AuthToken` alias for static and callable bearer tokens.
- [ ] Resolve callable auth tokens per request for sync and async clients.
- [ ] Preserve callable token behavior through copied clients.
- [ ] Document short-lived bearer token usage in the README.
- [ ] Add tests for callable tokens and credential precedence.

Testing instructions:
- [ ] Run `uv run python -m pytest tests\test_client.py -k "auth_token_callable or validate_headers" -n0`.
- [ ] Run `uv run python -m pytest tests\lib\test_credentials.py -k "auth_token_callable or explicit_auth_token_shadows or copy_with_explicit_api_key_shadows" -n0`.
- [ ] Run `uv run ruff check src\anthropic\_client.py src\anthropic\__init__.py tests\test_client.py tests\lib\test_credentials.py README.md`.
- [ ] Run `uv run pyright src\anthropic\_client.py src\anthropic\__init__.py`.


achandmsft:
This is the right fix and it still fills a real gap after the conflicting PR that included `credentials=` shipped in v0.109. `credentials=` covers the advanced case but requires the full `AccessTokenProvider` protocol (`__call__(self, *, force_refresh: bool = False) -> AccessToken`), so anyone who just has a `Callable[[], str]` that returns a bearer token has to hand-roll an adapter and a no-op `force_refresh`. Making `auth_token` accept `str | Callable[[], str]`, the way this PR does, is the missing simple path and is symmetric with how `api_key` should behave.

The branch is currently CONFLICTING, almost certainly because `credentials=` touched `_client.py` after this was opened. @wilyan09007 are you able to rebase onto main? Happy to help reconcile the auth-header resolution with the new credential precedence so the two paths coexist cleanly.

Flagging for maintainer triage as well, this has been open since early May with no review and it closes out #1496. (cc @RobertCraigie / @bradabrams

achandmsft:
Also worth considering, either in scope or as an immediate follow-up: the async client accepting `Callable[[], Awaitable[str]]`, awaited in `async_auth_flow` rather than bridged through `asyncify`, so async-native credential sources don't have to fall back to their sync variant. The sync callable in this PR is unaffected; this is an additive widen on the `AsyncAnthropic` path.

dtmeadows-ant:
@wilyan09007 thanks. Dynamic bearer tokens are what `credentials=` (an `AccessTokenProvider`, since 0.109) is for. It resolves per request and handles refresh-on-401, and it's the supported path today even if a plain callable needs a small adapter. Whether `auth_token` itself should accept a callable is tracked in #1496.

PR #1473, opened 2026-04-30
chore(examples): bump tools_runner.py to claude-sonnet-4-5-20250929

GopalGB:
## What

Bump the model in `examples/tools_runner.py` from `claude-3-5-sonnet-latest` to `claude-sonnet-4-5-20250929`.

## Why

`examples/tools_runner.py` is the only example file still pointing at `claude-3-5-sonnet-latest`. Every sibling example uses `claude-sonnet-4-5-20250929`:

```
$ grep -h 'model=' examples/*.py | sort -u
        model="anthropic.claude-sonnet-4-5-20250929-v1:0",
        model="claude-haiku-4-5",
        model="claude-sonnet-4-5",
        model="claude-sonnet-4-5-20250929",
        model="claude-sonnet-4-6",
        model="claude-3-5-sonnet-latest",   <- the outlier
```

Aligning this example so copy-paste users land on a current model.

## Notes

- Per CONTRIBUTING.md, `examples/` is not modified by the generator, so this edit is safe to land manually.
- The example logic is unchanged — only the model identifier moves to the current default seen elsewhere in `examples/`.

## How to test

```bash
$ ANTHROPIC_API_KEY=... uv run python examples/tools_runner.py
```

Output should match the previous behavior — the runner queries the weather tool and prints messages.

PR #1520, opened 2026-05-08
chore(examples): bump workload_identity.py to claude-opus-5

lucky-verma:
`examples/workload_identity.py` still pins `claude-opus-4-5`.

rebuilt the branch — it was stacked on #1497 and #1473, which have since landed upstream in a different form, hence the conflict. it's now one commit against current `next`, one file.

`claude-opus-5` keeps the example on the tier it already used. most of `examples/` uses `claude-sonnet-5` though, so happy to switch if you'd rather they converge.

federated identity flow untouched. ruff check + format clean.


dtmeadows-ant:
Hey there, sorry for the delay on this. If you can rebase this onto the latest main, I'm happy to approve.

PR #1535, opened 2026-05-13
fix(tools): raise clear error for duplicate tool names in tool_runner

Zawwarsami16:
I was poking around `tool_runner` and noticed that the runner accepts duplicate tool names without complaining. If you pass two tools that happen to share a name (say, two different `get_weather` impls imported from different modules), the second silently overwrites the first in `_tools_by_name = {tool.name: tool for tool in tools}`, and your first tool is gone before the request is even built. The API would reject the duplicate, but only with a generic 400 after the round trip, and by then you have no idea which side of your code was wrong.

This patch adds an upfront check in `BaseToolRunner.__init__` that walks the tools, collects any names that show up more than once, and raises a `ValueError` listing them. It runs before any HTTP work, so you get the error at the call site of `tool_runner(...)` instead of inside a request handler. Behaviour for unique names is unchanged. I left the `_tools_by_name` dict comprehension in place and just added the validation pass above it.

Three new tests in `tests/lib/tools/test_runners.py` cover the sync runner, the async runner, and the case where multiple distinct names each have duplicates (the error should list all of them). `uv run pytest tests/lib/ -n auto` reports 496 passed, 1 skipped, 1 xfailed, and `uv run ruff check . && uv run pyright && uv run mypy .` are clean.

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -72,7 +72,18 @@ class BaseToolRunner(Generic[AnyFunctionToolT, ResponseFormatT]):
         max_iterations: int | None = None,
         compaction_control: CompactionControl | None = None,
     ) -> None:
-        self._tools_by_name = {tool.name: tool for tool in tools}
+        tools_list = list(tools)
+        seen: dict[str, int] = {}
+        for tool in tools_list:
+            seen[tool.name] = seen.get(tool.name, 0) + 1
+        duplicates = sorted(name for name, count in seen.items() if count > 1)
+        if duplicates:
+            raise ValueError(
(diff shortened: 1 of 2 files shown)


Zawwarsami16:
coming back to this after a few weeks. looks like tool registration got pulled out into a shared `_tool_dispatch.tool_registry()` helper since I opened this, and both the messages and sessions runners go through it now. the docstring there says on a duplicate name the later tool wins on purpose, so a hard raise like this one cuts against the behavior you've settled on, and the patch conflicts on top of that anyway. closing it rather than letting it sit. if a stricter opt-in ever sounds useful (a flag that errors instead of last-wins) I'm happy to redo it against the new helper, just say the word.

PR #1541, opened 2026-05-14
fix: only suppress auto-append when caller explicitly provides a tool result

### LINKED ISSUE #1536: [Bug] append_messages() inside tool runner loop causes infinite loop (advanced usage docs example)
[Link to Advanced Usage Docs Example
](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-runner#:~:text=(block.text)-,Advanced%20usage,-Within%20the%20loop)
# Root cause
  
Calling `runner.append_messages(user_message)` inside the loop sets
`_messages_modified = True`, which causes `__run__` to skip the auto-append
of the assistant message + tool result (line 281 of `_beta_runner.py`).
The next iteration sends a request with no tool result in history, so Claude
sees the original question unanswered and makes the tool call again.

## Minimal repro

Compacted and consolidated from the website

```
from anthropic import beta_tool, Anthropic

client = Anthropic()


@beta_tool
def get_weather(location: str, unit: str = 'fahrenheit') -> str:
    return json.dumps({
        'temperature': '20°c', 'condition': 'sunny'
    })


    model="claude-haiku-4-5",
    max_tokens=1024,
    tools=[get_weather],
    messages=[{
        "role": "user",
        "content": "what's the weather in San Francisco?"
    }]
)


for message in runner:
    tool_response = runner.generate_tool_call_response()
    if tool_response:
        print(f"Tool result: {tool_response}")

    # increase tokens for next request
    runner.set_messages_params(
        lambda params: {
            **params,
            "max_tokens": 2048,
        }
    )

    # add additional messages
    runner.append_messages(
        {"role": "user", "content": "be concise in your response"}
    )
```

## Expected behavior
The "be concise" user message is appended, tool results are still
auto-appended, loop terminates.

## Actual behavior
Infinite loop — tool result never reaches history.

## Note
The "Modifying tool results" example on the same page correctly calls
`runner.append_messages(message, tool_response)` (both the assistant
message and the tool result), which is the pattern that works correctly.

The "Advanced usage" example is inconsistent with this.

Echolonius:
## Summary

Fixes #1536.

Calling `runner.append_messages()` with a plain user message inside the tool runner loop caused an infinite loop. Any call to `append_messages()` unconditionally set `_messages_modified = True`, which the runner interpreted as the caller manually handling the tool result. It would skip its own auto-append of the assistant message and tool result, so the next iteration sent a request with no tool result in history — Claude saw the original question unanswered, re-issued the same tool call, and the loop never terminated.

## Root cause

`_messages_modified` was serving two distinct purposes with no way to tell them apart:

1. **Caller is adding extra context** (e.g. `append_messages({"role": "user", "content": "be concise"})`) — runner should still auto-append the tool result
2. **Caller is manually providing the tool result** (e.g. `append_messages(assistant_msg, tool_result_msg)`) — runner should skip auto-append

Both paths set the flag, so the runner always skipped auto-append regardless of intent.

## Fix

`_messages_modified` is now only set when at least one of the appended messages contains a `tool_result` content block — the unambiguous signal that the caller is handling the tool result themselves. Appending messages without tool results leaves the flag untouched and the loop continues correctly.

## What this affects

- Sync runner (`BetaToolRunner`) ✅
- Async runner (`BetaAsyncToolRunner`) ✅ — both share the same `append_messages()` on the base class
- The working "Modifying tool results" pattern (passing both assistant message + tool result) continues to work unchanged
- `set_messages_params()` is unaffected — it never touches the flag

## Tests

Added `test_append_messages_non_tool_result_does_not_suppress_auto_append` — a pure unit test (no API calls required) that directly asserts the flag behaviour for both the plain-message case and the tool-result case.

The existing `test_custom_message_handling` (marked `xfail`) is a related but separate issue and remains untouched.

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -111,12 +111,32 @@ class BaseToolRunner(Generic[AnyFunctionToolT, ResponseFormatT]):
 
         This invalidates the cached tool response, i.e. if tools were already called, then they will
         be called again on the next loop iteration.
+
+        If any of the appended messages contain a ``tool_result`` content block, the runner treats
+        this as the caller manually handling the tool response and will skip its own auto-append of
+        the assistant message and tool result for that iteration.  Appending messages that do *not*
+        contain tool results (e.g. an extra user instruction) leaves the auto-append behaviour
+        unchanged so the loop continues correctly.
         """
         message_params: List[BetaMessageParam] = [
             {"role": message.role, "content": message.content} if isinstance(message, BetaMessage) else message
             for message in messages
         ]
-        self._messages_modified = True
+        # Only suppress the runner's auto-append when the caller is explicitly providing a tool
+        # result.  Previously this flag was always set, which caused an infinite loop when callers
+        # appended non-tool-result messages (e.g. extra instructions) inside the loop because the
+        # tool result was never added to the conversation history.
+        has_tool_result = any(
+            isinstance(msg, dict)
+            and isinstance(msg.get("content"), list)
+            and any(
+                isinstance(block, dict) and block.get("type") == "tool_result"
+                for block in msg.get("content", [])
+            )
+            for msg in message_params
+        )
+        if has_tool_result:
+            self._messages_modified = True
         self.set_messages_params(lambda params: {**params, "messages": [*params["messages"], *message_params]})
         self._cached_tool_call_response = None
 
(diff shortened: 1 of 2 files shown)


Echolonius:
Closing in favour of #1538 by @Zelys-DFKH, which was submitted earlier and addresses both failure modes more completely — including the case where a manually-provided `tool_result` block is sent without the preceding assistant turn, and the removal of the existing `xfail` test with a proper HTTP fixture update.

I should have checked for open PRs against this issue before opening a duplicate. Apologies for the noise.

PR #1542, opened 2026-05-14
fix(streaming): use parse_obj fallback to preserve nested event types (fixes #941)

### LINKED ISSUE #941: content_block_delta event not deserialized correctly during streaming

Echolonius:
## Problem

When Pydantic's union discriminator silently returns the raw dict instead of a typed model, `accumulate_event` raises `TypeError` at the second `isinstance` check:

```
Unexpected event runtime type, after deserialising twice - {'type': 'content_block_delta', 'index': 0, 'delta': {'type': 'text_delta', 'text': '...\}} - <class 'dict'>
```

## Why existing PRs (#1167, #1287, #1448, #1467) do not fully fix it

PRs #1287, #1448, and #1467 all add a type-map fallback, but use `model_construct(**raw)`. That method skips Pydantic validation and leaves **nested fields as raw dicts** — so `event.delta` inside a `content_block_delta` event remains `{"type": "text_delta", "text": "..."}`. The moment `accumulate_event` accesses `event.delta.type`, a secondary `AttributeError` is raised and the text is never accumulated.

PR #1167 fixes the root cause in `_models.py` but changes core model-construction logic in a way that is hard to verify will not have side effects elsewhere.

## This fix

Replace `model_construct` with `parse_obj()` — the SDK's existing Pydantic v1/v2 compatibility helper (in `_compat.py`). `parse_obj` runs full validation, so nested union fields like `delta` are properly resolved to their typed objects (`TextDelta`, `InputJSONDelta`, etc.), not left as raw dicts. If `parse_obj` itself raises, the original `TypeError` is preserved.

The same fallback is applied symmetrically to `_beta_messages.py`.

## Changes

- `src/anthropic/lib/streaming/_messages.py` — add `_RAW_EVENT_TYPE_MAP`, import concrete event classes and `parse_obj`, apply fallback in `accumulate_event`
- `src/anthropic/lib/streaming/_beta_messages.py` — same for the beta path
- `tests/lib/streaming/test_messages.py` — three regression tests for the non-beta path (single delta accumulates, multiple deltas accumulate, unknown type still errors)
- `tests/lib/streaming/test_partial_json.py` — three matching regression tests for the beta path (`TestBetaRawDictFallback`)

## Test plan

- `test_accumulate_event_raw_dict_text_delta_is_fully_typed` — passes a raw dict `content_block_delta` through `accumulate_event` and asserts `snapshot.content[0].text == "hello"`, which only passes if `event.delta.type` was accessible (i.e. delta is a typed `TextDelta`, not a dict)
- `test_accumulate_event_raw_dict_multiple_deltas_accumulate` — three raw dict deltas build up `"Hello there!"`
- `test_accumulate_event_raw_dict_unknown_type_still_raises` — unknown type string still raises `(TypeError, RuntimeError)`
- Beta variants of all three in `TestBetaRawDictFallback` (in `test_partial_json.py`)

---

*Written with [Claude Code](https://claude.ai/code)*

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -14,6 +14,7 @@ from anthropic.types.beta.beta_server_tool_use_block import BetaServerToolUseBlo
 
 from ..._types import NOT_GIVEN, NotGiven
 from ..._utils import consume_sync_iterator, consume_async_iterator
+from ..._compat import parse_obj
 from ..._models import build, construct_type, construct_type_unchecked
 from ._beta_types import (
     BetaCitationEvent,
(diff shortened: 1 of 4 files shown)


Echolonius:
## Comparison with #1467

#1467 uses `construct_type_unchecked(type_=RawContentBlockDeltaEvent, value=event)` as the fallback. That converts the outer event into a `BaseModel`, so the first `isinstance` check passes — but `construct_type_unchecked` calls `model_construct` internally, which **skips Pydantic validation and leaves nested fields as raw dicts**.

Concretely, after #1467's fix, for a raw-dict `content_block_delta` event:

```python
event.type        # "content_block_delta" ✓
event.delta       # {"type": "text_delta", "text": "hello"}  ← still a dict
event.delta.type  # AttributeError: 'dict' object has no attribute 'type'
```

`accumulate_event` then hits this line:

```python
if event.delta.type == "text_delta":   # AttributeError here
    content.text += event.delta.text
```

So the text is never accumulated — the crash just moves from `TypeError` to `AttributeError`.

This PR uses `parse_obj(target_cls, raw)` instead. `parse_obj` runs full Pydantic validation, so nested union fields are resolved to their typed objects:

```python
event.type        # "content_block_delta" ✓
event.delta       # TextDelta(type="text_delta", text="hello") ✓
event.delta.type  # "text_delta" ✓  — no AttributeError
content.text += event.delta.text  # works correctly
```

The regression test `test_accumulate_event_raw_dict_text_delta_is_fully_typed` specifically catches this: it asserts `snapshot.content[0].text == "hello"`, which only passes if `event.delta` was a typed object (not a dict) when `accumulate_event` ran.

dtmeadows-ant:
@Echolonius thanks for this. The underlying issue (#941) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1545, opened 2026-05-14
fix(tools): preserve tool flow when caller appends unrelated messages (closes #1536)

### LINKED ISSUE #1536: [Bug] append_messages() inside tool runner loop causes infinite loop (advanced usage docs example)

0xdeadd:
Closes #1536.

## The bug

The reporter found that calling `append_messages` inside the `tool_runner` loop body causes an infinite tool-call loop. Their root-cause analysis is correct: any `append_messages` call sets `_messages_modified = True`, which causes `__run__` to skip the auto-append of the assistant message + tool result. The next iteration sends a request with no tool result, the model re-asks for the same tool call, repeat forever.

## Why the flag is wrong

`_messages_modified` was conflating two distinct use cases:

1. **Modifying tool results** — caller appends a custom `tool_result` to override the SDK's auto-generated one. Auto-append should skip.
2. **Append an unrelated user message** (#1536) — caller appends, say, `"be concise in your response"` between iterations. The tool result still needs to be threaded back in, otherwise the loop breaks.

A single boolean can't distinguish these — one means "I handled the tool flow," the other means "I added context, please continue normally."

## The fix

Replace the flag-based gate with a **content-based check**: auto-append is skipped only when the last user message already contains a `tool_result` block whose `tool_use_id` matches the one we were about to write. Unrelated messages don't match, so the tool flow stays correctly threaded.

Applied symmetrically to sync and async `__run__`.

## Tests

- New unit test `test_tool_response_already_appended_detects_matching_tool_result` covers:
  - Empty history → no match
  - History with an unrelated user message (the #1536 case) → no match → auto-append still runs
  - History with a matching `tool_result` (modifying-tool-results case) → match → auto-append skipped
  - Mismatched `tool_use_id` → no match
- Full `tests/lib/tools/test_runners.py` passes: **13 passed, 1 xfailed**.
- Existing `test_custom_message_handling` xfail is preserved — that pattern depends on additional re-threading logic (inserting the assistant message before the caller's tool result) and snapshot updates that are out of scope for this fix.

## Repro from the issue

The reporter's repro (the "be concise" message inside the loop body) now terminates correctly: the tool result is auto-appended, the model sees its own tool answer, and produces the final text on the next iteration. Verified locally.

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -125,6 +125,44 @@ class BaseToolRunner(Generic[AnyFunctionToolT, ResponseFormatT]):
             return True
         return False
 
+    def _tool_response_already_appended(self, response: BetaMessageParam) -> bool:
+        """Return True if the tool-result message we would auto-append has already
+        been added to the conversation history (e.g. the caller appended a custom
+        tool result themselves via ``append_messages``).
+
+        The check matches by ``tool_use_id`` on the tool_result blocks in the last
+        user message. This lets callers append unrelated messages (e.g. an extra
+        user instruction) inside the loop body without breaking the
+        tool-call/tool-result threading: no tool_use_ids match, so the assistant
+        + tool result still get auto-appended on the next iteration.
+        """
+        response_content = response.get("content")
+        if not isinstance(response_content, list):
+            return False
+        response_ids = {
+            block.get("tool_use_id")
+            for block in response_content
+            if isinstance(block, dict) and block.get("type") == "tool_result" and block.get("tool_use_id")
+        }
+        if not response_ids:
+            return False
+
+        msgs = self._params.get("messages", [])
+        if not msgs:
+            return False
+        last = msgs[-1]
+        if not isinstance(last, dict) or last.get("role") != "user":
+            return False
+        last_content = last.get("content")
+        if not isinstance(last_content, list):
+            return False
+        last_ids = {
+            block.get("tool_use_id")
+            for block in last_content
+            if isinstance(block, dict) and block.get("type") == "tool_result" and block.get("tool_use_id")
+        }
+        return response_ids.issubset(last_ids)
+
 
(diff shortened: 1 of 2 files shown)


0xdeadd:
Friendly ping on this one. It's a small, self-contained fix for #1536 (tool flow breaks when a caller appends unrelated messages between tool calls). Branch is mergeable and CI is clean. Happy to rebase or adjust if a maintainer can take a look. Thanks!

PR #1550, opened 2026-05-15
Wrap mid-stream httpx.TransportError as APIConnectionError

mengtingli-ant:
## What

Wrap mid-stream `httpx.TransportError` (during SSE body iteration) as `anthropic.APIConnectionError`, in both `Stream._iter_events` and `AsyncStream._iter_events`.

## Why

Mid-stream transport drops (`RemoteProtocolError`, `ReadError`, `ConnectError`) currently leak through as bare `httpx` exceptions because the SDK's wrapping in `_base_client._request` only covers the *pre-body* request. Once the SSE 200 is sent and body iteration starts, there's no try/except in `_iter_events`.

This means customers' standard retry ladders:
```python
except anthropic.APIConnectionError:
    retry()
```
…miss mid-stream drops. They have to know to also catch `httpx.TransportError`, which nobody discovers without debugging.

**Found while reproducing DeepSearchQA via the public API**: 35/45 terminal question failures were `RemoteProtocolError("peer closed connection without sending complete message body")` that the standard retry ladder missed.

## Changes

- `Stream._iter_events` / `AsyncStream._iter_events`: wrap `httpx.TransportError` as `APIConnectionError` (same pattern as `_base_client.py:1104`); let `TimeoutException` pass through unchanged so it doesn't get double-wrapped (`APITimeoutError` is already an `APIConnectionError` subclass).
- 2 tests: mid-stream `RemoteProtocolError` is wrapped (with `__cause__` preserved); mid-stream `ReadTimeout` passes through.

## Not in this PR

The follow-up — auto-retry the full request inside `MessageStream.get_final_message()` on mid-stream `APIConnectionError` — is a larger behavior change. This PR just makes the exception catchable with the right type.

### DIFF
--- a/src/anthropic/_streaming.py
+++ b/src/anthropic/_streaming.py
@@ -11,6 +11,7 @@ from typing_extensions import Self, Protocol, TypeGuard, override, get_origin, r
 
 import httpx
 
+from ._exceptions import APIConnectionError
 from ._utils import is_dict, extract_type_var_from_base
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


src/anthropic/_streaming.py
```diff
@@ -72,7 +73,23 @@ def __iter__(self) -> Iterator[_T]:
...
+            # the latter will still see it; this clause only exists so the next clause
+            # doesn't double-wrap it (TimeoutException is also a TransportError).
+            raise
+        except httpx.TransportError as exc:
+            # Mid-stream transport drops (RemoteProtocolError, ReadError, ConnectError, …)
+            # leak through as bare httpx exceptions because the SDK's wrapping in
+            # `_base_client._request` only covers the pre-body request. Re-wrap them so
+            # `except anthropic.APIConnectionError:` catches mid-stream drops the same way
+            # it catches connection failures, and the original is preserved as `__cause__`.
+            raise APIConnectionError(message=f"Stream interrupted: {exc}", request=self.response.request) from exc
```
craigie-ant, line 92: Unfortunately we cannot ship this right now as it is a breaking change because any previously written code that catches `httpx.TransportError` will now be silently skipped.

I've added this change to our list of changes to consider for our v1 release.

PR #1553, opened 2026-05-16
fix(bedrock): detect region from AWS_DEFAULT_REGION and aws_profile arg

JiwaniZakir:
The `_infer_region` function in `src/anthropic/lib/bedrock/_client.py` only checked `AWS_REGION` and ignored the standard `AWS_DEFAULT_REGION` environment variable, and did not pass the `aws_profile` argument to the boto3 session, so a non-default profile's region was never picked up. This PR adds `AWS_DEFAULT_REGION` as a fallback after `AWS_REGION`, updates `_infer_region` to accept and forward `aws_profile` to `boto3.Session`, and threads that argument through both `AnthropicBedrock` and `AsyncAnthropicBedrock`. Run `pytest tests/lib/test_bedrock.py` to verify the new `test_region_infer_from_aws_default_region_env` and `test_region_infer_from_aws_profile_arg` tests pass alongside the existing suite.

Fixes #892

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -67,17 +67,17 @@ def _prepare_options(input_options: FinalRequestOptions) -> FinalRequestOptions:
     return options
 
 
-def _infer_region() -> str:
+def _infer_region(aws_profile: str | None = None) -> str:
     """
     Infer the AWS region from the environment variables or
     from the boto3 session if available.
     """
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@JiwaniZakir thanks for this. The underlying issue (#892) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1555, opened 2026-05-16
fix(bedrock): convert non-200 SSE error events to APIStatusError

### LINKED ISSUE #1477: Bedrock streaming SSE events errors are not handled gracefully
In some cases bedrock SSE stream produce error SSE events like following:
```
{'status_code': 400, 'headers': {':exception-type': 'internalServerException', ':content-type': 'application/json', ':message-type': 'exception'}, 'body': b'{"message":"The system encountered an unexpected error during processing. Try your request again."}
```

**Current behaviour**
Anthropic SDK raises `ValueError`
```python
  File "/app/.venv/lib/python3.12/site-packages/anthropic/_streaming.py", line 211, in __aiter__
    async for item in self._iterator:
  File "/app/.venv/lib/python3.12/site-packages/anthropic/_streaming.py", line 225, in __stream__
    async for sse in iterator:
  File "/app/.venv/lib/python3.12/site-packages/anthropic/_streaming.py", line 215, in _iter_events
    async for sse in self._decoder.aiter_bytes(self.response.aiter_bytes()):
  File "/app/.venv/lib/python3.12/site-packages/anthropic/lib/bedrock/_stream_decoder.py", line 50, in aiter_bytes
    message = self._parse_message_from_event(event)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/anthropic/lib/bedrock/_stream_decoder.py", line 58, in _parse_message_from_event
    raise ValueError(f"Bad response code, expected 200: {response_dict}")
ValueError: Bad response code, expected 200: {'status_code': 400, 'headers': {':exception-type': 'internalServerException', ':content-type': 'application/json', ':message-type': 'exception'}, 'body': b'{"message":"The system encountered an unexpected error during processing. Try your request again."}'}
```

**Expected behaviour**
SDK should either retry this internally or propagate to user-facing `APIStatusError`

**Environment**
anthropic SDK v0.96.0
python 3.12

nileshpatil6:
## Problem

When Bedrock returns a non-200 status code inside an SSE event frame (for example `internalServerException` with `status_code: 400` or `internalServerException` with `status_code: 500`), `AWSEventStreamDecoder._parse_message_from_event` raises a raw `ValueError`:

```
ValueError: Bad response code, expected 200: {'status_code': 400, 'headers': {':exception-type': 'internalServerException', ...}, 'body': b'{"message":"The system encountered an unexpected error..."}'}
```

This `ValueError` is not part of the SDK error hierarchy, so user code that catches `anthropic.APIError` / `anthropic.InternalServerError` / etc. silently misses the error and has to special-case a raw Python built-in exception type.

Fixes #1477

## Root cause

`_parse_message_from_event` used `raise ValueError(...)` directly instead of routing the error through the existing SSE error path.

## Fix

Instead of raising `ValueError`, the method now returns a `ServerSentEvent(event="error", data=<json>)`. The existing `Stream.__stream__` and `AsyncStream.__stream__` already handle `event == "error"` by calling `self._client._make_status_error(...)`, which maps the exception to the correct `APIStatusError` subclass (`InternalServerError`, `BadRequestError`, etc.).

The error body is constructed from the Bedrock event headers (`:exception-type`) and the JSON body (`message` field), giving callers the full context they need.

## What changed

- `src/anthropic/lib/bedrock/_stream_decoder.py`: replace `raise ValueError(...)` with a yielded `ServerSentEvent(event="error", ...)` containing the Bedrock exception type and message. Also changed the return type of `_parse_message_from_event` from `str | None` to `ServerSentEvent | None` (the callers already just checked truthiness and yielded the result), and updated `iter_bytes` / `aiter_bytes` to yield the returned `ServerSentEvent` directly rather than constructing a new one.

## Before / after

Before:
```python
except ValueError:
    # uncatchable via anthropic.APIError
```

After:
```python
except anthropic.InternalServerError:
    # properly caught, e.message contains Bedrock's error detail
```

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -99,31 +99,38 @@ class BaseBedrockClient(BaseClient[_HttpxClientT, _DefaultStreamT]):
         body: object,
         response: httpx.Response,
     ) -> APIStatusError:
-        if response.status_code == 400:
+        # Bedrock streaming errors arrive inside 200 OK event frames.
+        # The decoder embeds the real status code in the body so we map
+        # to the correct exception subclass here.
+        status_code = response.status_code
+        if isinstance(body, dict) and body.get("_bedrock_status") is not None:
+            status_code = int(body["_bedrock_status"])
+
+        if status_code == 400:
             return _exceptions.BadRequestError(err_msg, response=response, body=body)
 
-        if response.status_code == 401:
+        if status_code == 401:
             return _exceptions.AuthenticationError(err_msg, response=response, body=body)
 
-        if response.status_code == 403:
+        if status_code == 403:
             return _exceptions.PermissionDeniedError(err_msg, response=response, body=body)
 
-        if response.status_code == 404:
+        if status_code == 404:
             return _exceptions.NotFoundError(err_msg, response=response, body=body)
 
-        if response.status_code == 409:
+        if status_code == 409:
             return _exceptions.ConflictError(err_msg, response=response, body=body)
 
-        if response.status_code == 422:
+        if status_code == 422:
             return _exceptions.UnprocessableEntityError(err_msg, response=response, body=body)
 
-        if response.status_code == 429:
+        if status_code == 429:
             return _exceptions.RateLimitError(err_msg, response=response, body=body)
 
-        if response.status_code == 503:
+        if status_code == 503:
             return _exceptions.ServiceUnavailableError(err_msg, response=response, body=body)
 
-        if response.status_code >= 500:
+        if status_code >= 500:
             return _exceptions.InternalServerError(err_msg, response=response, body=body)
         return APIStatusError(err_msg, response=response, body=body)
 
--- a/src/anthropic/lib/bedrock/_mantle.py
+++ b/src/anthropic/lib/bedrock/_mantle.py
@@ -66,34 +66,41 @@ class BaseMantleClient(BaseClient[_HttpxClientT, _DefaultStreamT]):
         body: object,
         response: httpx.Response,
     ) -> APIStatusError:
-        if response.status_code == 400:
+        # Bedrock streaming errors arrive inside 200 OK event frames.
+        # The decoder embeds the real status code in the body so we map
+        # to the correct exception subclass here.
+        status_code = response.status_code
+        if isinstance(body, dict) and body.get("_bedrock_status") is not None:
+            status_code = int(body["_bedrock_status"])
+
+        if status_code == 400:
             return _exceptions.BadRequestError(err_msg, response=response, body=body)
 
-        if response.status_code == 401:
+        if status_code == 401:
             return _exceptions.AuthenticationError(err_msg, response=response, body=body)
 
-        if response.status_code == 403:
+        if status_code == 403:
             return _exceptions.PermissionDeniedError(err_msg, response=response, body=body)
 
-        if response.status_code == 404:
+        if status_code == 404:
             return _exceptions.NotFoundError(err_msg, response=response, body=body)
 
-        if response.status_code == 409:
+        if status_code == 409:
             return _exceptions.ConflictError(err_msg, response=response, body=body)
 
-        if response.status_code == 413:
+        if status_code == 413:
             return _exceptions.RequestTooLargeError(err_msg, response=response, body=body)
 
-        if response.status_code == 422:
+        if status_code == 422:
             return _exceptions.UnprocessableEntityError(err_msg, response=response, body=body)
 
-        if response.status_code == 429:
+        if status_code == 429:
             return _exceptions.RateLimitError(err_msg, response=response, body=body)
 
-        if response.status_code == 529:
+        if status_code == 529:
             return _exceptions.OverloadedError(err_msg, response=response, body=body)
 
-        if response.status_code >= 500:
+        if status_code >= 500:
             return _exceptions.InternalServerError(err_msg, response=response, body=body)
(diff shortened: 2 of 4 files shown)


Zawwarsami16 (commented):
Direction is right — the raw `ValueError` was uncatchable through the SDK exception hierarchy and that needed fixing. But I think the status-code mapping the PR description claims will not actually fire with the current shape, and there is no test that would catch the regression. Two specific concerns and a minor one.

## 1. The error-subclass mapping is keyed on the HTTP response, not the SSE frame

The PR body says:

> The existing `Stream.__stream__` and `AsyncStream.__stream__` already handle `event == "error"` by calling `self._client._make_status_error(...)`, which maps the exception to the correct `APIStatusError` subclass (`InternalServerError`, `BadRequestError`, etc.).

That mapping lives in `_client.py` `_make_status_error` (and `BaseMantleClient._make_status_error` in `lib/bedrock/_mantle.py`), and both versions branch on `response.status_code`:

```python
if response.status_code == 400:
    return _exceptions.BadRequestError(...)
...
if response.status_code >= 500:
    return _exceptions.InternalServerError(...)
return APIStatusError(err_msg, response=response, body=body)
```

The `response` passed in is `self.response` from `Stream`/`AsyncStream`, which is the HTTP-level `httpx.Response`. For a Bedrock streaming call where the HTTP response was 200 OK and the error lives inside an SSE frame, `response.status_code` is 200, every branch above misses, and we land on the final fallback `APIStatusError`. So a Bedrock 400-frame yields a generic `APIStatusError` rather than `BadRequestError`, and a 500-frame yields generic `APIStatusError` rather than `InternalServerError`.

The end result is still catchable via `anthropic.APIError`, which is a genuine improvement over the previous `ValueError`. But the more specific subclass mapping the PR body promises will not happen.

Two cleanest fix paths I can see:

**(a)** Embed the Bedrock status code in the error body and override `BaseMantleClient._make_status_error` to consult it first. Then the SSE error decoder includes something like `{"_bedrock_status": 400, "type": "error", "error": {...}}`, and `_make_status_error` checks `isinstance(body, dict) and body.get("_bedrock_status")` before falling back to `response.status_code`.

**(b)** Raise the typed exception directly from `_parse_message_from_event` (or right after the decoder returns the error SSE), bypassing `_make_status_error` entirely for the Bedrock path. This requires the decoder to import `_exceptions` but keeps the routing local to the Bedrock code that already understands the frame format.

Either one is fine; (a) is the more general pattern.

## 2. No test coverage for the new path

The one file changed is the decoder itself; no new test exercises the non-200 frame path. The existing decoder tests presumably only cover 200 frames, so neither the old `ValueError` nor the new `ServerSentEvent` shape was protected against regression. A unit test with a synthesized `EventStreamMessage` carrying `status_code=400` + `:exception-type: validationException` + a JSON body would lock the contract.

## 3. Minor: bytes fallback formatting

In the JSON-decode fallback:

```python
except Exception:
    err_message = str(raw_body)
```

If `raw_body` is `bytes` and JSON parsing fails, this yields the bytes-literal repr (`b'{\n  "..."'`) inside the error message. `raw_body.decode(errors="replace")` (with a small isinstance guard) gives a cleaner string for the caller to read.

---

Happy to PR (a) or (b) on top of this branch if helpful, but worth confirming the maintainer's preference first since the choice affects how downstream callers catch the error.

nileshpatil6:
Thanks for the detailed review @Zawwarsami16 -- all three points addressed in the latest commit:

**1. Status-code-based exception mapping**

The approach: `_stream_decoder.py` now embeds `"_bedrock_status": <int>` in the error SSE body JSON. Both `BaseBedrockClient._make_status_error` (`_client.py`) and `BaseMantleClient._make_status_error` (`_mantle.py`) now check `body.get("_bedrock_status")` first and use that status code for exception mapping instead of `response.status_code` (which is always 200 for Bedrock streams). So a frame with status 429 raises `RateLimitError`, 400 raises `BadRequestError`, etc.

**2. Unit tests**

Added `tests/lib/test_bedrock_stream_decoder.py` covering:
- Non-200 frames emit `ServerSentEvent(event="error")` with `_bedrock_status` embedded
- 400/429/500 frame statuses map to `BadRequestError`/`RateLimitError`/`InternalServerError`
- Invalid UTF-8 body doesn't crash (uses `decode(errors="replace")`)
- Normal 200 path (no `_bedrock_status` in body) still uses `response.status_code`

**3. `str(raw_body)` fallback**

Changed to `raw_body.decode(errors="replace")` so the error message is readable text instead of `b'...'` repr.

Zawwarsami16 (approved):
LGTM @nileshpatil6 — all three points cleanly addressed.

The `_bedrock_status` overlay in `_make_status_error` is the right shape: `response.status_code` stays the default and only the Bedrock-frame status takes precedence when the body explicitly carries it, so the non-Bedrock paths in `_client.py` and `_mantle.py` are unchanged. The 200-path test (no `_bedrock_status` in body, falls back to `response.status_code`) makes that invariant explicit — that's the case most likely to regress silently if someone later refactors the overlay, so good to have it pinned.

One small nit, not blocking: `_bedrock_status` becomes a visible field in the exception's `body` (any caller pattern-matching on `e.body` will see it). A one-line doc comment in `_stream_decoder.py` flagging it as a private signaling key — not part of the Bedrock error contract — would save future readers from treating it as stable API. Pure documentation, no code change needed.

The `.decode(errors='replace')` swap on the bytes fallback is also the right call.

PR #1577, opened 2026-05-20
docs: clarify skill upload file layout

slegarraga:
## Summary
- document that skill upload paths must be rooted under a directory matching the `name` field in `SKILL.md`
- add a README file upload example and clarify the comprehensive agent example
- add regression coverage showing `anthropic.lib.files_from_dir()` preserves the skill top-level directory

Closes #1575.

## Tests
- `uv run pytest tests/lib/test_files.py -q`
- `uv run pytest tests/lib/test_files.py tests/lib/tools/test_skills.py -q`
- `uv run ruff check examples/agents_comprehensive.py src/anthropic/resources/beta/skills/skills.py src/anthropic/resources/beta/skills/versions.py src/anthropic/types/beta/skill_create_params.py src/anthropic/types/beta/skills/version_create_params.py tests/lib/test_files.py`
- `uv run pyright tests/lib/test_files.py`
- `git diff --check`

PR #1590, opened 2026-05-22
fix(skills): add normalize_skill_files() helper and document path naming convention

### LINKED ISSUE #1575: beta.skills.create: undocumented requirement that files path prefix must match SKILL.md name: field
### Summary

`beta.skills.create()` requires that every file path in the `files` parameter is prefixed with a top-level directory whose name exactly matches the `name:` field in `SKILL.md`. This requirement is not mentioned in the SDK docstring, the README, or any error message until after the first failed attempt. The error messages themselves require multiple iterations to decode.

### Reproduction

```python
import anthropic

SKILL_MD = b"""---
name: my-skill
description: Example skill.
---

Skill content here.
"""

# Fails with 400: "SKILL.md file must be exactly in the top-level folder."
client.beta.skills.create(
    display_title="my-skill",
    files=[("SKILL.md", SKILL_MD, "text/markdown")],
)
```

Second attempt after reading the error:

```python
# Fails with 400: "The folder name 'skills' must match the skill name 'my-skill' in SKILL.md."
client.beta.skills.create(
    display_title="my-skill",
    files=[("skills/SKILL.md", SKILL_MD, "text/markdown")],
)
```

### Expected behaviour

Either the API accepts `"SKILL.md"` as a valid top-level path, or the error on first failure clearly states the required format (e.g. `"Path must be '<name>/SKILL.md' where <name> matches the name: field in SKILL.md"`).

### What works

```python
# OK — directory name matches name: field in frontmatter exactly
client.beta.skills.create(
    display_title="my-skill",
    files=[("my-skill/SKILL.md", SKILL_MD, "text/markdown")],
)
```

## Suggested fix

Either:

1. **Document the requirement** in the `create()` docstring and/or README — a single sentence: *"The path of each file must be prefixed with a directory whose name matches the `name:` field in `SKILL.md`."*

2. **Handle it in the SDK** — parse the `name:` field from the supplied `SKILL.md` bytes and automatically rewrite bare filenames to `{name}/filename` before sending, removing the need for callers to know about this layout constraint.

### Environment

- `anthropic` SDK: 0.99.0
- Python: 3.14

ppradyoth:
## Summary

Closes #1575.

`beta.skills.create(files=...)` silently requires every path to be prefixed
with a top-level directory whose name **exactly matches** the `name:` field in
`SKILL.md` frontmatter (e.g. `my-skill/SKILL.md`, `my-skill/scripts/tool.py`).
When callers pass bare names like `"SKILL.md"` they receive a cryptic 400 with
no guidance on what went wrong.

## What this PR does

**Normalization is now automatic** — `Skills.create` / `AsyncSkills.create`
call `normalize_skill_files()` internally before building the multipart request,
so callers never need to know about the prefix requirement:

```python
# All three are now equivalent — the SDK fixes the paths transparently:
client.beta.skills.create(files=[("SKILL.md",        skill_md, "text/markdown")])
client.beta.skills.create(files=[("wrong/SKILL.md",  skill_md, "text/markdown")])
client.beta.skills.create(files=[("my-skill/SKILL.md", skill_md, "text/markdown")])
```

This makes upload symmetric with `_archive_top_dir` on the download side, which
already strips the wrapper directory when extracting — the constraint was always
there in both directions, but only enforced on upload.

### Name resolution order

1. `name:` field in `SKILL.md` YAML frontmatter *(primary)*
2. `display_title` normalised to `lowercase-with-hyphens` *(fallback when frontmatter omits `name:`)*
3. If neither is available the original file list is passed through unchanged and the API will surface a descriptive error

### Key advantages over the other PRs addressing #1575

| | This PR (#1590) | #1577 (docs only) | #1604 (auto-wire) |
|---|:---:|:---:|:---:|
| Auto-wired into `create()` | ✅ | ❌ | ✅ |
| Wrong prefix replaced (not double-prefixed) | ✅ | — | ❌ bug |
| `display_title` fallback | ✅ | — | ✅ |
| Idempotent (apply twice → same result) | ✅ | — | ✅ |
| PathLike file content | ✅ | — | ❌ |
| `normalize_skill_files()` exported for explicit use | ✅ | — | ❌ |

## Changes

- **`src/anthropic/lib/tools/_skills.py`** — `normalize_skill_files(files, *, display_title=None)` with wrong-prefix replacement, `display_title` fallback, and idempotency; `_parse_skill_name_from_frontmatter` returns `str | None` instead of raising
- **`src/anthropic/lib/tools/__init__.py`** — re-exports `normalize_skill_files` for callers who want to inspect normalized paths before uploading
- **`src/anthropic/resources/beta/skills/skills.py`** — auto-wires normalization into both `Skills.create` and `AsyncSkills.create`; updated `files=` docstring to say normalization is automatic
- **`tests/lib/tools/test_skills.py`** — 27 tests total (10 new: idempotency, display_title fallback, special-char normalization, 2-tuple entries, PathLike content, frontmatter returning None, wrong-prefix replacement)

## Test plan

```
uv run pytest tests/lib/tools/test_skills.py -v  # 27 passed
```

### DIFF
--- a/src/anthropic/lib/tools/__init__.py
+++ b/src/anthropic/lib/tools/__init__.py
@@ -1,3 +1,4 @@
+from ._skills import normalize_skill_files
 from ._beta_runner import BetaToolRunner, BetaAsyncToolRunner, BetaStreamingToolRunner, BetaAsyncStreamingToolRunner
 from ._beta_functions import (
     ToolError,
@@ -26,4 +27,5 @@ __all__ = [
     "BetaAbstractMemoryTool",
     "BetaAsyncAbstractMemoryTool",
     "ToolError",
+    "normalize_skill_files",
 ]
--- a/src/anthropic/lib/tools/_skills.py
+++ b/src/anthropic/lib/tools/_skills.py
@@ -43,31 +43,31 @@ def _read_file_entry_content(content: object) -> bytes | None:
         return None
 
 
-def _parse_skill_name_from_frontmatter(skill_md: bytes) -> str:
-    """Extract the ``name:`` field from SKILL.md YAML frontmatter.
-
-    Raises :exc:`ValueError` if the frontmatter block is absent or contains no
-    ``name:`` key.
-    """
+def _parse_skill_name_from_frontmatter(skill_md: bytes) -> str | None:
+    """Extract the ``name:`` field from SKILL.md YAML frontmatter, or ``None``."""
     text = skill_md.decode("utf-8", errors="replace")
     fm = re.search(r"^---\s*\n(.*?)\n---", text, re.DOTALL)
     if fm:
         name_match = re.search(r"^name:\s*(\S+)", fm.group(1), re.MULTILINE)
         if name_match:
             return name_match.group(1).strip()
-    raise ValueError("SKILL.md must contain a 'name:' field in its YAML frontmatter, e.g.:\n---\nname: my-skill\n---")
+    return None
 
 
-def normalize_skill_files(files: Sequence["FileTypes"]) -> list["FileTypes"]:
+def normalize_skill_files(
+    files: Sequence["FileTypes"],
+    *,
+    display_title: str | None = None,
+) -> list["FileTypes"]:
     """Normalize file paths for :meth:`~anthropic.resources.beta.Skills.create`.
 
-    The ``beta.skills.create()`` API requires every file path to be prefixed
-    with a top-level directory whose name exactly matches the ``name:`` field in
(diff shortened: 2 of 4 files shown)


dtmeadows-ant:
@ppradyoth thanks for this. Same reasoning as on #1575 itself: the path/name constraint is enforced and described by the API, and we'd rather see the validation message and docs improved there. Having `skills.create` silently re-root files before upload (including rewriting a `wrong/` prefix) could mask a genuine mistake. Not taking the helper for that reason, but the report landed where it needed to.

PR #1595, opened 2026-05-23
ci: migrate claude.yml to Workload Identity Federation auth

zenexer-ant:
Replaces the static `ANTHROPIC_API_KEY` repo secret with Workload Identity Federation: `claude-code-action` exchanges this job's GitHub OIDC token for a short-lived access token instead of reading a long-lived key.

- The federation rule is bound to this repository (repository_id-pinned), so the IDs below are repo-specific and not reusable elsewhere.
- Pins `claude-code-action` to v1.0.132 — the WIF inputs (`anthropic_federation_rule_id` etc.) landed in v1.0.130.
- `permissions: id-token: write` was already present.
- No `anthropic_workspace_id` — the rule targets the org's default workspace, so the input is optional (per the action's [docs/setup.md](https://github.com/anthropics/claude-code-action/blob/main/docs/setup.md)).

Once merged, the `ANTHROPIC_API_KEY` repo secret can be removed.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -37,9 +37,15 @@ jobs:
 
       - name: Run Claude Code
         id: claude
-        uses: anthropics/claude-code-action@v1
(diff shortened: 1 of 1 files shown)


dtmeadows-ant (approved):
:shipit:

PR #1604, opened 2026-05-27
fix(skills): auto-prefix upload paths with skill name directory

### LINKED ISSUE #1575: beta.skills.create: undocumented requirement that files path prefix must match SKILL.md name: field

daveCode-dot:
## Context

`beta.skills.create` requires every uploaded file path to be prefixed with a top-level directory whose name matches the `name:` field in `SKILL.md` (e.g. `my-skill/SKILL.md`, `my-skill/scripts/run.py`). This constraint is not documented in the method's docstring or parameter descriptions. Callers who pass bare paths (`"SKILL.md"`) receive a cryptic 400 response:

```
SKILL.md file must be exactly in the top-level folder
folder name scripts must match skill name my-skill
```

There is no hint in the error that the fix is a path prefix, and the required layout is only discoverable by trial-and-error or reading the download-side code (`_archive_top_dir` in `lib/tools/_skills.py`). Tracked in #1575.

Closes #1575.

## Change

**`src/anthropic/lib/tools/_skills.py`** — three new helpers:
- `_parse_skill_name_from_frontmatter(content)` — extracts the `name:` value from a SKILL.md YAML front-matter block.
- `_read_file_entry_bytes(content)` — reads raw bytes from any `FileContent` value (bytes, `PathLike`, `IOBase`), rewinding IO streams after reading.
- `normalize_skill_upload_paths(files, *, display_title)` — locates `SKILL.md` in the file list, resolves the skill name (front-matter → first already-prefixed path → `display_title` fallback), then prefixes every bare path. Already-prefixed paths pass through unchanged; returns original sequence if name cannot be determined.

**`src/anthropic/resources/beta/skills/skills.py`** — wires `normalize_skill_upload_paths` into both `Skills.create` and `AsyncSkills.create` before the multipart body is built.

## Why this approach

The fix is purely additive: it normalises paths before they reach the HTTP layer, so existing callers who already use the correct layout are unaffected (idempotent). The skill name is derived from the same `SKILL.md` front-matter field that the server validates against, keeping the client and server in sync without hard-coding any naming rules. The approach mirrors what `_archive_top_dir` + `_extract_skill_archive` already do on the download side, making upload/download symmetric.

Alternative considered: document the constraint in the docstring and raise a `ValueError` for bare paths. Rejected — a `ValueError` would break callers silently passing valid-looking paths, and documentation alone doesn't help interactive callers who don't read docstrings before hitting the 400.

## Tests

Added 10 new tests in `tests/lib/tools/test_skills.py` covering:
- `_parse_skill_name_from_frontmatter` found / missing
- Bare paths prefixed from SKILL.md `name:` field
- Already-prefixed paths unchanged (idempotent)
- `normalize_skill_upload_paths` idempotency (apply twice → same result)
- `display_title` fallback when SKILL.md has no `name:` field
- Special character normalisation in `display_title`
- IO stream (`BytesIO`) read + rewind
- No SKILL.md + no `display_title` → passthrough unchanged
- 2-tuple entries (filename, content) without MIME type

All 22 tests pass (`uv run pytest tests/lib/tools/test_skills.py -v`).

> **Note:** `tests/api_resources/beta/test_webhooks.py` fails with `ModuleNotFoundError: No module named 'standardwebhooks'` on the upstream `main` branch before any changes in this PR. Unrelated to this fix.

## Adjacent gaps

- The `SKILL.md` front-matter `name:` constraint on the server side is undocumented in `Skills.create`'s docstring and in the `SkillCreateParams` typed dict. A follow-up could add a one-line note to both. Left out of this PR to keep the diff focused on the behavioural fix.
- `normalize_skill_upload_paths` is exported from `__all__` in `_skills.py` so callers who build the file list outside `skills.create` can normalise manually if needed.

### DIFF
--- a/src/anthropic/lib/tools/_skills.py
+++ b/src/anthropic/lib/tools/_skills.py
@@ -7,13 +7,15 @@ tool implementations themselves.
 
 from __future__ import annotations
 
+import io
 import os
+import re
 import shutil
 import logging
 import tarfile
 import zipfile
 import tempfile
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Sequence
 from pathlib import Path, PurePosixPath
 from functools import partial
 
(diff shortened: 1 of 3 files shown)


avalyset (commented):
Thanks for taking this — as the original reporter of #1575, this is the right shape. The symmetry with `_archive_top_dir` is exactly what the issue was pointing at, and the idempotency check (`not startswith(prefix)`) handles the already-prefixed case cleanly.

One non-blocking observation on `_parse_skill_name_from_frontmatter`: the regex `^\s*name:\s*(.+?)\s*$` matches `name:` anywhere in the file, not just inside the `---` frontmatter block. A SKILL.md that contains `name:` in its body (e.g. inside a code example or prose) before or without a frontmatter block would parse the wrong value, prefix incorrectly, and surface the same cryptic 400 the issue is about — just from a different trigger. Scoping the match to the frontmatter block (parse up to the closing `---`, or require the opening delimiter) would close that gap. Everything else looks solid, including the BytesIO rewind and tuple-structure preservation in the tests.

daveCode-dot:
@avalyset good catch — you're right that the old regex matched `name:` anywhere, so a body occurrence (code example/prose) without a leading `---` block would parse the wrong value and surface the same 400 from a different trigger.

Pushed a fix scoping the match to the leading `---` ... `---` front-matter block (BOM and CRLF tolerated), plus two regression tests: a body-only `name:` with no front-matter (now returns `None`) and a body decoy after a real front-matter block (returns the front-matter value). Thanks for the careful review.

avalyset (commented):
Verified the fix against the head commit — this resolves it. The `\A`-anchored `_FRONTMATTER_RE` plus running the name match only on `block.group(1)` means a body `name:` is now physically outside the search scope, which is exactly the failure mode from the review. BOM (`\A﻿?`) and CRLF (`\r?\n` at all three fence boundaries) are both handled, and the two regression tests cover the cases directly: body-only → None, and a decoy after real front-matter → the front-matter value. LGTM from my side.

One non-blocking note, just to confirm it's intentional: the opening fence requires a newline (`---[ \t]*\r?\n`), so a single-line `---name: x` won't match — which is correct, since that isn't valid front-matter, but flagging in case you want a test pinning that behaviour. Nice and tight overall.

avalyset:
This PR addresses #1575 but only mentions it — adding `Closes #1575` to the description would auto-close and link the issue on merge.

dtmeadows-ant:
@daveCode-dot thanks, and thanks for iterating on the front-matter parsing with the reporter. We closed #1575 as an API-side problem: the layout rule and its error messages live in the API, and that's where we want it addressed. Having the Python client quietly rewrite upload paths would hide what the API actually requires and behave differently from the other SDKs. I'm going to pass on the client-side normalization for that reason.

Issue #1598, opened 2026-05-23
[Bug] API returns 400 'Invalid signature in thinking block' when replaying conversation history — no recovery path


craigderrick:
## Bug Description

When using extended thinking with claude-sonnet-4.x/claude-opus-4.x in a multi-turn conversation, the API begins returning `400: Invalid `signature` in `thinking` block` after N turns. Once triggered, **every subsequent message in the session fails permanently** — there is no recovery without manually stripping thinking blocks from history.

## Reproduction

1. Enable extended thinking on claude-sonnet-4-6 or claude-opus-4
2. Run a long multi-turn conversation (20+ turns with tool use)
3. Pass full conversation history on each turn (including `type: "thinking"` blocks from prior assistant messages)
4. At some turn, the API begins rejecting with 400: `messages.X.content.Y: Invalid `signature` in `thinking` block`
5. The position (X, Y) is consistent — the same thinking block triggers it every time
6. The session cannot recover — every subsequent call with that history fails

## Error

```json
{
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "messages.27.content.0: Invalid `signature` in `thinking` block"
  }
}
```

## Questions / Issues

1. **What invalidates a signature?** The API docs say to pass thinking blocks back unmodified. If signatures expire or become invalidated server-side, clients have no way to know without trying.
2. **No recovery path** — the error is a hard 400, not retryable. Clients must strip thinking blocks from history to continue, but the API docs don't describe this scenario or recommend this workaround.
3. **Cascading failure** — subagents or forked sessions that inherit the conversation history also fail immediately.

## Environment

- SDK: used via OpenClaw gateway (direct Anthropic REST API)
- Model: claude-sonnet-4-6
- Platform: macOS 26.4.1, Node.js 22.x

## Related

- anthropics/claude-code#21726 (session resume fails with same error)
- anthropics/claude-code#10627

## Requested

1. Document under what conditions a thinking block signature becomes invalid
2. Provide a recommended recovery path (e.g., strip thinking blocks from history on 400, retry)
3. Ideally: return a more specific error type so clients can handle it programmatically

YorickLane:
This is exactly the kind of failure mode that's hard to handle gracefully — the API contract ("pass thinking blocks back unmodified") discourages exactly the defensive code that would save you here. The "no recovery path" framing is the killer.

[...]

PR #1605, opened 2026-05-27
feat(lib): add strip_thinking_blocks utility for conversation history recovery

daveCode-dot:
## Context

Multi-turn conversations with extended thinking accumulate `ThinkingBlock` entries in the assistant messages. When those messages are stored and replayed later — after session expiry, model upgrades, or history reconstruction from a database — the signatures become invalid and every subsequent call fails permanently with:

```
400 Invalid `signature` in `thinking` block
```

The API documentation says to pass thinking blocks back unmodified, but does not describe what happens when signatures expire or how to recover. The only escape is to strip the thinking blocks from history, but there is no SDK utility for this and no documented procedure. Tracked in #1598.

## Change

**`src/anthropic/lib/_thinking.py`** — new module with:
- `strip_thinking_blocks(messages)` — accepts `Iterable[MessageParam]`, returns `list[MessageParam]` with every `type: "thinking"` and `type: "redacted_thinking"` content block removed. String content is left unchanged. All other message-level keys (`role`, etc.) are copied verbatim. The function works with both plain dicts (TypedDict values) and `BaseModel` instances returned by the SDK. Idempotent; does not mutate the input list.

**`src/anthropic/lib/__init__.py`** — exports `strip_thinking_blocks` alongside `files_from_dir` so callers can reach it via `from anthropic.lib import strip_thinking_blocks`.

Intended usage pattern:

```python
try:
    response = client.messages.create(model=model, messages=history, ...)
except anthropic.BadRequestError as exc:
    if "Invalid `signature` in `thinking` block" in str(exc):
        history = strip_thinking_blocks(history)
        response = client.messages.create(model=model, messages=history, ...)
    else:
        raise
```

## Why this approach

The fix lives in `lib/` as a pure utility, symmetric with the existing `files_from_dir` helper: it has no I/O, no API calls, and no dependencies beyond the SDK's own type definitions. It does not add any automatic behaviour to the message-sending path — the caller decides when stripping is appropriate (e.g. only on error recovery, not on every call) — keeping the SDK's default behaviour unchanged for callers who never encounter expired signatures.

Alternative considered: automatically strip thinking blocks when a 400 "Invalid signature" is received and retry the request transparently. Rejected because silent retry changes observable behaviour (thinking blocks are dropped without the caller's knowledge), complicates streaming retry semantics, and may not be what callers want in all scenarios.

## Tests

Added 14 tests in `tests/lib/test_thinking.py` covering:
- String-content messages pass through unchanged
- `thinking` blocks removed from assistant messages
- `redacted_thinking` blocks removed
- Mixed content (`thinking` + text + `tool_use`): only thinking types removed
- User messages without thinking: unchanged
- Non-thinking blocks preserved verbatim
- All-thinking content produces empty list (edge case)
- Empty messages list
- Multi-message conversations processed independently
- `BaseModel` instances (`_FakeThinkingBlock`, `_FakeRedactedBlock`): stripped via `getattr`
- Does not mutate the original input list
- Role and other message keys preserved
- Idempotent (apply twice → same result)

All 14 tests pass (`uv run pytest tests/lib/test_thinking.py -v`).

> **Note:** `tests/api_resources/beta/test_webhooks.py` fails with `ModuleNotFoundError: No module named 'standardwebhooks'` on the upstream `main` branch before any changes in this PR. Unrelated to this fix.

## Adjacent gaps

- The `BadRequestError` message for expired thinking signatures does not suggest `strip_thinking_blocks` as a recovery path. A follow-up could add a note in the exception handling layer or in the extended-thinking documentation. Left out of this PR to keep the diff focused.
- `strip_thinking_blocks` is not yet re-exported from the top-level `anthropic` namespace (`from anthropic import strip_thinking_blocks`). If maintainers prefer it to be a first-class export, that would be a one-line addition to `src/anthropic/__init__.py`.

### DIFF
--- a/src/anthropic/lib/__init__.py
+++ b/src/anthropic/lib/__init__.py
@@ -1,1 +1,2 @@
 from ._files import files_from_dir as files_from_dir, async_files_from_dir as async_files_from_dir
+from ._thinking import strip_thinking_blocks as strip_thinking_blocks
--- a/src/anthropic/lib/_thinking.py
+++ b/src/anthropic/lib/_thinking.py
@@ -1,0 +1,82 @@
+"""Utilities for managing extended thinking blocks in conversation history.
+
+Extended thinking blocks carry a ``signature`` field that is valid only for
+the specific conversation context they were created in.  Long multi-turn
+sessions or any operation that reconstructs history from stored messages can
+hit a 400 "Invalid ``signature`` in ``thinking`` block" error because the
+server-side context that produced the signature no longer matches.
+``strip_thinking_blocks`` removes those blocks before the history is replayed,
+providing the only documented recovery path.
+"""
+
+from __future__ import annotations
+
+from typing import Iterable
+
+from ..types.message_param import MessageParam
+
+__all__ = ["strip_thinking_blocks"]
+
+_THINKING_TYPES = frozenset({"thinking", "redacted_thinking"})
+
+
+def _block_type(block: object) -> str:
+    """Return the ``type`` field of a content block regardless of representation.
+
+    Accepts plain dicts (TypedDict values, JSON-decoded payloads) and Pydantic
+    ``BaseModel`` instances (``ThinkingBlock``, ``RedactedThinkingBlock``, …).
+    Returns ``""`` for anything that carries no ``type``.
+    """
+    if isinstance(block, dict):
+        return block.get("type", "")
+    return getattr(block, "type", "")
+
+
+def strip_thinking_blocks(
+    messages: Iterable[MessageParam],
+) -> list[MessageParam]:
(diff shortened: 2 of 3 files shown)


daveCode-dot:
Gentle ping — happy to address any feedback or make changes if this is in the review queue. Let me know if there's anything blocking from your side.

dtmeadows-ant:
@daveCode-dot thanks. I followed up on #1598 directly: thinking signatures don't expire, and that 400 means the blocks reaching the API differ from what was returned. A helper that strips thinking from history would mostly paper over whatever is altering them, and it costs the model its prior reasoning. I'd rather find the actual cause there than ship the workaround as a library function. Closing this one on that basis.

PR #1608, opened 2026-05-28
Fix Bedrock region detection to respect aws_profile and AWS_DEFAULT_REGION

Sarmkadan:
## Summary

Fixes #892

The Bedrock client's `_infer_region()` function had two issues causing incorrect region detection:

- **`aws_profile` not forwarded**: When a user passed `aws_profile="my-profile"` to `AnthropicBedrock()`, the profile parameter was not passed through to `_infer_region()`, so `boto3.Session()` was created without the profile name. This meant the profile-specific region from `~/.aws/config` was ignored, and the client would fall back to `us-east-1` instead of using the profile's configured region. (The `AWS_PROFILE` env var still worked because boto3 reads it internally, but the explicit constructor parameter did not.)

- **`AWS_DEFAULT_REGION` not checked**: The function only checked the `AWS_REGION` env var but not `AWS_DEFAULT_REGION`, which is also supported by boto3 and the AWS CLI.

### Changes

- `_infer_region()` now accepts an `aws_profile` parameter and passes it to `boto3.Session(profile_name=...)`
- `AWS_DEFAULT_REGION` is checked as a fallback when `AWS_REGION` is unset
- `self.aws_profile` is assigned before `_infer_region()` is called so the profile is available
- Removed redundant `or "us-east-1"` fallbacks in `_prepare_request` (both sync and async) since `_infer_region()` already guarantees a non-None return

### Resolution order (updated)

1. `AWS_REGION` environment variable
2. `AWS_DEFAULT_REGION` environment variable
3. Region from `boto3.Session` (respects `AWS_PROFILE` env var, `~/.aws/config`, and the explicit `aws_profile` constructor parameter)
4. Hard-coded `us-east-1` (legacy fallback)

## Test plan

- [x] Added `test_region_infer_from_aws_default_region` - verifies `AWS_DEFAULT_REGION` is respected
- [x] Added `test_aws_region_takes_precedence_over_aws_default_region` - verifies `AWS_REGION` wins over `AWS_DEFAULT_REGION`
- [x] Added `test_region_infer_from_explicit_aws_profile_param` - verifies the `aws_profile` constructor param is used for region inference (sync client)
- [x] Added `test_region_infer_from_explicit_aws_profile_param_async` - same for async client
- [x] All 16 existing + new tests pass

### DIFF
--- a/src/anthropic/lib/bedrock/_client.py
+++ b/src/anthropic/lib/bedrock/_client.py
@@ -67,17 +67,24 @@ def _prepare_options(input_options: FinalRequestOptions) -> FinalRequestOptions:
     return options
 
 
-def _infer_region() -> str:
+def _infer_region(aws_profile: str | None = None) -> str:
     """
     Infer the AWS region from the environment variables or
     from the boto3 session if available.
+
+    The resolution order is:
+    1. ``AWS_REGION`` environment variable
+    2. ``AWS_DEFAULT_REGION`` environment variable
+    3. Region resolved by a ``boto3.Session`` (respects ``AWS_PROFILE``,
+       ``~/.aws/config``, and the explicit *aws_profile* parameter)
+    4. Hard-coded ``us-east-1`` fallback (legacy behaviour)
     """
-    aws_region = os.environ.get("AWS_REGION")
+    aws_region = os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION")
     if aws_region is None:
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@Sarmkadan thanks for this. The underlying issue (#892) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1620, opened 2026-05-31
fix(_models): guard against bare dict in construct_type()

### LINKED ISSUE #1619: BUG: construct_type() crashes with ValueError when type_ is a bare dict
## What breaks

\`\`\`
\`\`\`

Traceback:
\`\`\`
    ^^^^^^^^^^^^^
\`\`\`

## How to trigger

\`\`\`python
\`\`\`

## Root cause

Line 650 in \`_models.py\`:
\`\`\`python
\`\`\`

## Fix

devteamaegis:
## What's broken

Calling `construct_type()` with `type_=dict` (a bare, unparameterised dict) raises `ValueError: not enough values to unpack (expected 2, got 0)`. This hits whenever a model field is annotated as plain `dict` rather than `Dict[str, SomeType]`.

```
```

## Why it happens

The `origin == dict` branch unconditionally unpacks `get_args(type_)` as exactly two values. For a bare `dict` with no type parameters, `get_args` returns an empty tuple `()`, making the two-value unpack fail.

## Fix

Capture `get_args(type_)` into a local `args` variable and fall back to `object` as `items_type` when `args` is empty. When `items_type` is `object`, every value passes through `construct_type` unchanged, which is the correct behaviour for an untyped dict.

## Test

Added `test_construct_type_bare_dict` in `tests/test_models.py` — passes `type_=dict` and asserts the original dict is returned without raising.

Fixes #1619

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -647,7 +647,8 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        args = get_args(type_)
+        items_type = args[1] if args else object  # Dict[_, items_type]; plain dict has no args
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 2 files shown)


craigie-ant:
👋 thanks for the PR! closing because of https://github.com/anthropics/anthropic-sdk-python/issues/1619#issuecomment-5257021164

PR #1627, opened 2026-06-01
fix(_models): guard against IndexError in construct_type for bare list

### LINKED ISSUE #1626: BUG: construct_type() raises IndexError when called with bare unparameterized list

Minimal repro:

\`\`\`python
\`\`\`

Full traceback:

\`\`\`
                 ~~~~^^^
\`\`\`

devteamaegis:
## What's broken

Calling `construct_type(value=[...], type_=list)` with a bare, unparameterized `list` raises `IndexError: tuple index out of range`. The `origin == list` branch at line 671 of `src/anthropic/_models.py` unconditionally executes `inner_type = args[0]`, but `get_args(list)` returns an empty tuple when no type parameter is present.

## Why it happens

`get_origin(list)` returns `list` (matching the branch), but `get_args(list)` returns `()` rather than a one-element tuple, so `args[0]` is always out of range for unparameterized bare `list`.

## Fix

Changed `inner_type = args[0]` to `inner_type = args[0] if args else object`. When no type parameter is present, elements are returned unchanged (identity via `construct_type(value=entry, type_=object)`), which is the correct safe fallback.

## Test

Added `test_construct_type_bare_list` in `tests/test_models.py` confirming that `construct_type(value=["a", "b", "c"], type_=list)` returns `["a", "b", "c"]` without raising.

Fixes #1626

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -668,7 +668,7 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_list(value):
             return value
 
-        inner_type = args[0]  # List[inner_type]
+        inner_type = args[0] if args else object  # List[inner_type] or bare list
         return [construct_type(value=entry, type_=inner_type) for entry in value]
 
     if origin == float:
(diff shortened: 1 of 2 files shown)


craigie-ant:
👋 thanks for the PR! closing because of https://github.com/anthropics/anthropic-sdk-python/issues/1626#issuecomment-5257036785

PR #1629, opened 2026-06-01
fix(_transform): guard bare dict annotation to avoid IndexError

### LINKED ISSUE #1628: IndexError on bare (unparameterized) dict annotation in _transform_recursive
## Bug

Calling `transform(data, dict)` with a bare `dict` annotation (not `Dict[K, V]`) raises `IndexError: tuple index out of range` in both `_transform_recursive` (line 183) and `_async_transform_recursive` (line 349).

```
  File "_transform.py", line 183, in _transform_recursive
    items_type = get_args(stripped_type)[1]
```

## Root cause

`get_args(dict)` returns `()`, so indexing `[1]` fails unconditionally when the annotation is unparameterized.

## Reproduction

```python
transform({"key": "value"}, dict)  # raises IndexError
```

## Relation to existing issues

Issue #1619 tracks the same pattern in `_models.py:650` (`construct_type`). This is the corresponding crash in `_transform.py` and is not covered there.

devteamaegis:
## What's broken

`transform({"key": "value"}, dict)` raises `IndexError: tuple index out of range` in `_transform_recursive` (line 183) and `_async_transform_recursive` (line 349). Both branches unconditionally access `get_args(stripped_type)[1]` to get the value type. When the annotation is a bare `dict` rather than `Dict[K, V]`, `get_args(dict)` returns `()`, so the index fails.

```
  File "_transform.py", line 183, in _transform_recursive
    items_type = get_args(stripped_type)[1]
```

## Why it happens

`get_args` on an unparameterized `dict` returns an empty tuple, and the code assumed at least two type arguments were always present.

## Fix

Added a `len(args) >= 2` guard before indexing. Falls back to `Any` for bare dicts, matching the same pattern already used in `_models.py`. Both sync and async variants are fixed.

## Test

Added `test_transform_recursive_bare_dict_no_index_error` which calls `transform({"key": "value"}, dict)` for both sync and async paths and asserts no exception and correct output.

Fixes #1628

### DIFF
--- a/src/anthropic/_utils/_transform.py
+++ b/src/anthropic/_utils/_transform.py
@@ -180,7 +180,8 @@ def _transform_recursive(
         return _transform_typeddict(data, stripped_type)
 
     if origin == dict and is_mapping(data):
-        items_type = get_args(stripped_type)[1]
+        args = get_args(stripped_type)
+        items_type = args[1] if len(args) >= 2 else Any
         return {key: _transform_recursive(value, annotation=items_type) for key, value in data.items()}
 
     if (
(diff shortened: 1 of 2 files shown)


craigie-ant:

PR #1639, opened 2026-06-03
fix: guard against bare unparameterized dict/list in construct_type and transform

### LINKED ISSUE #1619: BUG: construct_type() crashes with ValueError when type_ is a bare dict

kiankhooban:
`get_args(dict)` and `get_args(list)` return empty tuples when no type parameters are present. Four sites were unconditionally indexing into those tuples, raising `ValueError` or `IndexError` when a model field is annotated as a bare `dict` or `list` instead of `Dict[K, V]` or `List[T]`.

## Crash sites fixed

**`src/anthropic/_models.py`**
- `construct_type`: `_, items_type = get_args(type_)` → `ValueError` on bare `dict` (issue #1619)
- `construct_type`: `inner_type = args[0]` → `IndexError` on bare `list` (issue #1626)

**`src/anthropic/_utils/_transform.py`**
- `_transform_recursive` (sync): `get_args(stripped_type)[1]` → `IndexError` on bare `dict` (issue #1628)
- `_async_transform_recursive`: same pattern, same fix

## Fix

Guard each index access with a length check and fall back to `object` when no type args are present — preserving the value as-is, which is the correct behaviour for unparameterized containers.

## Reproduction

```python
construct_type(value={"key": "value"}, type_=dict)  # was: ValueError
construct_type(value=["a", "b"], type_=list)         # was: IndexError

transform({"key": "value"}, dict)                    # was: IndexError
```

Fixes #1619, #1626, #1628

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -647,7 +647,7 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        items_type = args[1] if len(args) >= 2 else object  # Dict[_, items_type]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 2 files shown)


craigie-ant:

PR #1641, opened 2026-06-04
fix(client): treat empty-string env credentials as unset

### LINKED ISSUE #1640: `Anthropic()` emits a malformed `Authorization: Bearer ` header when `ANTHROPIC_AUTH_TOKEN` is an empty string
## Summary
When `ANTHROPIC_AUTH_TOKEN` (or `ANTHROPIC_API_KEY`) is present in the environment but set to an **empty string**, the SDK treats the empty string as a real credential instead of as "absent". It then builds an `Authorization: Bearer ` header (literally `Bearer` + a trailing space, from the empty token). The underlying HTTP layer (h11, via httpx) validates header values at write time and rejects the trailing space, so **every request fails** with:

```
h11._util.LocalProtocolError: Illegal header value b'Bearer '
```

surfaced by the SDK as `anthropic.APIConnectionError`. Empty-string env vars are common (CI, containers, `export VAR=`, wrapper shells), so this turns a benign/empty credential into a hard, confusing failure. The conventional behavior is to treat an empty-string credential as unset.

This is reliably triggered when running inside the **Claude Code** CLI shell, which exports `ANTHROPIC_AUTH_TOKEN=` and `ANTHROPIC_API_KEY=` (both empty) into the environment — so any script there that constructs `Anthropic()` without an explicit non-empty key fails on every request.

## Environment
- `anthropic` **0.103.1** (observed). The same code path is present on **main** (→ latest 0.105.2), so upgrading does not fix it.
- `httpx` 0.28.1, `h11` 0.16.0, Python 3.14.2, Windows 11.

## Reproduction (self-contained, no network — full script in the collapsible section below)
```python
import os
os.environ["ANTHROPIC_AUTH_TOKEN"] = ""   # present but empty
os.environ["ANTHROPIC_API_KEY"] = ""      # present but empty
from anthropic import Anthropic

client = Anthropic()                       # no explicit credentials
print(repr(client.auth_token))             # ''
print(dict(client.auth_headers))           # {'X-Api-Key': '', 'Authorization': 'Bearer '}
```
`client.auth_headers["Authorization"]` is `'Bearer '` (length 7, trailing space). Validating that exact value reproduces the error with no network call:
```python
from h11._headers import normalize_and_validate
normalize_and_validate([("Authorization", "Bearer ")])
# h11._util.LocalProtocolError: Illegal header value b'Bearer '
# ('Bearer x' validates fine; only the trailing space from the empty token is rejected.)
```
On a real call the failure surfaces at write time (after TCP connect), wrapped as `anthropic.APIConnectionError`.

## Root cause
The SDK guards credential presence with `is None`, so an empty string passes every guard:

1. Env read — no empty-string coercion ([`_client.py#L211-L212`](https://github.com/anthropics/anthropic-sdk-python/blob/ddd43b7ceb79f433dfafe488d95f48c22801186b/src/anthropic/_client.py#L211-L212)):
   ```python
   api_key = os.environ.get("ANTHROPIC_API_KEY")      # -> "" (not None)
   auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") # -> "" (not None)
   ```
2. "Missing auth" fallback uses `is None`, so it never triggers ([`_client.py#L258`](https://github.com/anthropics/anthropic-sdk-python/blob/ddd43b7ceb79f433dfafe488d95f48c22801186b/src/anthropic/_client.py#L258)):
   ```python
   if credentials is None and api_key is None and auth_token is None and _is_base_client(self):
   ```
3. Header builders emit on `is not None` ([`_client.py#L334-L351`](https://github.com/anthropics/anthropic-sdk-python/blob/ddd43b7ceb79f433dfafe488d95f48c22801186b/src/anthropic/_client.py#L334-L351)):
   ```python
   @property
   def _api_key_auth(self) -> dict[str, str]:
       api_key = self.api_key
       if api_key is None:
           return {}
       return {"X-Api-Key": api_key}        # -> {"X-Api-Key": ""}

   @property
   def _bearer_auth(self) -> dict[str, str]:
       auth_token = self.auth_token
       if auth_token is None:
           return {}
       return {"Authorization": f"Bearer {auth_token}"}   # -> {"Authorization": "Bearer "}
   ```
   (The `_bearer_auth` comment says "always emit if `self.auth_token` is set" — but "set" is implemented as `is not None`, which also matches `""`.)

## Suggested fix
Treat an empty-string credential as absent. Most robust at the env read (covers all downstream guards):
```python
api_key = os.environ.get("ANTHROPIC_API_KEY") or None
auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") or None
```
and/or change the guards at L258 / L336 / L349 to truthiness (`if not api_key:` / `if not auth_token:`). With either change, an empty-string env credential falls through to the normal "no credential" path (or `default_credentials`), instead of producing a malformed header.

## Workaround (callers, today)
- Pass a non-empty credential explicitly: `Anthropic(api_key=MY_REAL_KEY)` — this sets `has_explicit_credential`, the empty env vars are not read, and only `X-Api-Key` is sent (confirmed in the repro). Note: passing `api_key=""` explicitly does **not** help (still empty); pass a real key, or
- Clear the empty vars before constructing the client:
  ```python
  for k in ("ANTHROPIC_AUTH_TOKEN", "ANTHROPIC_API_KEY"):
      if os.environ.get(k) == "":
          os.environ.pop(k)
  ```

<details>
<summary>Full repro script (<code>repro_sdk_empty_bearer.py</code>, no network)</summary>

```python
"""
Minimal, self-contained, NO-NETWORK repro for:
  anthropic-sdk-python emits a malformed `Authorization: Bearer ` header when
  ANTHROPIC_AUTH_TOKEN (or ANTHROPIC_API_KEY) is set to an EMPTY STRING.

Sets the env vars to "" itself, so it reproduces on any machine (you do NOT need
to run it inside a Claude Code shell). Makes no API calls.

    pip install anthropic
    python repro_sdk_empty_bearer.py

Observed: anthropic 0.103.1 (bug also present on main -> 0.105.2),
          httpx 0.28.1, h11 0.16.0, Python 3.14.
"""
import os

# Simulate the environment that triggers the bug: credentials present-but-empty.
# (e.g. a parent process exported `ANTHROPIC_AUTH_TOKEN=` / `ANTHROPIC_API_KEY=`.)
os.environ["ANTHROPIC_AUTH_TOKEN"] = ""
os.environ["ANTHROPIC_API_KEY"] = ""

from anthropic import Anthropic  # noqa: E402

client = Anthropic()  # no explicit credentials -> reads the empty env vars

print("self.api_key   :", repr(client.api_key))      # -> ''
print("self.auth_token:", repr(client.auth_token))   # -> ''
print("auth_headers   :", dict(client.auth_headers))
# -> {'X-Api-Key': '', 'Authorization': 'Bearer '}    <-- note the trailing space

auth = client.auth_headers.get("Authorization")
print("Authorization repr:", repr(auth), "(len", len(auth or ""), ")")

# Why every request then fails (shown WITHOUT a network call): the underlying
# HTTP layer (h11, via httpx) validates header values at write time and rejects
# the trailing space in 'Bearer '.
print("\n-- h11 header validation of the value the SDK produced --")
try:
    from h11._headers import normalize_and_validate
    normalize_and_validate([("Authorization", auth)])
    print("  unexpectedly OK")
except Exception as e:
    print(f"  {type(e).__module__}.{type(e).__name__}: {e}")
    # -> h11._util.LocalProtocolError: Illegal header value b'Bearer '

print("\nExpected real-call failure: anthropic.APIConnectionError wrapping the above"
      "\n('Illegal header value b\\'Bearer \\'') - raised on the first request.")

# The fix on the caller side (no bug if a non-empty credential is passed explicitly):
print("\n-- workaround: explicit non-empty api_key --")
fixed = Anthropic(api_key="sk-ant-DUMMY-not-real")
print("auth_headers   :", {k: ("<set>" if v else repr(v)) for k, v in fixed.auth_headers.items()})
# -> {'X-Api-Key': '<set>'}   (no Authorization header, no malformed Bearer)
```
</details>

LiFeiYu-boost:
## Summary

Fixes #1640.

When `ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` are present in the environment but set to an **empty string** (common in CI, containers, and shells that `export VAR=` — including the Claude Code CLI shell), the client treated `""` as a real credential because every guard checks `is None`. It then built a malformed `Authorization: Bearer ` header (trailing space), which h11 rejects at request-write time — so every request failed with `APIConnectionError`.

## Change

Coerce empty-string env credentials to `None` at the read site, in both `Anthropic.__init__` and `AsyncAnthropic.__init__`:

```python
api_key = os.environ.get("ANTHROPIC_API_KEY") or None
auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") or None
```

Empty env credentials now fall through to the normal no-credential path: `default_credentials()` auto-discovery (profile / federation), or a clear `Could not resolve authentication method` error — instead of emitting a malformed header.

This also makes `_client.py` consistent with the existing credential chain: `lib/credentials/_chain.py` already treats empty env vars as absent via truthy checks (`if os.environ.get(ENV_API_KEY):`), and uses the same `or None` idiom for `ANTHROPIC_WORKSPACE_ID`.

Explicitly-passed `api_key=""` / `auth_token=""` constructor arguments are intentionally left unchanged — this PR only normalizes the implicit env-var read.

## Tests

Added `test_empty_env_credentials_treated_as_unset` to both `TestAnthropic` and `TestAsyncAnthropic`, mirroring the existing `test_validate_headers` pattern (with `default_credentials` mocked out): asserts `api_key` / `auth_token` resolve to `None`, `auth_headers` is empty (previously `{"X-Api-Key": "", "Authorization": "Bearer "}`), and request building raises the standard auth-resolution `TypeError`.

`tests/test_client.py` passes; `ruff check`, `ruff format --check`, and `pyright` are clean on the touched files.

### DIFF
--- a/src/anthropic/_client.py
+++ b/src/anthropic/_client.py
@@ -208,8 +208,10 @@ class Anthropic(SyncAPIClient):
             or profile is not None
         )
         if not has_explicit_credential:
-            api_key = os.environ.get("ANTHROPIC_API_KEY")
-            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN")
+            # An empty-string env credential is treated as unset — it would
+            # otherwise produce a malformed `Authorization: Bearer ` header.
+            api_key = os.environ.get("ANTHROPIC_API_KEY") or None
+            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") or None
         self.api_key = api_key
         self.auth_token = auth_token
         # --- end credentials support ---
@@ -610,8 +612,10 @@ class AsyncAnthropic(AsyncAPIClient):
             or profile is not None
         )
         if not has_explicit_credential:
-            api_key = os.environ.get("ANTHROPIC_API_KEY")
-            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN")
+            # An empty-string env credential is treated as unset — it would
+            # otherwise produce a malformed `Authorization: Bearer ` header.
+            api_key = os.environ.get("ANTHROPIC_API_KEY") or None
+            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") or None
         self.api_key = api_key
         self.auth_token = auth_token
         # --- end credentials support ---
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -434,6 +434,23 @@ class TestAnthropic:
         request2 = client2._build_request(FinalRequestOptions(method="get", url="/foo", headers={"X-Api-Key": Omit()}))
         assert request2.headers.get("X-Api-Key") is None
 
+    def test_empty_env_credentials_treated_as_unset(self) -> None:
+        with mock.patch("anthropic._client.default_credentials", return_value=None):
+            with update_env(ANTHROPIC_API_KEY="", ANTHROPIC_AUTH_TOKEN=""):
+                client = Anthropic(base_url=base_url, _strict_response_validation=True)
+
+        assert client.api_key is None
+        assert client.auth_token is None
+        # previously produced {'X-Api-Key': '', 'Authorization': 'Bearer '} —
+        # the trailing space is rejected by h11 at request-write time
+        assert client.auth_headers == {}
+
+        with pytest.raises(
+            TypeError,
+            match="Could not resolve authentication method. Expected one of api_key, auth_token, or credentials to be set. Or for one of the `X-Api-Key` or `Authorization` headers to be explicitly omitted",
+        ):
+            client._build_request(FinalRequestOptions(method="get", url="/foo"))
+
     def test_default_query_option(self) -> None:
         client = Anthropic(
             base_url=base_url, api_key=api_key, _strict_response_validation=True, default_query={"query_param": "bar"}
@@ -1463,6 +1480,23 @@ class TestAsyncAnthropic:
 
         request2 = client2._build_request(FinalRequestOptions(method="get", url="/foo", headers={"X-Api-Key": Omit()}))
         assert request2.headers.get("X-Api-Key") is None
+
+    def test_empty_env_credentials_treated_as_unset(self) -> None:
+        with mock.patch("anthropic._client.default_credentials", return_value=None):
+            with update_env(ANTHROPIC_API_KEY="", ANTHROPIC_AUTH_TOKEN=""):
+                client = AsyncAnthropic(base_url=base_url, _strict_response_validation=True)
+
+        assert client.api_key is None
+        assert client.auth_token is None
+        # previously produced {'X-Api-Key': '', 'Authorization': 'Bearer '} —
+        # the trailing space is rejected by h11 at request-write time
+        assert client.auth_headers == {}
+
+        with pytest.raises(
+            TypeError,
+            match="Could not resolve authentication method. Expected one of api_key, auth_token, or credentials to be set. Or for one of the `X-Api-Key` or `Authorization` headers to be explicitly omitted",
+        ):
+            client._build_request(FinalRequestOptions(method="get", url="/foo"))
 
     async def test_default_query_option(self) -> None:
         client = AsyncAnthropic(


hiroki-tamba-research:
Thanks for the fast fix — this resolves the reported (env) case cleanly, and coercing at the read site is the robust choice: it covers every downstream guard (the missing-auth check and both `_api_key_auth` / `_bearer_auth`) in one place, for both sync and async. Tests over `ANTHROPIC_API_KEY` + `ANTHROPIC_AUTH_TOKEN` are 👍.

One optional, non-blocking note for defense-in-depth: an explicitly-passed empty string still slips through, since it skips the env-read coercion (`has_explicit_credential` is `is not None`) and `_bearer_auth` guards on `is None`:

```python
>>> from anthropic import Anthropic
>>> dict(Anthropic(auth_token="").auth_headers)
{'Authorization': 'Bearer '}   # same malformed header
```

Making the `auth_headers` builders truthy (`if not auth_token` / `if not api_key`) would close that path too. Not needed for #1640 (the env case) — just flagging while we're here.

tomer-ant:
Thanks, but this is fixed in v0.122.0 (https://github.com/anthropics/anthropic-sdk-python/commit/76a2e68531d50c9043a5ac58919527287e56a842). Empty ANTHROPIC_API_KEY / ANTHROPIC_AUTH_TOKEN values are now treated as unset.

PR #1643, opened 2026-06-04
Fix construct_type() crash on bare list/dict types

### LINKED ISSUE #1619: BUG: construct_type() crashes with ValueError when type_ is a bare dict

### LINKED ISSUE #1626: BUG: construct_type() raises IndexError when called with bare unparameterized list

PHclaw:
## Summary

\construct_type()\ crashes when given bare generic types (without type parameters):

1. bare \list\ → \IndexError\ at \rgs[0]\
2. bare \dict\ → \ValueError\ at \_, items_type = get_args(type_)\

## Fix

Both branches now guard against missing type arguments and return the value unchanged.

- \dict\ branch: guard with \if len(args_tuple) < 2: return value\
- \list\ branch: guard with \if not args: return value\

Fixes #1626.
Fixes #1619.

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -647,7 +647,10 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        args_tuple = get_args(type_)
+        if len(args_tuple) < 2:
+            return value
+        _, items_type = args_tuple  # Dict[_, items_type]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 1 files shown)


PHclaw:
Bumping this - `construct_type()` crashes on bare `list`/`dict` type annotations (no type parameters). This is a real crash, not a theoretical issue (see #1626, #1619). The fix adds 2 guard lines. Would love any feedback on what's needed to merge. Thanks!

craigie-ant:
👋 thanks for the PR! closing because of https://github.com/anthropics/anthropic-sdk-python/issues/1619#issuecomment-5257021164

PR #1651, opened 2026-06-04
fix(bedrock): preserve SSE event type in stream decoder instead of hardcoding "completion"

### LINKED ISSUE #1647: Bedrock stream decoder drops SSE event types; type-less chunks construct as RawMessageStartEvent(message=None)
### Summary

The Bedrock stream decoder (`anthropic/lib/bedrock/_stream_decoder.py`) hardcodes `ServerSentEvent(event="completion")` for every chunk, discarding the real event type. Downstream, the `"completion"` branch in `_streaming.py::__stream__` never backfills `data["type"]` (unlike the direct-SSE branch). Chunks without a `type` field — notably Bedrock's `amazon-bedrock-invocationMetrics` trailer — then hit non-validating `construct_type` against the stream-event union, which falls back to the **first union member** and yields `RawMessageStartEvent(message=None)` — an object that violates the SDK's own type contract (`message` is typed as non-optional).

### Reproduction (anthropic 0.105.2, offline)

```python

event = construct_type(
    value={"amazon-bedrock-invocationMetrics": {"inputTokenCount": 1}},
    type_=BetaRawMessageStreamEvent,
)
print(type(event).__name__, event.message)
# BetaRawMessageStartEvent None
```

Any consumer that trusts the type annotations (e.g. `event.message.usage`) crashes with `AttributeError: 'NoneType' object has no attribute 'usage'`. We hit this in production via pydantic-ai's streaming path on `AsyncAnthropicBedrock` — intermittent 500s whenever a stream contained the invocationMetrics trailer.

### Expected

Either:
- the Bedrock decoder preserves/backfills the event type from the chunk payload's `type` field (the payloads do carry one for real events), and drops/ignores non-Messages chunks like `amazon-bedrock-invocationMetrics`, or
- stream-event construction validates and raises a clear error instead of silently producing a contract-violating first-union-member object.

### Notes

- Related but distinct from #1477 (decoder raises `ValueError` on error SSE events) — same file, different gap: this one is about *successful* streams containing type-less chunks.
- Direct (non-Bedrock) SSE is unaffected: that branch carries the event type and backfills `data["type"]`.

MaitreyeeDeshmukh:
## Summary

`AWSEventStreamDecoder.iter_bytes()` and `aiter_bytes()` in `src/anthropic/lib/bedrock/_stream_decoder.py` hardcode `event="completion"` for every `ServerSentEvent` they yield, discarding the real event type from the decoded chunk.

This causes downstream issues in `_streaming.py`'s `__stream__` method, which relies on `sse.event` to backfill `data["type"]` when it is missing. Chunks without a `type` field (such as Bedrock's `amazon-bedrock-invocationMetrics` trailer) fall back to `construct_type` against the stream-event union, which picks the first union member — yielding `RawMessageStartEvent(message=None)`, an object that violates the SDK's own type contract.

## Fix

Parse the `"type"` field from the decoded JSON bytes and pass it as the `event` argument to `ServerSentEvent`. Falls back to `"completion"` if the bytes cannot be parsed as JSON, preserving existing behavior for malformed chunks.

Changed file: `src/anthropic/lib/bedrock/_stream_decoder.py` — 3 lines changed in each of `iter_bytes` and `aiter_bytes`.

## Testing

Added two regression tests to `tests/lib/test_bedrock.py`:
- `test_stream_decoder_preserves_sse_event_type` (sync)
- `test_stream_decoder_preserves_sse_event_type_async` (async)

Both assert that a chunk with `"type": "message_start"` yields a `ServerSentEvent` with `event == "message_start"`, not `"completion"`. All 14 tests pass across Python 3.9, 3.14 with both Pydantic v1 and v2.

Closes #1647

### DIFF
--- a/src/anthropic/lib/bedrock/_stream_decoder.py
+++ b/src/anthropic/lib/bedrock/_stream_decoder.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
 from typing import TYPE_CHECKING, Iterator, AsyncIterator
 
 from ..._utils import lru_cache
@@ -37,7 +38,11 @@ class AWSEventStreamDecoder:
             for event in event_stream_buffer:
                 message = self._parse_message_from_event(event)
                 if message:
-                    yield ServerSentEvent(data=message, event="completion")
+                    try:
+                        event_type = json.loads(message).get("type", "completion")
+                    except Exception:
+                        event_type = "completion"
+                    yield ServerSentEvent(data=message, event=event_type)
 
     async def aiter_bytes(self, iterator: AsyncIterator[bytes]) -> AsyncIterator[ServerSentEvent]:
         """Given an async iterator that yields lines, iterate over it & yield every event encountered"""
@@ -49,7 +54,11 @@ class AWSEventStreamDecoder:
             for event in event_stream_buffer:
                 message = self._parse_message_from_event(event)
                 if message:
-                    yield ServerSentEvent(data=message, event="completion")
+                    try:
+                        event_type = json.loads(message).get("type", "completion")
+                    except Exception:
+                        event_type = "completion"
+                    yield ServerSentEvent(data=message, event=event_type)
 
     def _parse_message_from_event(self, event: EventStreamMessage) -> str | None:
         response_dict = event.to_response_dict()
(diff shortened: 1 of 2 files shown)


tomer-ant:
Thanks, but this is fixed in v0.110.0 (https://github.com/anthropics/anthropic-sdk-python/commit/ed0af5bdd96ab0be9922edc401fb2eb83a54101b). The Bedrock stream decoder now forwards the payload's type as the SSE event type instead of hardcoding completion.

PR #1653, opened 2026-06-05
fix: correct content_block_delta deserialization in streaming

### LINKED ISSUE #941: content_block_delta event not deserialized correctly during streaming

Oxygen56:
Fixes content_block_delta event deserialization during streaming.

Fixes #941

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -601,6 +601,14 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
     else:
         meta = tuple()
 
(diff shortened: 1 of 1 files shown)


dtmeadows-ant:
@Oxygen56 thanks for this. The underlying issue (#941) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1672, opened 2026-06-13
fix(client): handle bare dict/list annotations in construct_type and transform

Zawwarsami16:
## What

A field annotated with a bare, unparameterized `dict` or `list` (rather than `Dict[K, V]` / `List[T]`) crashes the runtime helpers, because `get_args()` returns an empty tuple in that case and the code indexes into it unconditionally.

Three reported crashes, one root cause:

- `construct_type(value={...}, type_=dict)` → `ValueError: not enough values to unpack (expected 2, got 0)` (`_models.py`) — #1619
- `construct_type(value=[...], type_=list)` → `IndexError: tuple index out of range` (`_models.py`) — #1626
- `transform({...}, dict)` → `IndexError: tuple index out of range`, in both the sync and async paths (`_transform.py`) — #1628

## Fix

Guard each `get_args()` lookup and fall back to `object` as the element type when no type parameters are present. With `object` as the element type, values pass straight through unchanged — which is the right behavior for an unparameterized container, since there's no inner type to coerce to.

## Repro (before)

```python

construct_type(value={"k": "v"}, type_=dict)   # ValueError
construct_type(value=["a", "b"], type_=list)   # IndexError
transform({"k": "v"}, dict)                     # IndexError
```

All three (plus the async transform path) now return the input unchanged.

## Tests

- `tests/test_models.py`: `test_bare_dict_annotation`, `test_bare_list_annotation`
- `tests/test_transform.py`: `test_bare_dict_annotation` (parametrized sync + async)

Full `test_models.py` and `test_transform.py` suites pass locally (122 tests); ruff check + format are clean.

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -647,7 +647,8 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        args = get_args(type_)
+        items_type = args[1] if len(args) >= 2 else object  # Dict[_, items_type]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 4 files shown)


craigie-ant:

PR #1674, opened 2026-06-13
fix(agent_toolset): read images/PDFs as content blocks instead of crashing

### LINKED ISSUE #1637: Self-hosted agent_toolset `read` tool raises UnicodeDecodeError on binary files (images/PDFs) instead of returning content blocks
### Summary

The self-hosted `agent_toolset_20260401` `read` tool (`src/anthropic/lib/tools/agent_toolset.py`, `beta_read_tool`) decodes every file as UTF-8 via `target.read_text()`. Reading a **binary** file (image or PDF) raises an uncaught `UnicodeDecodeError`, surfaced to the model as a raw tool error — even though the tool-result contract already supports image/document content blocks.

So a Managed Agents agent running in a **self-hosted** environment (`client.beta.environments.work.worker(...).handle_item()` / `SessionToolRunner`) cannot `read` an image or PDF. This bites the document skills (docx/pdf/pptx/xlsx) that render slides/pages to images for visual QA — every such `read` fails. (The hosted product and Claude Code's `Read` both handle images, so this is specific to the open-source self-hosted toolset.)

### Version

`anthropic==0.103.1`. Still present on `main` / `v0.105.2` — `beta_read_tool` is `read_text()`-only there too. I couldn't find an existing issue for it.

### Repro

1. Self-hosted CMA agent with the `agent_toolset_20260401` toolset (a StandardSandbox / `agent_toolset_20260401` tool).
2. The agent creates or has an image, e.g. `/workspace/slide-1.jpg`.
3. The agent calls `read(file_path="/workspace/slide-1.jpg")`.
4. The tool raises (surfaced as the tool result):

   ```
   UnicodeDecodeError('utf-8', b'\xff\xd8\xff\xe0...', 0, 1, 'invalid start byte')
   ```

   (`\xff\xd8` is the JPEG SOI marker.) The same happens for PDFs (`%PDF`).

### Root cause

`beta_read_tool`'s inner `read`:

```python
text = target.read_text()   # UTF-8; raises UnicodeDecodeError on binary
```

Only `ToolError` and `OSError` are caught; `UnicodeDecodeError` (a `ValueError`) propagates uncaught.

### Why this is straightforward to fix

The tool-result type already supports content blocks, and the runner forwards them:

- `BetaFunctionToolResultType = Union[str, Iterable[BetaContent]]` (`_beta_functions.py`)
- `ToolError`'s own docstring shows an `image` block example
- `_beta_session_runner._to_session_content` already forwards `image` / `document` / `search_result` blocks through to the session

So `read` can simply return an `image`/`document` block for binary files.

### Suggested fix

Detect image/PDF files (by extension and/or magic bytes) and return a base64 content block instead of decoding as text:

```python
# image/jpeg|png|gif|webp -> "image"; application/pdf -> "document"
media = _binary_media_type(target)
if media is not None:
    data = base64.standard_b64encode(target.read_bytes()).decode("ascii")
    kind = "document" if media == "application/pdf" else "image"
    return [{"type": kind, "source": {"type": "base64", "media_type": media, "data": data}}]
text = target.read_text()
...
```

A separate (larger) size cap for binary makes sense, since images routinely exceed the 256 KB `READ_MAX_BYTES` text cap. Happy to open a PR if useful.

### Interim workaround

For anyone else self-hosting CMA: we monkeypatch `agent_toolset.beta_read_tool` in our worker before `handle_item()` to do exactly the above (binary → base64 `image`/`document` block, text → delegate to the original `read`). Restores visual QA of rendered output for the document skills.

Zawwarsami16:
Closes #1637.

### Problem

The self-hosted `agent_toolset_20260401` `read` tool decodes every file as UTF-8 (`target.read_text()`). Reading a binary **image** or **PDF** raises an uncaught `UnicodeDecodeError` (a `ValueError`, not caught alongside `ToolError`/`OSError`), which surfaces to the model as a raw tool error:

```
UnicodeDecodeError('utf-8', b'\xff\xd8\xff\xe0...', 0, 1, 'invalid start byte')
```

So an agent running under `SessionToolRunner` / the self-hosted environment worker can't `read` an image or PDF — which breaks the document skills that render pages/slides to images for visual QA. The hosted product and Claude Code's `Read` both handle images, so this gap is specific to the open-source self-hosted toolset.

### Fix

`read` now sniffs the leading magic bytes and, for a recognized image/PDF, returns the matching base64 `image` / `document` content block instead of decoding as text. The plumbing for this already exists:

- `BetaFunctionToolResultType = Union[str, Iterable[BetaContent]]`, and `BetaContent` already includes `BetaImageBlockParam` / `BetaRequestDocumentBlockParam`.
- `_beta_session_runner._to_session_content` already forwards `image` / `document` / `search_result` blocks to the session.

Detected types: JPEG, PNG, GIF, WebP, and PDF (content-sniffed, not extension-trusted). `view_range` isn't meaningful for binary, so it's rejected with a clear message. Any other non-UTF-8 file now raises a clean `ToolError` ("not a UTF-8 text file and not a supported binary…") rather than letting the `UnicodeDecodeError` propagate uncaught.

The text path is unchanged except that decoding is now explicitly UTF-8 (was already UTF-8 via `read_text()`).

### Tests

- Parametrized read of JPEG/PNG/GIF/WebP/PDF returns the right block type + media type, with base64 that round-trips to the original bytes.
- `view_range` on a binary file raises.
- Arbitrary non-text/non-image bytes raise a clear `ToolError` instead of `UnicodeDecodeError`.

`pytest tests/lib/tools/test_agent_toolset.py tests/lib/tools/test_session_runner.py` → 80 passed. `ruff check` + `ruff format --check` clean; pyright reports no errors on the changed file.

### DIFF
--- a/src/anthropic/lib/tools/agent_toolset.py
+++ b/src/anthropic/lib/tools/agent_toolset.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 import os
 import re
 import uuid
+import base64
 import shutil
 import logging
 import subprocess
@@ -47,6 +48,7 @@ from itertools import islice
 from contextlib import asynccontextmanager
 from dataclasses import field, dataclass
 from collections.abc import Mapping, Callable, Awaitable, AsyncIterator
+from typing_extensions import Literal
 
 import anyio
 import anyio.abc
@@ -56,6 +58,8 @@ from ._skills import _within, download_session_skills
 from ..._types import NotGiven, not_given
 from ..._utils import is_given
 from ...types.beta import (
+    BetaImageBlockParam,
+    BetaRequestDocumentBlockParam,
     BetaManagedAgentsAgentToolset20260401BashInput,
     BetaManagedAgentsAgentToolset20260401EditInput,
     BetaManagedAgentsAgentToolset20260401GlobInput,
@@ -63,7 +67,8 @@ from ...types.beta import (
     BetaManagedAgentsAgentToolset20260401ReadInput,
     BetaManagedAgentsAgentToolset20260401WriteInput,
 )
-from ._beta_functions import ToolError, BetaAsyncFunctionTool, beta_async_tool
+from ._beta_functions import ToolError, BetaAsyncFunctionTool, BetaFunctionToolResultType, beta_async_tool
+from ...types.beta.beta_tool_result_block_param import Content as BetaContent
 
 if TYPE_CHECKING:
     from ..._client import AsyncAnthropic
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
@Zawwarsami16 thanks for this. The underlying issue (#1637) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1675, opened 2026-06-14
Fix: add missing .stream() method to Bedrock beta.messages

### LINKED ISSUE #1210: Bedrock beta.messages missing .stream() method
## Summary

`AnthropicBedrock().beta.messages` exposes `.create()` but not `.stream()`. The first-party `Anthropic().beta.messages` has both.

## Root cause

In `src/anthropic/lib/bedrock/_beta_messages.py`, the Bedrock `Messages` class only copies `create`:

```python
class Messages(SyncAPIResource):
    create = FirstPartyMessagesAPI.create
    # stream is missing
```

The first-party class in `resources/beta/messages/messages.py` has both `create` and `stream`.

## Impact

Any code using `client.beta.messages.stream(...)` breaks when switching from `Anthropic()` to `AnthropicBedrock()`. Streaming itself works fine on Bedrock — `create(stream=True)` returns events correctly. It's just the convenience `.stream()` method that's missing from the class definition.

## Expected fix

```python
class Messages(SyncAPIResource):
    create = FirstPartyMessagesAPI.create
    stream = FirstPartyMessagesAPI.stream
```

(Same for `AsyncMessages`.)

Since this file is Stainless-generated, the fix likely needs to happen in the generator config.

## SDK version

`anthropic==0.80.0`

ushaganeshdurai:
## Summary
Adds the missing `.stream()` method to `AnthropicBedrock().beta.messages` for parity with `Anthropic().beta.messages`.

## Changes
- Added `stream` alias to `Messages` class in `src/anthropic/lib/bedrock/_beta_messages.py`
- Added `stream` alias to `AsyncMessages` class for async support

## Fixes
Closes #1210

## Testing
Verified that Bedrock's `create(stream=True)` returns the same event types as the first-party `stream()` method.

### DIFF
--- a/src/anthropic/lib/bedrock/_beta_messages.py
+++ b/src/anthropic/lib/bedrock/_beta_messages.py
@@ -13,6 +13,7 @@ __all__ = ["Messages", "AsyncMessages"]
 
 class Messages(SyncAPIResource):
     create = FirstPartyMessagesAPI.create
+    stream = FirstPartyMessagesAPI.stream
 
     @cached_property
     def with_raw_response(self) -> MessagesWithRawResponse:
@@ -36,6 +37,7 @@ class Messages(SyncAPIResource):
 
 class AsyncMessages(AsyncAPIResource):
     create = FirstPartyAsyncMessagesAPI.create
+    stream = FirstPartyAsyncMessagesAPI.stream
 
     @cached_property
     def with_raw_response(self) -> AsyncMessagesWithRawResponse:


tomer-ant:
Thanks, but this is fixed in v0.122.0 (https://github.com/anthropics/anthropic-sdk-python/commit/6eca7bb19f968856b5652d5471e3ca9fc3fe8641). AnthropicBedrock().beta.messages now exposes stream, parse and tool_runner.

PR #1682, opened 2026-06-16
fix(bedrock): preserve stream event type and drop invocation-metrics trailer

### LINKED ISSUE #1647: Bedrock stream decoder drops SSE event types; type-less chunks construct as RawMessageStartEvent(message=None)

qing-ant:
## Problem

The Bedrock stream decoder wraps every chunk as `ServerSentEvent(event="completion")`, discarding the payload's actual event type. Downstream, the `"completion"` branch in `_streaming.py::__stream__` does not backfill `data["type"]` (unlike the typed-event branch), so chunks without a `type` field reach `construct_type` against the `RawMessageStreamEvent` union with no discriminator. That falls back to the first union member and yields `RawMessageStartEvent(message=None)` — an object that violates the SDK's own type contract.

The `amazon-bedrock-invocationMetrics` trailer that Bedrock appends to streams has no `type` field, so it hits exactly this. Any consumer reading `event.message.usage` crashes with `AttributeError: 'NoneType' object has no attribute 'usage'`.

Fixes #1647.

## Fix

In `lib/bedrock/_stream_decoder.py` (hand-maintained, not Stainless-generated):

- Parse each chunk's JSON and use its `type` field as the SSE event name, so typed Messages events route through the branch in `__stream__` that backfills `data["type"]`.
- Drop the metrics-only trailer (no `type`, no `completion` key) instead of forwarding it to the stream-event union.
- Fall back to `event="completion"` for legacy untyped payloads (and any non-JSON payload), preserving the existing Completions path. A legacy completion chunk that also carries `amazon-bedrock-invocationMetrics` is kept, not dropped.

## Relation to #1651

#1651 preserves the event type but still forwards the metrics trailer (its `.get("type", "completion")` returns `"completion"` for the type-less trailer), so the reported crash still occurs. This PR adds the missing drop and a regression test for it.

## Behavior notes

- Bedrock streams now route by the same event names as direct SSE. A chunk whose `type` is not in `__stream__`'s recognized list is dropped rather than yielded as a mis-constructed object — this matches non-Bedrock behavior.
- A `{"type": "error", ...}` chunk inside a 200 stream now raises (same as direct SSE) instead of being yielded as data.

## Tests

Four unit tests on `_chunk_bytes_to_sse`: typed event preserved, metrics-only trailer dropped, legacy completion kept, legacy completion with merged metrics kept.

### DIFF
--- a/src/anthropic/lib/bedrock/_stream_decoder.py
+++ b/src/anthropic/lib/bedrock/_stream_decoder.py
@@ -79,11 +79,6 @@ def _chunk_bytes_to_sse(raw: bytes) -> ServerSentEvent | None:
     payload = cast("Dict[str, Any]", data)
     event_type = payload.get("type")
     if not isinstance(event_type, str):
-        # Bedrock appends a trailing chunk that only carries invocation
-        # metrics; it is not a model stream event, so drop it rather than
-        # let it be mis-constructed against the stream-event union.
-        if "amazon-bedrock-invocationMetrics" in payload and "completion" not in payload:
-            return None
         event_type = "completion"
 
     return ServerSentEvent(data=decoded, event=event_type)
--- a/src/anthropic/lib/bedrock/_stream_decoder.py
+++ b/src/anthropic/lib/bedrock/_stream_decoder.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Iterator, AsyncIterator
+import json
+from typing import TYPE_CHECKING, Any, Dict, Iterator, AsyncIterator, cast
 
 from ..._utils import lru_cache
 from ..._streaming import ServerSentEvent
@@ -35,9 +36,9 @@ class AWSEventStreamDecoder:
         for chunk in iterator:
             event_stream_buffer.add_data(chunk)
             for event in event_stream_buffer:
-                message = self._parse_message_from_event(event)
-                if message:
-                    yield ServerSentEvent(data=message, event="completion")
+                sse = self._parse_message_from_event(event)
+                if sse:
+                    yield sse
 
     async def aiter_bytes(self, iterator: AsyncIterator[bytes]) -> AsyncIterator[ServerSentEvent]:
         """Given an async iterator that yields lines, iterate over it & yield every event encountered"""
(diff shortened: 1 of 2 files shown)


src/anthropic/lib/bedrock/_stream_decoder.py
```diff
@@ -61,4 +62,28 @@ def _parse_message_from_event(self, event: EventStreamMessage) -> str | None:
...
+        return ServerSentEvent(data=decoded, event="completion")
+
+    payload = cast("Dict[str, Any]", data)
+    event_type = payload.get("type")
+    if not isinstance(event_type, str):
+        # Bedrock appends a trailing chunk that only carries invocation
+        # metrics; it is not a model stream event, so drop it rather than
+        # let it be mis-constructed against the stream-event union.
+        if "amazon-bedrock-invocationMetrics" in payload and "completion" not in payload:
+            return None
```
craigie-ant, line 86: this is something we should fix but it is a breaking change, as a user may be using this value, so I'm removing the logic here for now

craigie-ant (approved):
Thanks!

sumanth9977 (approved):

PR #1690, opened 2026-06-19
fix(streaming): validate content_block_start index before appending

WalkingDreams798:
## What

`accumulate_event` (in both `lib/streaming/_messages.py` and `_beta_messages.py`) appended each `content_block_start` block to the snapshot **without checking `event.index`** — it silently relied on blocks arriving strictly in order. This was the long-standing `# TODO: check index` in both files.

If a `content_block_start` ever arrived with an out-of-order or skipped index, the block would land at the wrong position in `current_snapshot.content`, and subsequent `content_block_delta` / `content_block_stop` events (which index by `event.index`) would mutate the wrong block or raise an opaque `IndexError`.

## Change

Validate that `event.index == len(current_snapshot.content)` before appending, and raise a clear `RuntimeError` on mismatch — consistent with the existing `'Unexpected event order, got ... before "message_start"'` style already used in the same function.

Applied identically to the standard and beta accumulators, resolving both TODOs.

## Tests

Added `tests/lib/streaming/test_content_block_index.py` covering, for both the standard and beta paths:
- sequential indices append correctly
- a mismatched index raises `RuntimeError`

## Verification

- `uv run pytest tests/lib/streaming` → 44 passed
- `uv run ruff check` → all checks passed
- `uv run pyright` (strict) on changed files → 0 errors, 0 warnings

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -475,7 +475,12 @@ def accumulate_event(
         raise RuntimeError(f'Unexpected event order, got {event.type} before "message_start"')
 
     if event.type == "content_block_start":
-        # TODO: check index
+        if event.index != len(current_snapshot.content):
+            raise RuntimeError(
+                f'Unexpected "content_block_start" index, got {event.index} '
+                f"but expected {len(current_snapshot.content)}"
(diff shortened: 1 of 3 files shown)


tomer-ant:

PR #1707, opened 2026-06-25
Fix MCP async tool_runner docs

fallintoplace:
## Problem
Some MCP async examples show `await` on `client.beta.messages.tool_runner(...)`, but `tool_runner()` returns an async iterator and is not awaitable.

Users who copy that pattern hit a runtime `TypeError` before entering the runner loop.

## Fix
- Remove the incorrect `await` in the MCP helpers guide example (`helpers.md`).
- Remove the incorrect `await` from the `async_mcp_tool()` docstring example (`src/anthropic/lib/tools/mcp.py`).

Both examples now directly assign `runner = client.beta.messages.tool_runner(...)` and then iterate with `async for`.

## Testing
Not run (docs-only change).

### DIFF
--- a/helpers.md
+++ b/helpers.md
@@ -162,7 +162,7 @@ async with stdio_client(StdioServerParameters(command="mcp-server")) as (read, w
         await mcp_client.initialize()
 
         tools_result = await mcp_client.list_tools()
-        runner = await client.beta.messages.tool_runner(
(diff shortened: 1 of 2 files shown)


dtmeadows-ant:
Thanks @fallintoplace. Could you handle the conflicts and then I'll approve running CI here?

PR #1727, opened 2026-07-02
fix: handle binary files in agent_toolset read tool

### LINKED ISSUE #1637: Self-hosted agent_toolset `read` tool raises UnicodeDecodeError on binary files (images/PDFs) instead of returning content blocks

okxint:
## Summary

`beta_read_tool` calls `target.read_text()` unconditionally. Binary files (JPEG, PNG, GIF, WebP, PDF) raise `UnicodeDecodeError` at that call, which surfaces to the model as a raw tool error — even though the tool-result type already supports `image` and `document` content blocks.

This bites any self-hosted CMA agent that needs to read an image or PDF: rendered slides, screenshots, attached documents. The hosted product and Claude Code both handle binary reads; this brings the self-hosted toolset in line.

## Fix

Detect binary files by extension before attempting `read_text()`. For known image and PDF suffixes, read the raw bytes and return a base64 content block:

- `.jpg`/`.jpeg`/`.png`/`.gif`/`.webp` → `{"type": "image", "source": {"type": "base64", "media_type": "...", "data": "..."}}`
- `.pdf` → `{"type": "document", "source": {"type": "base64", "media_type": "application/pdf", "data": "..."}}`

Binary files get their own size cap (`DEFAULT_MAX_BINARY_BYTES = 5 MiB`, separate from the 256 KiB text cap) since images regularly exceed the text limit. `view_range` is rejected for binary files with an explicit error.

Text files and unknown extensions go through the existing `read_text()` path unchanged.

## Tests

- Parametrized across all 6 supported extensions/formats — verifies the correct `type`, `media_type`, and round-tripped base64 payload
- `view_range` rejection for binary files
- Oversized binary file rejection

Fixes #1637

### DIFF
--- a/src/anthropic/_client.py
+++ b/src/anthropic/_client.py
@@ -211,8 +211,8 @@ class Anthropic(SyncAPIClient):
             or profile is not None
         )
         if not has_explicit_credential:
-            api_key = os.environ.get("ANTHROPIC_API_KEY")
-            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN")
+            api_key = os.environ.get("ANTHROPIC_API_KEY") or None
+            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") or None
         self.api_key = api_key
         self.auth_token = auth_token
         # --- end credentials support ---
@@ -628,8 +628,8 @@ class AsyncAnthropic(AsyncAPIClient):
             or profile is not None
         )
         if not has_explicit_credential:
-            api_key = os.environ.get("ANTHROPIC_API_KEY")
-            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN")
+            api_key = os.environ.get("ANTHROPIC_API_KEY") or None
+            auth_token = os.environ.get("ANTHROPIC_AUTH_TOKEN") or None
         self.api_key = api_key
         self.auth_token = auth_token
         # --- end credentials support ---
--- a/src/anthropic/lib/tools/agent_toolset.py
+++ b/src/anthropic/lib/tools/agent_toolset.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 import os
 import re
 import uuid
+import base64
 import shutil
 import logging
 import subprocess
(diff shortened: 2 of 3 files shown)


dtmeadows-ant:
@okxint thanks for this. The underlying issue (#1637) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1731, opened 2026-07-03
fix(client): handle bare dict/list annotations in construct_type and transform

### LINKED ISSUE #1619: BUG: construct_type() crashes with ValueError when type_ is a bare dict

kazuki-netizen:
Fixes #1619, #1626, #1628.

## Problem

`construct_type()` and `_transform_recursive` / `_async_transform_recursive` index the result of `get_args()` unconditionally. For an unparameterized annotation, `get_args(dict)` and `get_args(list)` return `()`, so the index access fails whenever a field is annotated as a bare `dict` or `list` instead of `Dict[K, V]` / `List[T]`:

- `construct_type(value={"k": "v"}, type_=dict)` → `ValueError: not enough values to unpack (expected 2, got 0)` (`_models.py`)
- `construct_type(value=["a", "b"], type_=list)` → `IndexError: tuple index out of range` (`_models.py`)
- `transform({"k": "v"}, dict)` / `async_transform({"k": "v"}, dict)` → `IndexError: tuple index out of range` (`_utils/_transform.py`)

## Fix

Guard each access and fall back to `object` as the item type when no type parameter is present. This matches the semantics of `Dict[str, object]` / `List[object]`: items are returned as-is with no further coercion. Parameterized annotations and the non-mapping / non-list passthrough branches are unchanged.

## Verification

- Added regression tests in `tests/test_models.py` and `tests/test_transform.py` (sync + async). They fail on `main` and pass with this change.
- `tests/test_models.py` + `tests/test_transform.py`: 122 passed.
- `ruff check` and `ruff format --check` clean.

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -647,7 +647,7 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_mapping(value):
             return value
 
-        _, items_type = get_args(type_)  # Dict[_, items_type]
+        items_type = args[1] if len(args) >= 2 else object  # Dict[_, items_type]
         return {key: construct_type(value=item, type_=items_type) for key, item in value.items()}
 
     if (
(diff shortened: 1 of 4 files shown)


craigie-ant:

PR #1740, opened 2026-07-06
fix(streaming): emit input_json events for server_tool_use blocks

sean-kim05:
## What

When streaming a message that contains a **server-side tool call** (`server_tool_use`, e.g. `web_search`), the SDK accumulates the tool's streamed input into the message snapshot but never emits an `input_json` stream event for it. `on_input_json` callbacks and code iterating the event stream therefore never see the input of server tools — only client `tool_use` (and, in the beta path, `mcp_tool_use`) fire `input_json`.

## Root cause

`accumulate_event` decides *what to accumulate* with `TRACKS_TOOL_INPUT`:

```python
# src/anthropic/lib/streaming/_messages.py
TRACKS_TOOL_INPUT = (ToolUseBlock, ServerToolUseBlock)
...
if isinstance(content, TRACKS_TOOL_INPUT):
    ...  # build up content.input from input_json_delta
```

But `build_events` decides *what to emit* with a narrower, hard-coded check:

```python
elif event.delta.type == "input_json_delta":
    if content_block.type == "tool_use":          # non-beta: server_tool_use missing
        events_to_fire.append(build(InputJsonEvent, ...))
```

The beta path (`_beta_messages.py`) tracks `(BetaToolUseBlock, BetaServerToolUseBlock, BetaMCPToolUseBlock)` but only emits for `tool_use` / `mcp_tool_use` — again dropping `server_tool_use`. So the emitter and the accumulator disagree about which blocks carry input.

## Fix

Emit `input_json` for exactly the block types whose input is accumulated, by reusing `TRACKS_TOOL_INPUT` via `isinstance` in `build_events` (both non-beta and beta). Emission and accumulation now share one source of truth and can't drift apart.

## Tests

- New fixture `tests/lib/streaming/fixtures/server_tool_use_response.txt` — a stream with a `web_search` `server_tool_use` block whose input streams in as `{"query": "weather in Paris"}`.
- `test_server_tool_use_input_json` (sync + async) in both `test_messages.py` and `test_beta_messages.py`, asserting `input_json` events fire and the final snapshot equals the accumulated input.
- Verified RED→GREEN: without the fix, `input_json_events == []` for the `server_tool_use` block in all four tests; with it, they fire.
- Full `tests/lib/streaming/` suite (44 tests) passes; `./scripts/lint` (ruff + pyright + mypy) clean.

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -365,7 +365,11 @@ def build_events(
                     )
                 )
         elif event.delta.type == "input_json_delta":
-            if content_block.type == "tool_use" or content_block.type == "mcp_tool_use":
+            # Emit `input_json` for every block whose input we accumulate below
+            # (see `accumulate_event`/`TRACKS_TOOL_INPUT`), not just `tool_use`
+            # and `mcp_tool_use` — otherwise `server_tool_use` blocks stream
+            # their input into the snapshot but never fire an `input_json` event.
+            if isinstance(content_block, TRACKS_TOOL_INPUT):
                 events_to_fire.append(
                     build(
                         BetaInputJsonEvent,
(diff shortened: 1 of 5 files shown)


tomer-ant:

PR #1745, opened 2026-07-07
fix(tools): surface non-UTF-8 files as ToolError in read/edit

sean-kim05:
## Summary

The agent `read` and `edit` tools (`src/anthropic/lib/tools/agent_toolset.py`) decode files with `Path.read_text()` (UTF-8) inside a `try` that only catches `ToolError` and `OSError`:

```python
text = target.read_text()
except ToolError:
    raise
except OSError as e:
    raise _fs_error("read", file_path, e) from e
```

`UnicodeDecodeError` is a subclass of `ValueError`, **not** `OSError`, so it escapes uncaught. The moment the model calls `read`/`edit` on a file that isn't valid UTF-8 — a binary artifact, or an ordinary latin-1 / UTF-16 text file — the exception bypasses the tools' error-normalization: instead of the intended clean `"read: <path>: ..."` `ToolError`, the messages runner logs a full traceback (`log.exception`) and hands the model a raw `UnicodeDecodeError` repr.

The sibling `grep` and `bash` tools in the same module already decode with `errors="replace"`, so non-UTF-8 files are a known, expected input here — `read`/`edit` just don't handle them.

### Reproduction

```python
env = AgentToolContext(workdir=d)
(Path(d) / "notes.txt").write_bytes(b"caf\xe9 menu\n")   # latin-1 text
await beta_read_tool(env).call({"file_path": "notes.txt"})
# -> UnicodeDecodeError leaks out instead of a ToolError
```

## Fix

Catch `UnicodeDecodeError` in both tools and raise a clean `ToolError`. The `edit` path deliberately does **not** fall back to a lossy `errors="replace"` decode, since the replacement round-trip would corrupt the file on write-back.

## Testing

Added `test_read_rejects_non_utf8_file` and `test_edit_rejects_non_utf8_file` in `tests/lib/tools/test_agent_toolset.py` (a latin-1 file must raise `ToolError` matching "not valid UTF-8"). Verified RED (raw `UnicodeDecodeError`) → GREEN. Full `test_agent_toolset.py` passes (42 tests); `./scripts/lint` clean (ruff + pyright + mypy).

### DIFF
--- a/src/anthropic/lib/tools/agent_toolset.py
+++ b/src/anthropic/lib/tools/agent_toolset.py
@@ -512,6 +512,11 @@ def beta_read_tool(ctx: AgentToolContext) -> BetaAsyncFunctionTool[Any]:
             text = target.read_text()
         except ToolError:
             raise
+        except UnicodeDecodeError as e:
+            # read_text() decodes as UTF-8; a non-UTF-8 file (binary artifact or
+            # e.g. a latin-1 text file) would otherwise escape as a raw
+            # UnicodeDecodeError, bypassing this tool's ToolError contract.
+            raise ToolError(f"read: {file_path}: not valid UTF-8 text") from e
         except OSError as e:
             raise _fs_error("read", file_path, e) from e
         if not view_range:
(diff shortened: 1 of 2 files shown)


tomer-ant:
Thanks, but this is fixed in v0.119.0 (https://github.com/anthropics/anthropic-sdk-python/commit/d2bad4fc771493d245a7fb182ef6860907ab728f). read and edit now raise a ToolError on non-UTF-8 text files and return images/PDFs as content blocks.

Issue #1744, opened 2026-07-07
Managed Agents: intermittent 400 "does not match any agent.tool_use event" confirming a tool_use on a non-primary session_thread_id


nadavshohat:
### Summary

Using a self-hosted `EnvironmentWorker` with multiagent enabled, confirming a `tool_result` for a tool that ran inside a **non-primary `session_thread_id`** (a sub-thread) intermittently fails with a `400`:

```
Error code: 400 - invalid_request_error:
`tool_use_id` "sevt_..." does not match any `agent.tool_use` event in this session
```

The tool itself executes successfully; only the result confirmation is rejected. Because `_send_result` treats the `400` as fatal (no re-fetch, no retry after `SEND_RETRIES`), a transient failure becomes a permanent strand: the session sits in `requires_action` until an external `user.interrupt`.

### Environment

- `anthropic==0.112.0`
- Self-hosted `EnvironmentWorker(...).handle_item(...)`
- Multiagent enabled via `beta_agent_toolset_20260401` (orchestrator spawns sub-threads)
- The worker only supplies tools; all tool-result posting is done inside the SDK.

### What I observe (from session event traces)

- The failing `tool_use_id`s are `agent.tool_use` events whose `session_thread_id` is a sub-thread (`sthr_...`), not the primary session.
- The failing tools are ordinary toolset tools (`bash`, `read`) with `evaluated_permission: "allow"` — not `always_ask`.
- It is **intermittent**: the same tool, in the same sub-thread, with the same permission, sometimes succeeds. It frequently affects the **second of two near-simultaneous `agent.tool_use` blocks** issued in one sub-thread turn (the first is confirmed, the second `400`s).
- It **self-heals ~11 minutes later** on a re-drive: a `user.tool_result` for the same `tool_use_id` eventually appears, carrying the correct `session_thread_id`. So the SDK does route to the sub-thread correctly when it succeeds.
- The **final** pending `agent.tool_use` in a chain does not self-heal; the session stays in `requires_action` until interrupted.
- Custom-tool results (`agent.custom_tool_use` / `user.custom_tool_result`) are unaffected in the same sessions.

Example trace shape (one session): 6 `agent.tool_use` but only 4 `user.tool_result`; all shortfall ids are sub-thread `agent.tool_use`.

### Questions

[...]

PR #1750, opened 2026-07-11
fix(runner): never re-execute an already-run tool call in SessionToolRunner

### LINKED ISSUE #1749: SessionToolRunner can re-execute a tool call after a stream reconnect if its result post failed
### Summary

`SessionToolRunner` (the implementation behind `client.beta.sessions.events.tool_runner()`) can execute a tool call **twice** for the same `tool_use_id` when a stream reconnect happens after a result-post attempt has failed permanently (or exhausted its retries). For a side-effecting tool — `bash`, a file write, anything in `agent_toolset` — this is an at-least-once instead of at-most-once execution hazard: the same shell command can run a second time.

### Root cause

`_reconcile()` (called on every `_stream_loop` reconnect, including after any transient disconnect) re-lists the session's history and re-enqueues every `agent.tool_use` / `agent.custom_tool_use` event whose id is not yet in `self._answered`:

```python
unanswered = [ev for ev in pending if ev.id not in self._answered]
...
for ev in unanswered:
    await self._send_work.send(ev)
```

`self._answered` is only populated when `_send_result` successfully posts the result. A call whose post permanently failed (e.g. a 400) or exhausted `SEND_RETRIES` is therefore indistinguishable, to `_reconcile`, from a call that was never dispatched at all — it gets sent back through `_send_work`, and `_dispatch_loop` hands it to `_execute`, which unconditionally re-runs the tool:

```python
async def _execute(self, ev: DispatchedToolUseEvent) -> None:
    ...
    else:
        content = await run_runnable_tool(tool, input_)   # <-- runs again
        ...
```

Any stream disconnect (network blip, load balancer timeout, a redeploy) that happens to land after a tool ran but before its result was successfully posted will trigger this on the next `_reconcile()` pass.

### Reproduction

Deterministic, no live API needed — using the existing `FakeAsyncEvents` test harness in `tests/lib/tools/test_session_runner.py`:

```python
@pytest.mark.asyncio()
async def test_reconnect_does_not_re_execute_tool_after_failed_send(monkeypatch):
    monkeypatch.setattr(session_runner_mod, "STREAM_BACKOFF_START", 0.001)
    counter = {"calls": 0}

    async def increment(_input):
        counter["calls"] += 1
        return "done"

    tool = _FakeTool("inc", increment)
    events = FakeAsyncEvents(
        list_events=[_tool_use("tu_1", "inc", {})],
        streams=[
            _FakeStream([_StubEvent("noop")], raise_after=1, raise_with=_api_status_error(500)),
            _FakeStream([_terminated()]),
        ],
        send_failures=[_api_status_error(400)],
    )

    [_ async for _ in _run_with_fakes(events=events, tools=[tool])]

    assert counter["calls"] == 1  # currently fails: counter["calls"] == 2
```

On current `main` (`d2f6543`) this fails with `counter["calls"] == 2` — the `inc` tool runs twice for `tu_1`.

### Expected behavior

Once a tool has run for a given `tool_use_id`, the runner should never run it again — a reconcile-triggered re-dispatch of an id whose tool already executed should only retry *posting* the already-computed result, not recompute it.

### Impact

Any self-hosted `EnvironmentWorker` / `SessionToolRunner` deployment using tools with side effects (the bundled `agent_toolset` bash/file tools, or a custom tool) is exposed to this on ordinary network instability, not just an unusual edge case — no multiagent/subagent setup or MCP involvement required to trigger it.

I have a fix + regression test ready and will open a PR shortly.

PranavMishra28:
## What was broken

`SessionToolRunner` (the implementation behind `client.beta.sessions.events.tool_runner()`) can execute a tool call **twice** for the same `tool_use_id` when a stream reconnect happens after a result-post attempt has permanently failed (or exhausted `SEND_RETRIES`). For a side-effecting tool — `bash`, a file write, anything registered on `agent_toolset` — this is an at-least-once instead of at-most-once execution hazard.

## Root cause

`_reconcile()` runs on every stream reconnect (including after any ordinary transient disconnect — network blip, load balancer timeout, a redeploy) and re-enqueues every `agent.tool_use` / `agent.custom_tool_use` whose id is not yet in `self._answered`. `self._answered` is only populated on a *successful* post, so a call whose post failed is indistinguishable, to `_reconcile`, from a call never dispatched at all — it goes back through `_send_work`, and `_dispatch_loop` hands it to `_execute`, which unconditionally re-runs the tool itself, not just the post.

## The fix

Adds `self._executed: dict[str, tuple[DispatchedToolResultParams, bool]]`, populated with a call's computed result as soon as its tool has run (regardless of whether the subsequent send succeeds), and cleared once the send is confirmed. `_dispatch_loop` checks it before calling `_execute`: an id already in `_executed` is routed to a new `_resend` (retries posting the cached result only) instead of `_execute` (which would re-run the tool). This preserves the runner's existing self-healing intent — a failed post keeps getting retried on the next reconcile, per the pre-existing code comment — while removing the unsafe re-execution. `_execute`'s and `_resend`'s shared tail (yielding the `DispatchedToolCall` to the consumer) is factored into `_post_result`.

Also updated `DispatchedToolCall.posted`'s docstring to describe the new retry-until-success behavior (the consumer may now see a second yield for the same `tool_use_id` with `posted=True` after an initial `posted=False`).

## How it's tested

Reproduced deterministically with the existing `FakeAsyncEvents` fake-event harness in `tests/lib/tools/test_session_runner.py` — no live API needed:

- `test_reconnect_does_not_re_execute_tool_after_failed_send` (new): a permanent 4xx on the first send attempt, followed by a stream reconnect (transient disconnect + reconnect), previously ran the tool twice for the same `tool_use_id`; confirmed failing on `main` before the fix (`counter["calls"] == 2`). With the fix: the tool runs exactly once, the first yield reports `posted=False`, and the reconcile-triggered retry succeeds and yields a second `DispatchedToolCall` with `posted=True` and the *same*, not recomputed, result content.

Full verification run for this PR:
- `uv run pytest tests/lib/tools/` — 188 passed, 1 skipped, 1 xfailed (no regressions; includes the pre-existing `test_skips_already_answered_events`, `test_reconcile_list_error_does_not_dispatch_partial`, `test_yields_with_posted_false_on_permanent_4xx`, and the concurrency/idle-clock tests, all still green)
- `uv run ruff check .` / `uv run ruff format --check .` on the changed files — clean
- `uv run mypy .` — clean (1052 source files)
- `uv run pyright` — 0 errors, 0 warnings

(`tests/api_resources/` requires the local Prism mock server via `./scripts/test` and 2 unrelated failures in `tests/lib/test_credentials.py` reproduce identically on clean `main` before this change — neither is touched by this diff.)

## What this deliberately does not change

- Does not change the outer retry/backoff policy for the post itself (`SEND_RETRIES`, `is_fatal_status_error` in `lib/_retry.py`) — only which code path a reconcile-triggered redispatch takes.
- Does not address #1744 (an unrelated, separate report about `session_thread_id` routing for multiagent sub-thread confirmations) — I looked into that issue first but could not confirm the suspected fix without either live API access or a maintainer confirming server-side behavior for an undocumented field; flagging it here in case it's useful context, but this PR does not touch that code path.

Fixes #1749

Developed with Claude Code; reviewed and tested by Pranav before marking ready for review.

### DIFF
--- a/src/anthropic/lib/tools/_beta_session_runner.py
+++ b/src/anthropic/lib/tools/_beta_session_runner.py
@@ -487,6 +487,12 @@ class SessionToolRunner:
         # by :meth:`_note_confirmation` / the next reconcile pass. Like ``_seen``
         # and ``_answered``, ``_confirmations`` is per-session O(tool calls):
         # recorded verdicts persist for the life of the run.
+        # ``_executed`` holds the computed (unconfirmed) result for an id whose
+        # tool has already run: reconcile re-enqueues anything not yet in
+        # ``_answered``, and a call already in ``_executed`` must only have its
+        # result re-posted, never re-run the tool itself (see ``_dispatch_loop``
+        # / ``_resend``).
+        self._executed: dict[str, tuple[DispatchedToolResultParams, bool, Literal["allow"] | None]] = {}
         self._confirmations: dict[str, Literal["allow", "deny"]] = {}
         self._awaiting_confirmation: dict[str, DispatchedToolUseEvent] = {}
         self._stop = anyio.Event()
@@ -775,6 +781,28 @@ class SessionToolRunner:
             )
         )
 
+    async def _resend(self, ev: DispatchedToolUseEvent) -> None:
+        """Retry posting an already-computed result without re-running the tool.
+
+        Reached from :meth:`_dispatch_loop` when reconcile re-enqueues a call
+        whose tool already ran but whose result was not yet confirmed posted.
+        """
+        tool_result, is_error, confirmation = self._executed[ev.id]
+        sent = await self._send_result(tool_result, ev.id)
+        if sent:
+            self._executed.pop(ev.id, None)
+        await self._surface_call(
+            DispatchedToolCall(
+                event=ev,
+                result=tool_result,
+                tool_use_id=ev.id,
+                name=ev.name,
+                is_error=is_error,
+                posted=sent,
+                confirmation=confirmation,
+            )
+        )
+
     async def _surface_call(self, call: DispatchedToolCall) -> None:
         """Yield ``call`` to the consumer, tolerating a consumer that left early.
 
@@ -807,7 +835,16 @@ class SessionToolRunner:
                         # posted and the DispatchedToolCall enqueued before the
                         # cancel propagates.
                         with anyio.CancelScope(shield=True):
-                            await self._execute(ev, confirmation)
+                            if ev.id in self._executed:
+                                # Reconcile re-enqueued a call whose tool already
+                                # ran (the earlier post failed or was never
+                                # confirmed). Retry posting the result we already
+                                # have; never call _execute again, which would
+                                # re-run the tool itself and is unsafe for a
+                                # side-effecting tool such as bash or a file write.
+                                await self._resend(ev)
+                            else:
+                                await self._execute(ev, confirmation)
                 finally:
                     if confirmation == "allow":
                         # The user-approved call is fully disposed of (executed,
@@ -872,7 +909,13 @@ class SessionToolRunner:
                 content = tool_error_content(e)
                 is_error = True
             tool_result = _build_result_event(ev, content, is_error)
+            # Recorded before the send attempt: the tool has now run, so a later
+            # reconcile must never dispatch this id through _execute again,
+            # regardless of whether the send below succeeds.
+            self._executed[ev.id] = (tool_result, is_error, confirmation)
             sent = await self._send_result(tool_result, ev.id)
+            if sent:
+                self._executed.pop(ev.id, None)
         await self._surface_call(
             DispatchedToolCall(
                 event=ev,
--- a/tests/lib/tools/test_session_runner.py
+++ b/tests/lib/tools/test_session_runner.py
@@ -1595,3 +1595,48 @@ def test_to_session_content_tool_reference_stringified() -> None:
     block = {"type": "tool_reference", "tool_name": "weather"}
     out = _to_session_content([block])
     assert out == [{"type": "text", "text": session_runner_mod.json.dumps(block)}]
+
+
+@pytest.mark.asyncio()
+async def test_reconnect_does_not_re_execute_tool_after_failed_send(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A tool whose result post permanently failed must not be re-executed when
+    a later stream reconnect re-runs ``_reconcile``.
+
+    ``_reconcile`` re-enqueues every ``agent.tool_use`` not yet in ``_answered``,
+    which is exactly the state of a call whose ``_send_result`` hit a permanent
+    4xx. The other duplicate-delivery sources this file guards against are
+    duplicates of already-settled work; this one is not, so the same event was
+    dispatched again and ``_execute`` re-ran the tool rather than just retrying
+    the post. For a side-effecting tool that is at-least-once instead of
+    at-most-once execution.
+    """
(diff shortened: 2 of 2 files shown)


PranavMishra28:
Small fix for a real at-most-once bug in the Managed Agents `SessionToolRunner` — a stream reconnect after a failed result-post could re-run the tool (so a `bash`/file-write call can fire twice), and this keeps the reconnect retrying only the post, with a deterministic regression test; would appreciate a look whenever someone has a cycle 🙏

PranavMishra28:
@dtmeadows-ant flagging this your way since you last touched the session runner. small fix — it re-executes a tool on stream reconnect when the result post failed (so bash/file-write tools can run twice). CI's paused awaiting approval since it's my first PR here, so an approve-and-run would get the checks green. no rush on the review itself, thanks!

kalyanamdewri (commented):
### Fault-injection result on `57f46e2bef99675a2405a51d495de44bfefe5ecb`

For one logical call ID `toolu_slot3_fixed_001`, the tool executed `1` time(s) and result posting was attempted `4` time(s) across `1` reconnect(s).

Cancellation during `tool result submission after execution: one logical submission failure exhausted the default three physical send attempts, then the SSE stream disconnected` left `0 background asyncio` live tasks (`asyncio.all_tasks()` excluding one harness task; 2 threads in-loop and 1 after loop close) and `0 tracked custom` open transports (one was created and closed; reconnects are stream GETs minus one, not TCP connections) after `2000` ms. Base produced `2 executions, with the successful fourth POST carrying the recomputed execution:2 result`; this PR produced `1 execution, with all four POST attempts carrying the cached execution:1 result`.

This `SUPPORTS` **at-most-once execution within one runner lifetime**. I did not test process restart, so this is not evidence of durable exactly-once behavior.

Seed `none-deterministic-fixture-v1`; minimized trace `40 per PR run (41 per base run)` events; artifact `/tmp/fault_injection_1750/fault_injection_1750_artifact.tar.gz` (`sha256:2edb9de6a575f760562cfe336370bc5b31feffa5d214865dbcb05007544dc8ef`).

PranavMishra28:
appreciate the fault-injection @kalyanamdewri — at-most-once within a single runner lifetime is exactly the guarantee this is going for (the cached result gets re-POSTed on reconnect, the tool itself never re-runs), so that lines up with what you saw. and agreed, durable exactly-once across a process restart is a separate problem and out of scope here. thanks for taking the time to test it!

PranavMishra28:
Rebased onto main. This had gone conflicted against the confirmation-gating work that landed since, and the overlap was in exactly the three places this touches, so it needed real adaptation rather than a merge resolution.

Two things changed in how the fix is expressed:

`_executed` now carries the call's confirmation alongside the cached result, so a resent `DispatchedToolCall` reports the same verdict as the original instead of dropping to `None`. Without that, a resend of an `always_ask` tool would surface with no confirmation attached.

`_resend` reuses the new `_surface_call` helper rather than the local one I had added, so the consumer-gone handling (`BrokenResourceError` plus the `ClosedResourceError` case that helper added for the deny path) stays in one place.

The dispatch-loop guard sits inside the existing `try` / `finally`, so the idle-clock release for an approved call still runs on the resend path.

Behavior is unchanged from the original: an id already in `_executed` retries the post only, never re-runs the tool.

Verification on the rebased branch: 53 passed in `tests/lib/tools/test_session_runner.py`, ruff clean, mypy and pyright clean on the changed file. `tests/lib/` shows 17 pre-existing failures that reproduce identically on clean main without this change.

PR #1751, opened 2026-07-11
feat(environments): add defer_ack to work poller to fix dispatch-mode strand (#1746)

### LINKED ISSUE #1746: environments work poller: ack-before-yield makes any consumer failure permanently strand the work item — request deferred ack or a nack/fail API
`anthropic.lib.environments._poller` acknowledges every work item **before** yielding it to the consumer (`_poller.py:234`; docstring at `:66`: "Each yielded `BetaSelfHostedWork` has already been ack'd"). There is no deferred-ack option and no nack/fail API. Verified identical from 0.109.1 through 0.116.0.

For self-hosted managed-agents environments this makes any consumer-side exception after the yield permanently strand that session:

- the acked item freezes in `state='starting'` with `latest_heartbeat_at=null`;
- `reclaim_older_than_ms` cannot reach it (it only reaches unacknowledged items), and lease-TTL re-queue never fires (requires a first heartbeat);
- the platform keeps one work item per session, so all subsequent `user.message` events park in `requires_action` with **no new work enqueued** — the session is wedged until an operator runs `work.stop(work_id, force=True)` and sends a `user.interrupt`.

We hit this in production twice in one day (different consumer exceptions, same terminal state) and separately counted 29 items stuck in `starting` with null heartbeats.

## Ask

Either of:
1. **Deferred ack**: an opt-in poller mode that acks after the consumer callback returns successfully (at-least-once semantics), or
2. **nack/fail API**: a way for a consumer to return a failed item to a re-pendable state.

Related, happy to file separately: documenting which work-item states re-enqueue on which user events (empirically: `stopped` re-pends on message/interrupt; a dead-heartbeat `started` item only after a server-side stop ~5 min later; an acked-but-never-heartbeated `starting` item never), and a server-side janitor for `starting` + null-heartbeat items older than a TTL.

## Workaround we ship today

Consumer retry + `work.stop(force=True)` on final failure, plus a scheduled sweep that force-stops stale `starting` items and interrupts sessions parked in `requires_action` with no live worker. Effective, but every self-hosted-environment consumer will independently rediscover this failure class — ack-after-success semantics in the SDK would eliminate it.

PranavMishra28:
## What was broken

In the poller's **drain-and-dispatch** shape (`auto_stop=False` — the webhook/handoff pattern the docstring describes), `iter_work` / `aiter_work` `ack` a claimed item (transitioning it `queued → starting`) *before* yielding it to the process that will actually run it. If that handoff process dies before its first heartbeat, the item is left in `starting` with a null heartbeat — and `poll(reclaim_older_than_ms=…)` only reclaims **unacknowledged** work, so the item is reachable by neither `reclaim` nor lease-TTL requeue. It's permanently stranded. (This is issue #1746, reported with 29 real items stuck in `starting`.)

## Why not a nack

The issue proposed either a deferred-ack mode *or* a nack/fail API. The `Work` resource exposes only `poll` / `ack` / `stop` / `heartbeat` — there is no `nack`/`fail` endpoint — so a true nack can't be implemented client-side. That leaves deferred ack.

## The fix

Adds an opt-in `defer_ack: bool = False` to `iter_work` / `aiter_work` and the public `poller()` entry point:

- **Default (`False`)** — exact current behavior (ack before yield). Fully backward-compatible; no existing caller changes.
- **`True`** — the poller yields the item still **unacknowledged** (`queued`); the consumer / handoff owner performs the `ack` (and `stop`) when it commits to processing. A handoff that dies before acking now leaves the item recoverable via `reclaim_older_than_ms` / lease TTL instead of stranded.
- **Guard** — `defer_ack=True` requires `auto_stop=False` (the dispatch pattern where the consumer owns the lifecycle); pairing it with `auto_stop=True` raises `ValueError`, since the poller can't `stop` an item whose `ack` it deferred. Validated before the first poll, sync and async.

## How it's tested

Uses the existing `FakeWork` / `FakeAsyncWork` harness in `tests/lib/environments/test_poller_method.py` — no live API. New tests (sync + async):
- `..._defer_ack_yields_unacked` — items are handed off with **no** `ack`/`stop` call by the poller.
- `..._defer_ack_does_not_ack_on_body_raise` — the strand path: a handoff that fails after the yield leaves the item unacknowledged (i.e. reclaimable), asserting `ack_calls == []`.
- `..._defer_ack_requires_auto_stop_false` — the `ValueError` guard fires before any poll.

Full local run for this PR: `uv run pytest tests/lib/` (752 passed; the 2 failures in `tests/lib/test_credentials.py` reproduce identically on clean `main` and are unrelated to this change), `ruff check` / `ruff format --check` clean, `mypy .` clean (1052 files), `pyright` 0/0/0.

## Scope / boundary note

The behavior change is entirely in the hand-written `src/anthropic/lib/environments/_poller.py`. The 8-line change to `poller()` in `resources/beta/environments/work.py` is a passthrough of the new kwarg; `poller()`/`worker()` are maintained custom methods there (they import from `lib/` and have this dedicated test file), but if the codegen config needs a matching entry for the new parameter, happy to adjust.

## What this deliberately does not change

- No change to the default ack-before-yield behavior, the retry/backoff policy, or the `auto_stop=True` runner path.
- Does not add a nack API (no server endpoint for it).

Fixes #1746

### DIFF
--- a/src/anthropic/lib/environments/_poller.py
+++ b/src/anthropic/lib/environments/_poller.py
@@ -59,15 +59,16 @@ def iter_work(
     reclaim_older_than_ms: int | None = None,
     drain: bool = False,
     auto_stop: bool = True,
+    defer_ack: bool = False,
     extra_headers: Headers | None = None,
 ) -> Iterator[BetaSelfHostedWork]:
     """Iterate work items claimed from a self-hosted environment.
 
-    Each yielded :class:`BetaSelfHostedWork` has already been ack'd. The ``work``
-    resource must be bound to a client authenticated for the environment — the
-    poller itself does not handle credentials. Use
-    ``client.beta.environments.work.poller(...)`` for the user-facing entry
-    point that constructs a scoped sub-client for you.
+    Each yielded :class:`BetaSelfHostedWork` has already been ack'd, unless
+    ``defer_ack=True`` (see below). The ``work`` resource must be bound to a
+    client authenticated for the environment — the poller itself does not handle
+    credentials. Use ``client.beta.environments.work.poller(...)`` for the
+    user-facing entry point that constructs a scoped sub-client for you.
 
     Two consumption shapes are supported:
 
@@ -92,6 +93,17 @@ def iter_work(
         loop body completes. Set False when the work item is handed off to
         another process that owns the stop call — otherwise the lease is
         terminated out from under it.
+      defer_ack: When True, yield the item still *unacknowledged* (``queued``)
+        instead of ack'ing it before the yield, leaving the consuming/handoff
+        owner to call ``work.ack`` when it commits to processing. This closes
+        the drain-and-dispatch strand: with the default ack-before-yield, an
+        item whose handoff process dies before its first heartbeat is left in
+        ``starting`` with a null heartbeat, and ``poll(reclaim_older_than_ms=…)``
+        only reclaims *unacknowledged* work, so the item never resurfaces.
+        Deferring the ack keeps a failed handoff recoverable via
+        ``reclaim_older_than_ms`` / lease TTL. Requires ``auto_stop=False``
+        (the consumer owns the full lifecycle — both ack and stop); pairing it
+        with ``auto_stop=True`` raises ``ValueError``.
       reclaim_older_than_ms: Forwarded to ``work.poll``. Reclaim un-ack'd work
         older than this many ms. Useful in drain mode so a dead runner's
         work re-surfaces on the next webhook delivery.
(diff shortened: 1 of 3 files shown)


PranavMishra28:
@arcaputo3 shd fix your issue now

PranavMishra28:
@TomerAberbach you wrote the poller so sending this your way — small opt-in fix (defer_ack) for the drain-mode strand where a handoff that dies before its first heartbeat leaves the work item stuck in starting instead of reclaimable. CI's paused awaiting approval since it's my first PR here; an approve-and-run would surface the checks. whenever you've got a sec, thanks!

PranavMishra28:
Hi @TomerAberbach or @anthropics/sdk team - gently following up on this PR when you have a moment! Could a maintainer please approve the 2 pending workflows and provide a code owner review? Thanks!

PranavMishra28:
I don't think you have the write access, and therefore I am unable to merge the PR.

PR #1754, opened 2026-07-13
fix: handle bare unparameterized list in construct_type

### LINKED ISSUE #1626: BUG: construct_type() raises IndexError when called with bare unparameterized list

gautamkishore:
Fixes #1626

When `construct_type` is called with a bare `list` (no type parameter, e.g. `construct_type(value=[...], type_=list)`), `get_args(list)` returns an empty tuple, causing `IndexError` at `args[0]`.

This PR adds a guard to return the value unchanged when there are no type arguments.

### DIFF
--- a/src/anthropic/_models.py
+++ b/src/anthropic/_models.py
@@ -668,6 +668,9 @@ def construct_type(*, value: object, type_: object, metadata: Optional[List[Any]
         if not is_list(value):
             return value
 
+        if not args:
+            return value
+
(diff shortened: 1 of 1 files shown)


craigie-ant:
👋 thanks for the PR! closing because of https://github.com/anthropics/anthropic-sdk-python/issues/1626#issuecomment-5257036785

PR #1757, opened 2026-07-13
fix(examples): replace dated model alias in tools example

### LINKED ISSUE #1596: fix(examples): replace dated model IDs with non-dated aliases
## Problem
Several files in the `examples/` directory use hardcoded dated model strings 
like `claude-sonnet-4-5-20250929` which will return a 404 once Anthropic 
deprecates that specific version snapshot.

Affected file:
- `examples/tools.py` — uses `claude-sonnet-4-5-20250929`

## Expected behaviour
Examples should use stable non-dated aliases like `claude-sonnet-4-5` 
which always resolve to the latest available version.

## Fix
Simple one-line change per file replacing the dated string with its alias.
I'm happy to submit a PR for this.

nightcityblade:
Fixes #1596

Replace the dated model ID in `examples/tools.py` with the stable `claude-sonnet-4-5` alias so the example does not drift into a 404 as snapshots are retired.

Local verification:
- `python3 -m compileall examples/tools.py`

### DIFF
--- a/examples/tools.py
+++ b/examples/tools.py
@@ -21,7 +21,7 @@ tools: list[ToolParam] = [
 ]
 
 message = client.messages.create(
-    model="claude-sonnet-4-5-20250929",
+    model="claude-sonnet-4-5",
     max_tokens=1024,
     messages=[user_message],
     tools=tools,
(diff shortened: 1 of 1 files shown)


nightcityblade:
Closing because upstream commit ab10964 has already replaced the dated model in examples/tools.py with the newer stable claude-sonnet-5 alias, so replaying this branch would regress the example to claude-sonnet-4-5. Thanks.

PR #1758, opened 2026-07-13
fix: honor aws_profile when inferring Bedrock region

nightcityblade:
Fixes #892

Thread the constructor `aws_profile` through `_infer_region()` so Bedrock region inference consults the same profile the client will actually use instead of always probing the default profile.

Local verification:
- `PYTHONPATH=/tmp/anthropic-sdk-python/src python3.9 -m pytest tests/lib/test_bedrock.py -k "region_infer_from_specified_profile or region_infer_from_constructor_profile" -q -o addopts=`


nightcityblade:
Updated the branch with current upstream main and resolved the test-file conflict while preserving both the new upstream Bedrock coverage and this PR's constructor-profile regression test. Targeted verification passes: 3 tests passed.

PR #1776, opened 2026-07-23
fix(streaming): get_final_message() drops container, breaking code execution continuation

### LINKED ISSUE #1424: `get_final_message()` does not propagate `message_delta.container` into aggregated Message (breaks code_execution continuation)

epatey:
When streaming with `client.messages.stream()`, the final message's `container` field is always `None`, even though the API sends it in the `message_delta` event. The streaming code copies `stop_reason`, `stop_sequence`, `stop_details`, and usage from that event, but not `container`. The equivalent code for `client.beta.messages.stream()` does copy it.

Without the container id, a caller can't continue a conversation that has unfinished code execution — the next request fails with `400: container_id is required when there are pending tool uses generated by code execution with tools.`

This adds the missing copy, plus sync and async tests that fail without it.

Fixes #1424.

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -501,6 +501,7 @@ def accumulate_event(
         if content_block.type == "text" and is_given(output_format):
             content_block.parsed_output = parse_text(content_block.text, output_format)
     elif event.type == "message_delta":
+        current_snapshot.container = event.delta.container
         current_snapshot.stop_reason = event.delta.stop_reason
         current_snapshot.stop_sequence = event.delta.stop_sequence
         if event.delta.stop_details is not None:
--- a/tests/lib/streaming/fixtures/container_response.txt
+++ b/tests/lib/streaming/fixtures/container_response.txt
@@ -1,0 +1,17 @@
+event: message_start
+data: {"type":"message_start","message":{"id":"msg_01Gh8kQEinLbnrkVJ9BqrmyQ","type":"message","role":"assistant","content":[],"model":"claude-opus-4-8","container":null,"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":11,"output_tokens":1}}}
+
+event: content_block_start
+data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Running your code now."}}
+
+event: content_block_stop
+data: {"type":"content_block_stop","index":0}
+
+event: message_delta
+data: {"type":"message_delta","delta":{"container":{"id":"container_011CPTa1PkYDZagcpQ8VZCcy","expires_at":"2026-07-23T16:16:14.426901Z"},"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":6}}
+
+event: message_stop
+data: {"type":"message_stop"}
--- a/tests/lib/streaming/test_messages.py
+++ b/tests/lib/streaming/test_messages.py
@@ -105,6 +105,11 @@ def assert_tool_use_response(events: list[ParsedMessageStreamEvent[None]], messa
     ]
 
 
+def assert_container_response(message: Message) -> None:
+    assert message.container is not None
+    assert message.container.id == "container_011CPTa1PkYDZagcpQ8VZCcy"
+
+
 def assert_refusal_response(message: Message) -> None:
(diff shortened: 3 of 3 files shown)


dtmeadows-ant:
Thanks @epatey for this! I just approved the workflows. Any chance you're able to retarget this PR to go into `anthropics:next` instead of main? That ensures it gets included in the next release.

tomer-ant:
See https://github.com/anthropics/anthropic-sdk-python/issues/1424#issuecomment-5257407970

PR #1782, opened 2026-07-27
fix(streaming): enforce strict index validation in event accumulator to prevent state corruption

l1ve709:
## Summary
This PR resolves an unhandled edge case in the streaming parser where out-of-sequence or dropped `content_block_start` events could silently corrupt the internal accumulator state, leading to cryptic `IndexError` exceptions during downstream `content_block_delta` processing.

## Root Cause Analysis
The accumulator previously appended `content_block_start` payloads blindly without asserting the sequence invariant (as noted by the legacy `# TODO: check index` comment). If an SSE event is dropped or misordered by the network layer, `current_snapshot.content` desynchronizes from the expected block sequence. 

When a subsequent `content_block_delta` arrives, the parser relies on `content = current_snapshot.content[event.index]`. If the internal state length is misaligned with the incoming `event.index`, it triggers an obscure `IndexError` inside the delta router, making it extremely difficult for developers to debug the underlying network/API failure.

## Changes Made
- Introduced a strict sequence invariant check during `content_block_start` evaluation for both standard (`_messages.py`) and beta (`_beta_messages.py`) streams.
- The parser now employs a fail-fast mechanism by explicitly raising a `RuntimeError` describing the index mismatch, guaranteeing stream integrity.

## Validation
- Adheres to existing `ruff` static analysis and formatting guidelines.
- The state machine mathematically ensures `event.index == len(current_snapshot.content)` before allowing `.append()` mutations.

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -475,7 +475,11 @@ def accumulate_event(
         raise RuntimeError(f'Unexpected event order, got {event.type} before "message_start"')
 
     if event.type == "content_block_start":
-        # TODO: check index
+        if event.index != len(current_snapshot.content):
+            raise RuntimeError(
+                f"Unexpected content block index: {event.index}. "
+                f"Expected index {len(current_snapshot.content)} based on current snapshot state."
+            )
         current_snapshot.content.append(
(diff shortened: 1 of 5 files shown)


tomer-ant:

PR #1786, opened 2026-07-28
fix(environments): warn when heartbeat signals shutdown on first beat

### LINKED ISSUE #1779: Self-hosted environment: worker spawned per claimed work item is told to shut down before executing (sandbox-per-session handoff broken; in-process pattern works)
## Summary

In a **self-hosted environment** (Managed Agents), the documented "sandbox per session" pattern — a dispatcher claims a work item via `ant beta:worker poll --on-work spawn.sh`, then spawns a fresh container whose worker process is supposed to execute that already-claimed item — reliably fails. The spawned worker starts fine, reaches `api.anthropic.com` fine, then **shuts itself down within 2–6 seconds without executing anything** (exit code 0). The session is left permanently stuck at `stop_reason: requires_action` and the work item is never re-offered to any worker.

We reproduced the identical failure with **two independent client implementations**:

- the `ant` CLI (`ant beta:worker run`), versions 1.15.0 **and** 1.17.0
- this SDK's `EnvironmentWorker.handle_item()`

Both fail at the same point, in the same way, which suggests the issue is server-side (or an undocumented requirement of the claim-handoff contract) rather than a client defect. Filing here because the SDK repro is the cleanest minimal case, and we don't have a direct support line.

## Environment

- Environment ID: `env_01UPQK3XYX67tnteLTvZw891` (type `self_hosted`)
- Agent ID: `agent_014h5P54ckDkXpUG1NVdqY4G`
- Runtime: Linux amd64, Kubernetes Job pod, Debian bookworm-slim base
- Reproduced identically across 5+ sessions, e.g. `sesn_011zBktK84XqgQapwZCSunbU`, `sesn_017jQMiZ3mSv2bbtPFAWtayn`, `sesn_01GSgDdFAJ9JPKgx8oSt1UA6`, `sesn_01BjrqoHYCysBHz4tFboyfDU`, `sesn_014f4Y2BQrjJFaHZavJCxxVH`
- SDK repro session: `sesn_01GvoUeeRqmHe6dKbfPkCtcH`

## Reproduction steps

1. Create a self-hosted environment (`POST /v1/environments`, `config.type: self_hosted`); generate its environment key via the Console.
2. Run a dispatcher: `ant beta:worker poll --on-work ./spawn.sh`, where `spawn.sh` follows the documented shape — one fresh container per claimed work item, with `ANTHROPIC_SESSION_ID`, `ANTHROPIC_WORK_ID`, `ANTHROPIC_ENVIRONMENT_ID`, `ANTHROPIC_ENVIRONMENT_KEY` injected as env vars (our implementation spawns a Kubernetes Job rather than `docker run`; functionally equivalent).
3. Create an Agent (`POST /v1/agents`) and a Session (`POST /v1/sessions` referencing agent + environment_id).
4. Send a trivial `user.message` event, e.g. "List every file in /workspace and print their sizes."

### SDK variant of the spawned worker (fails identically to the CLI)

```python
async with AsyncAnthropic(auth_token=environment_key) as client:
    worker = EnvironmentWorker(
        client, environment_id=environment_id,
        environment_key=environment_key, workdir="/workspace",
    )
    await worker.handle_item(
        work_id=work_id, environment_id=environment_id,
        session_id=session_id, environment_key=environment_key,
    )
```

## Observed behavior

- The dispatcher polls, claims exactly one work item (`work_type=session`), and spawns exactly one container per claim — this part works on every attempt.
- The spawned worker's logs show, consistently within 2–6 seconds of startup:

  ```
  msg="heartbeat reports shutdown" state=stopping
  msg="reconcile list failed" component=session-tool-runner error="context canceled"
  ```

  and the process exits 0. The SDK variant behaves the same (ran ~3s, exit 0, no error output).
- Meanwhile, via the Sessions API, the conversation proceeds normally on the model side: `user.message` → `agent.thinking` → `agent.message` → `agent.tool_use` (bash, `evaluated_permission: allow`), then `stop_reason: requires_action` referencing that tool_use, and `session.status_idle`.
- The `tool_use` is never fulfilled. No second work item is ever claimed for the session. It is stuck permanently.

## Ruled out

- **Our dispatch logic** — found and fixed two real bugs in it (Jobs keyed by `session_id` instead of `work_id`; a heredoc quoting bug); failure persisted unchanged after both fixes.
- **Env vars** — verified all four are non-empty and correct on the running container via `kubectl get pod -o jsonpath`.
- **Network** — DNS + TLS to `api.anthropic.com` confirmed working from inside the spawned pod.
- **`--max-idle` timing** — tried 300s vs. the 60s default, plus explicit `--environment-id`; no change. Reverted to the exact zero-flag documented pattern for this report.
- **Stale CLI** — bumped 1.15.0 → 1.17.0 (the version pinned in the self-hosted-sandboxes doc's own Dockerfile examples as of 2026-07-17); identical failure.

## Confirmed workaround (what narrowed it down)

The "in-process" pattern — a **single** long-running process that both claims and executes, `ant beta:worker poll --workdir /workspace` with no `--on-work` — works correctly on the first try: full tool-call cycle, results posted, clean `stop_reason: end_turn`.

So the failure is specifically in **resuming an already-claimed work item in a process other than the one that claimed it**. The observable shape (server heartbeat replying "shutdown" to the freshly spawned worker) is consistent with the claim/lease being bound to the claiming process's identity, so the handed-off worker is told to stop — but that's inference; we can't see the server side.

## Question

Is there a known issue — or an undocumented handshake/requirement — for how a spawned `ant beta:worker run` / `EnvironmentWorker.handle_item()` process adopts a work item claimed by a different (dispatcher) process via the injected env vars? The session IDs above should allow server-side correlation. We've temporarily rearchitected to the in-process pattern, but we'd like to return to sandbox-per-session isolation once this is resolved.

priyanka25aug:
## Problem

When using `ant beta:worker poll --on-work spawn.sh`, if `spawn.sh` exits immediately after launching a background container — e.g. `kubectl apply` or `docker run -d` — the poller releases the work-item lease before the spawned worker connects. The first heartbeat then returns `state: stopping`, `_heartbeat_loop` silently calls `stop.set()` and returns, and the session is permanently stuck at `stop_reason: requires_action` with no output to explain why.

This is the issue described in #1779: the failure is deterministic once you understand the cause, but the SDK gives the user nothing to go on — no warning, no pointer to the blocking requirement.

## Fix

`src/anthropic/lib/environments/_worker.py` — the hand-written carve-out, not Stainless-generated code.

**1. `_heartbeat_loop`: detect and warn on first-beat shutdown**

Add a `heartbeat_count` counter. When `state: stopping` / `lease_extended: false` arrives on the very first successful heartbeat, emit `log.warning` with an explicit message:
- names the root cause (spawn script exited before worker connected)
- gives the concrete fix (`docker run` foreground vs `-d`, `kubectl wait` vs `kubectl apply`)
- links to the self-hosted environments docs

All subsequent shutdowns continue to use `log.info` (normal end-of-session).

**2. `handle_item` docstring: add a `Note:` section**

Documents the blocking requirement inline where users reading the API are most likely to encounter it, complementing any docs-side updates.

## Testing

The change is a logging-only diagnostic path; existing behaviour for normal session teardown (`heartbeat_count > 1`) is unchanged. The new `log.warning` branch fires only when `heartbeat_count == 1` at shutdown, which is the exact condition that previously produced a silent hang.

Fixes #1779

### DIFF
--- a/src/anthropic/lib/environments/_worker.py
+++ b/src/anthropic/lib/environments/_worker.py
@@ -108,6 +108,7 @@ async def _heartbeat_loop(
     ttl = _HEARTBEAT_TTL_DEFAULT
     last = _NO_HEARTBEAT_SENTINEL
     last_success = time.monotonic()
+    heartbeat_count = 0
     while not stop.is_set():
         try:
             # Bound each heartbeat: a network blackhole must not leave us
@@ -139,11 +140,30 @@ async def _heartbeat_loop(
         else:
             last = resp.last_heartbeat
             last_success = time.monotonic()
+            heartbeat_count += 1
             if resp.ttl_seconds > 0:
                 ttl = resp.ttl_seconds
                 interval = max(1.0, min(resp.ttl_seconds / 2, _HEARTBEAT_DEFAULT))
             if resp.state in ("stopping", "stopped") or not resp.lease_extended:
-                log.info("heartbeat signals shutdown state=%s lease_extended=%s", resp.state, resp.lease_extended)
+                if heartbeat_count == 1:
+                    log.warning(
+                        "heartbeat signals shutdown on the very first beat "
+                        "(state=%s lease_extended=%s work_id=%s) — this almost always "
+                        "means your --on-work script exited before the spawned worker "
+                        "connected. Ensure spawn.sh BLOCKS until the sandbox exits: "
+                        "use 'docker run' (foreground), not 'docker run -d'; or "
+                        "'kubectl wait --for=condition=complete job/...' instead of "
+                        "'kubectl apply'. See the self-hosted environments docs.",
+                        resp.state,
+                        resp.lease_extended,
+                        work_id,
+                    )
+                else:
+                    log.info(
+                        "heartbeat signals shutdown state=%s lease_extended=%s",
+                        resp.state,
+                        resp.lease_extended,
+                    )
                 stop.set()
                 return
         # Sleep up to `interval` seconds, but wake immediately if stop is set.
@@ -351,6 +371,16 @@ class EnvironmentWorker:
         Raises:
           ValueError: if any of ``work_id`` / ``environment_id`` / ``session_id``
             / ``environment_key`` is still empty after the fallbacks.
+
+        Note:
+            When invoked from a ``worker poll --on-work`` script, that script MUST
+            block until the spawned worker process exits. Returning immediately after
+            launching a background container (e.g. ``kubectl apply``, ``docker run -d``)
+            causes the poller to mark the work item as complete before this worker
+            connects — the first heartbeat then returns ``state: stopping`` and the
+            session is permanently stuck at ``stop_reason: requires_action``. Use
+            ``docker run`` (foreground) or ``kubectl wait --for=condition=complete``
+            to block until the sandbox exits.
         """
         work_id = _require(work_id, name="work_id", env_var="ANTHROPIC_WORK_ID")
         environment_id = _require(environment_id, name="environment_id", env_var="ANTHROPIC_ENVIRONMENT_ID")


priyanka25aug:
@craigie-ant @raghu-giatec — tagging you both since this directly addresses the issue you reported and diagnosed. @craigie-ant, given you've already traced the root cause on the issue, would appreciate your review when you have a moment — happy to adjust anything that doesn't fit the codebase conventions.

dtmeadows-ant:
@priyanka25aug thanks for this. The underlying issue (#1779) has since been fixed and released, so I'm going to close this out. If you still hit it on the latest release, please open a fresh issue with a repro and we'll take a look.

PR #1788, opened 2026-07-30
Read PathLike inside (name, content) file tuples

### LINKED ISSUE #1769: AttributeError: 'WindowsPath' object has no attribute 'read'
### Description
The `file=` parameter of `client.beta.files.upload()` is annotated as `FileTypes`, which includes `Tuple[Optional[str], FileContent]`, and `FileContent` includes `os.PathLike`. Based on that, passing a filename together with a `Path` object should be valid — and it passes type checking.

At runtime it fails with an `AttributeError`.

### Repro
```python
from pathlib import Path
from anthropic import Anthropic

client = Anthropic()

path = Path("Wisen.pdf")
file_obj = client.beta.files.upload(file=("Wisen-IT-Solutions.pdf", path))
```

### Actual
AttributeError: 'WindowsPath' object has no attribute 'read'

### Expected
One of:
- the `Path` is read and uploaded (consistent with passing a bare `Path`, which works fine), or
- a clear SDK-level error explaining that a `Path` is not supported inside the tuple form.

The current `AttributeError` doesn't mention the SDK, the `file=` parameter,
or what was wrong with the input, which makes it hard to diagnose.

### Working alternatives
Passing the `Path` on its own works:
```python
client.beta.files.upload(file=path)                        # OK
```
Passing a handle or bytes inside the tuple also works:
```python
client.beta.files.upload(file=("Wisen-IT-Solutions.pdf", open("Wisen.pdf", "rb")))   # OK
client.beta.files.upload(file=("Wisen-IT-Solutions.pdf", path.read_bytes()))         # OK
```
Only the `Path`-inside-a-tuple combination fails.

### Environment
- anthropic 0.117.0
- Windows
- Python 3.11.15

kilanassah:
Fixes #1769.

Check tuple inputs before generic `FileContent` inputs in `_transform_file` (and the async version).

`Path` is valid file content, but the current check order returns a `(filename, Path)` tuple without reading the inner path, so httpx later calls `.read()` on it and raises `AttributeError`.

Adds sync and async regression tests for filename + `Path` tuples.

### DIFF
--- a/src/anthropic/_files.py
+++ b/src/anthropic/_files.py
@@ -63,15 +63,18 @@ def to_httpx_files(files: RequestFiles | None) -> HttpxRequestFiles | None:
 
 
 def _transform_file(file: FileTypes) -> HttpxFileTypes:
+    # Check the tuple form first: is_file_content() also matches tuples, so testing it before
+    # is_tuple_t() would return a (name, PathLike) tuple unchanged, leaving the PathLike for
+    # httpx to choke on (`AttributeError: 'WindowsPath' object has no attribute 'read'`, #1769).
+    if is_tuple_t(file):
+        return (file[0], read_file_content(file[1]), *file[2:])
+
     if is_file_content(file):
         if isinstance(file, os.PathLike):
             path = pathlib.Path(file)
             return (path.name, path.read_bytes())
 
         return file
-
-    if is_tuple_t(file):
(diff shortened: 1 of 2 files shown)


tomer-ant:
Thanks, but this is fixed (see https://github.com/anthropics/anthropic-sdk-python/issues/1769#issuecomment-5257978021)

PR #1789, opened 2026-07-30
fix(files): read PathLike content inside (filename, content) tuples (#1769)

kilanassah:
### Problem
`client.beta.files.upload(file=("name.pdf", Path(...)))` raises
`AttributeError: 'WindowsPath' object has no attribute 'read'`, even though the tuple form
`Tuple[Optional[str], FileContent]` (with `FileContent` including `os.PathLike`) type-checks.
A bare `Path` works; only a `Path` *inside the tuple* fails. Reported in #1769.

### Cause
In `_transform_file` / `_async_transform_file`, `is_file_content()` returns `True` for any
tuple and is checked **before** `is_tuple_t()`. So a `(filename, Path)` tuple enters the
file-content branch, fails the `isinstance(file, os.PathLike)` check, and is returned
verbatim — the `PathLike` content is never read. The `is_tuple_t` branch that routes
`file[1]` through `read_file_content` is unreachable for tuples.

### Fix
Handle the tuple form first in both the sync and async transformers. Minimal reorder; no
behavior change for any already-working input.

### Tests
Adds `test_tuple_with_pathlib_reads_content` + async counterpart. Both fail on `main`
(reproduce #1769) and pass with the fix. Full `tests/test_files.py` stays green (16 passed).

### DIFF
--- a/src/anthropic/_files.py
+++ b/src/anthropic/_files.py
@@ -63,15 +63,19 @@ def to_httpx_files(files: RequestFiles | None) -> HttpxRequestFiles | None:
 
 
 def _transform_file(file: FileTypes) -> HttpxFileTypes:
+    # A 2/3-tuple `(filename, content[, content_type])` must be handled before the
+    # generic file-content check: `is_file_content` also matches tuples, so checking it
+    # first would return the tuple verbatim and never read a PathLike `content` — the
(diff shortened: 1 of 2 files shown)


tomer-ant:
Thanks, but this is fixed (see https://github.com/anthropics/anthropic-sdk-python/issues/1769#issuecomment-5257978021)

PR #1797, opened 2026-08-04
fix(streaming): resolve unchecked content block index and type safety issues

vjaudir8123-debug:
## Summary
This PR addresses two critical bugs in the streaming message parser (`_messages.py` and `_beta_messages.py`) that can cause runtime crashes or data corruption.

### 1. Missing Content Block Index Validation
The parser previously had a `# TODO: check index` comment and blindly appended new blocks on `content_block_start`. If events arrived out of order or an index was skipped, the subsequent `content_block_delta` would either overwrite the wrong block or crash with an `IndexError`. This PR correctly pads the list up to `event.index` before inserting the block.

### 2. `construct()` Bypassing Type Validation
The snapshot initialization used Pydantic`s `construct(**event.message.to_dict())`. Because `construct()` skips validation and deep coercion, nested fields like `usage` were left as raw Python dictionaries rather than Pydantic objects. When a subsequent usage delta arrived, `current_snapshot.usage.output_tokens = ...` would crash with `AttributeError: 'dict' object has no attribute 'output_tokens'`. This PR fixes it by using `construct_type(type_=...)`, ensuring correct instantiation of nested models.

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -469,18 +469,18 @@ def accumulate_event(
     if current_snapshot is None:
         if event.type == "message_start":
             return cast(
-                ParsedBetaMessage[ResponseFormatT], ParsedBetaMessage.construct(**cast(Any, event.message.to_dict()))
+                ParsedBetaMessage[ResponseFormatT],
+                construct_type(type_=ParsedBetaMessage, value=event.message.to_dict()),
             )
 
         raise RuntimeError(f'Unexpected event order, got {event.type} before "message_start"')
 
     if event.type == "content_block_start":
-        # TODO: check index
-        current_snapshot.content.append(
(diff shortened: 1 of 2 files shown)


tomer-ant:

vjaudir8123-debug:
Understood on the index checking! Makes total sense that you do not want to add defensive padding for third-party proxies.

However, this PR also included a completely separate fix for a Pydantic type safety bug. `ParsedMessage.construct()` bypasses type coercion, which leaves nested fields (like `usage`) as raw Python dictionaries rather than Pydantic objects. When a subsequent usage delta arrives, `current_snapshot.usage.output_tokens = ...` will crash with `AttributeError: 'dict' object has no attribute 'output_tokens'`.

Since that is a pure Python bug internal to the SDK (unrelated to proxies), I have extracted just the Pydantic fix into a clean, targeted PR here: #1861.

Issue #1801, opened 2026-08-05
Streaming: response pausing for a client tool while a code execution is pending never includes the container id — continuation then rejected as "container_id is required"


Joon-June:
## Summary

When a streamed response ends with `stop_reason: "tool_use"` (pausing for a client tool call) **while a code execution is still pending** in the same message, the top-level `container` object is **never delivered** — not on `message_start`, not on `message_delta`, and the string `container` does not appear anywhere in the raw SSE bytes.

The continuation for that pause **requires** the container id whenever the request carries pending tool uses **generated by code execution** — which includes server tools invoked from sandbox code (`web_fetch`/`web_search` called inside the execution; no programmatic client-tool calling involved). In that state, sending the tool_result without `container` is rejected with

```
400 invalid_request_error: container_id is required when there are pending tool uses generated by code execution with tools.
```

The client is asked to return an id it was never given. A pause whose history carries no sandbox-generated uses happens to be accepted without the id — but the id was still never delivered, so any session where the sandbox has fetched anything is unrecoverable. There is no client-side workaround other than abandoning the pending execution (stripping its blocks from history and retrying).

This appears deterministic: **21/21 attempts** across every configuration we tried reproduce the omission, while control cases (code execution that completes within the message — successfully, with an in-code error, or followed by a client tool pause) deliver `container` on `message_delta` in **8/8 attempts**. The single discriminating variable is whether the execution is still pending when the message ends.

## Environment

- `anthropic` 0.102.0 and 0.120.2 (latest), Python 3.12.12, direct `api.anthropic.com` (no proxy)
- `model: claude-sonnet-5`, streaming via `client.messages.create(..., stream=True)`
- Reproduces identically on the GA and beta surfaces (`client.beta.messages.create`, with and without `betas=["code-execution-2025-08-25"]`), with the sandbox provisioned implicitly by `web_search_20260209`/`web_fetch_20260209`, and with `code_execution_20260521` registered explicitly

## Reproduction

```python
from anthropic import Anthropic

client = Anthropic()
TOOLS = [
    {"name": "read_data", "description": "Read rows from the operator's dataframe.",
     "input_schema": {"type": "object",
                      "properties": {"offset": {"type": "integer"}, "limit": {"type": "integer"}},
                      "required": ["offset", "limit"]}},
    {"type": "web_search_20260209", "name": "web_search", "max_uses": 10},
[...]

PR #1802, opened 2026-08-05
fix(streaming): accumulate container from message_delta in non-beta helper

Joon-June:
## What

The non-beta streaming accumulator (`lib/streaming/_messages.py`) does not copy `event.delta.container` into the message snapshot, while the beta accumulator (`lib/streaming/_beta_messages.py`) does. As a result, a final message built by `client.messages.stream()` always has `container=None`, even when the server delivers the id on `message_delta` — breaking code-execution container reuse for non-beta streaming users (the id is needed as the `container` request parameter on follow-up requests).

This PR mirrors the beta accumulator's line into the non-beta `accumulate_event`, plus a fixture + test asserting the accumulated message carries the container id.

## Test

`tests/lib/streaming/test_messages.py` — new `test_container_accumulated_from_message_delta` using a fixture whose `message_delta` carries a `container` object; full file passes (14/14).

## Related

Found while investigating #1801 (server-side: the pause-for-client-tool response omits `container` entirely). This PR is the independent client-side half: even when the server does send the id, the non-beta helper drops it.

### DIFF
--- a/src/anthropic/lib/streaming/_messages.py
+++ b/src/anthropic/lib/streaming/_messages.py
@@ -501,6 +501,7 @@ def accumulate_event(
         if content_block.type == "text" and is_given(output_format):
             content_block.parsed_output = parse_text(content_block.text, output_format)
     elif event.type == "message_delta":
+        current_snapshot.container = event.delta.container
         current_snapshot.stop_reason = event.delta.stop_reason
         current_snapshot.stop_sequence = event.delta.stop_sequence
         if event.delta.stop_details is not None:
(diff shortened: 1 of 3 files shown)


tomer-ant:
Thanks, but this should be fixed by https://github.com/anthropics/anthropic-sdk-python/pull/1810/changes/fc1599bd4c25ad5550d55f8f1c8f5c5664e19ed1 in v0.122.0 (next release)

Joon-June:
Thanks @tomer-ant  — just checked fc1599b and it indeed covers everything this PR did, so closing this makes sense. 👍

One ask though: this PR was only the client-side half. The server-side half is #1801 — when a message pauses for a client tool while a code execution is still pending, the API never sends container on the wire at all (verified via raw SSE), so no SDK fix can help. The continuation then fails with container_id is required when there are pending tool uses generated by code execution with tools, killing the turn.

This is still hitting us in production — when you have a moment, would you mind taking a look at #1801 as well (or routing it to whoever owns the API side)?

tomer-ant:
Sorry, just saw this!

You are correct that this was an API bug. It should now be fixed :)

https://github.com/anthropics/anthropic-sdk-python/issues/1801#issuecomment-5344398126

PR #1813, opened 2026-08-09
fix: support list-form JSON schema types

### LINKED ISSUE #1764: transform_schema raises AssertionError on valid list-form types, e.g. {"type": ["string", "null"]}
### Summary

`transform_schema` (the structured-output schema transformer used by `messages.parse` / `beta.messages.parse`) crashes with an `AssertionError` on any schema that uses the list form of `type` — e.g. `{"type": ["string", "null"]}` — which is valid JSON Schema and explicitly supported by the Anthropic structured-outputs API.

The failure happens at request-build time, before any API call.

### Repro

```python
from anthropic import transform_schema

# Valid JSON Schema; also documented as supported by the structured-outputs API
transform_schema({"type": ["string", "null"]})
```

One-liner:

```bash
python -c 'from anthropic import transform_schema; transform_schema({"type": ["string", "null"]})'
```

Actual (on `anthropic==0.112.0`; the relevant code is unchanged on current `main` of `src/anthropic/lib/_parse/_transform.py`):

```
```

A more realistic nested variant fails the same way:

```python
transform_schema({
    "type": "object",
    "properties": {"name": {"type": ["string", "null"]}},
    "required": ["name"],
})
```

### Expected

An equivalent transformed schema, e.g. rewriting the type union into `anyOf` branches:

```json
{"anyOf": [{"type": "string"}, {"type": "null"}]}
```

### Why this should be supported

- JSON Schema allows `type` to be "a string or an array of strings" — the list form is standard, not an extension.
- The [structured outputs docs](https://docs.claude.com/en/docs/build-with-claude/structured-outputs) explicitly document type arrays as supported (and meter them): *"Total parameters that use `anyOf` or type arrays (for example, `"type": ["string", "null"]`) across all strict schemas."* So the API accepts this shape; only the SDK-side transformer rejects it.
- Nullable-type unions are pervasive in real schemas: Pydantic `Optional[...]` fields in older/alternative generators, OpenAPI 3.1 output, GraphQL-derived schemas, and MCP tool `inputSchema`s commonly emit `"type": [..., "null"]`.

### Root cause

In `src/anthropic/lib/_parse/_transform.py`, `type_` is typed and handled as a scalar:

```python
...
    pass
else:
    assert_never(type_)
```

A list value matches none of the scalar branches and falls into `assert_never`.

### Suggested fix

When `type_` is a list, canonicalize it before the scalar handling, e.g. rewrite `{"type": [t1, t2, ...], **rest}` into an `anyOf` of single-type branches (with sibling constraints attached to the non-`null` branches) and recurse. Single-element lists can collapse to the scalar form.

### Related (not duplicates)

- #1691 handles empty/typeless `{}` sub-schemas (`ValueError` path) — a different gap in the same function.
- #1185 discusses grammar-size limits and mentions type arrays as API-supported; it does not cover this SDK-side crash.

### Environment

- `anthropic==0.112.0` (also reproduces against `main`'s `_transform.py`)
- Python 3.12.13, macOS arm64

nightcityblade:
Fixes #1764

## Summary

- canonicalize JSON Schema list-form `type` values into `anyOf` branches
- keep shared metadata at the union level while applying constraints to non-null variants
- compose type unions with existing `anyOf`, `oneOf`, and `allOf` schemas
- add regression coverage for top-level and nested nullable type arrays

## Tests

- `uv run pytest -n0 tests/lib/_parse/test_transform.py` (16 passed)
- `uv run ruff check src/anthropic/lib/_parse/_transform.py tests/lib/_parse/test_transform.py`
- `uv run ruff format --check src/anthropic/lib/_parse/_transform.py tests/lib/_parse/test_transform.py`
- `uv run pyright src/anthropic/lib/_parse/_transform.py tests/lib/_parse/test_transform.py`
- `./scripts/lint`

### DIFF
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -96,10 +96,32 @@ def transform_schema(
         strict_schema["$ref"] = ref
         return strict_schema
 
-    type_: Optional[SupportedTypes] = json_schema.pop("type", None)
+    type_: Optional[SupportedTypes | list[SupportedTypes]] = json_schema.pop("type", None)
     any_of = json_schema.pop("anyOf", None)
     one_of = json_schema.pop("oneOf", None)
     all_of = json_schema.pop("allOf", None)
+
+    if is_list(type_):
+        if not type_:
+            raise ValueError("Schema 'type' array must contain at least one type.")
+        type_constraints = {
+            key: json_schema.pop(key) for key in tuple(json_schema) if key not in ("enum", "description", "title")
+        }
+        type_union = {
+            "anyOf": [{"type": variant, **(type_constraints if variant != "null" else {})} for variant in type_]
+        }
+        if is_list(any_of):
+            all_of = [type_union, {"anyOf": any_of}]
+            any_of = None
+        elif is_list(one_of):
+            all_of = [type_union, {"oneOf": one_of}]
+            one_of = None
+        elif is_list(all_of):
+            all_of = [type_union, *all_of]
+        else:
+            any_of = type_union["anyOf"]
+        type_ = None
+    type_ = cast("Optional[SupportedTypes]", type_)
 
     if is_list(any_of):
         strict_schema["anyOf"] = [transform_schema(cast("dict[str, Any]", variant)) for variant in any_of]
--- a/tests/lib/_parse/test_transform.py
+++ b/tests/lib/_parse/test_transform.py
@@ -61,6 +61,30 @@ def test_anyof_schema():
     )
 
 
+def test_type_array_schema():
+    assert transform_schema({"type": ["string", "null"]}) == snapshot(
+        {"anyOf": [{"type": "string"}, {"type": "null"}]}
+    )
+    assert transform_schema({"type": ["string"]}) == {"anyOf": [{"type": "string"}]}
+    with pytest.raises(ValueError, match="must contain at least one type"):
+        transform_schema({"type": []})
+
+
+def test_type_array_schema_with_composition():
+    union = {"anyOf": [{"type": "string"}, {"type": "null"}]}
+    for keyword in ("anyOf", "oneOf", "allOf"):
+        schema = {"type": ["string", "null"], keyword: [{"type": "string"}]}
+        assert transform_schema(schema)["allOf"][0] == union
+
+
+def test_nested_type_array_schema():
+    constraints = {"properties": {"name": {"type": "string"}}, "required": ["name"]}
+    schema = {"type": "object", "properties": {"profile": {"type": ["object", "null"], **constraints}}}
+    object_branch = {"type": "object", **constraints, "additionalProperties": False}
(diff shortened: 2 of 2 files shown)


percymcn:
I reimplemented this diff verbatim against `anthropic==0.121.0` and ran it over a battery before commenting, so everything below is measured rather than read off the diff.

**It works, including the case that separates a correct fix from a plausible one.** The risk with this bug is a fix that copies `type` through to every branch — that leaves the per-type dispatch untouched, and an object union quietly loses its `properties`. This diff moves the type-specific keys onto the non-null variants instead, so:

```py
{"type": ["object", "null"], "properties": {"o": {...}}, "required": ["a"]}
# ->
{"anyOf": [{"type": "object", "properties": {...}, "additionalProperties": False, "required": ["a"]},
           {"type": "null"}]}
```

`$defs` survive, `format`/`minItems`/`minLength` land on the right branch and demote correctly, and a 1-element `["string"]` no longer raises. All four composition branches (`anyOf`/`oneOf`/`allOf`) produce well-formed `allOf` output.

**One regression, and it changes a loud failure into a silent one.** `is_list([])` is `True`, so an empty type list falls into the new path:

```py
{"type": []}   # before: AssertionError
               # after:  {"anyOf": []}
```

An empty `anyOf` matches nothing, so the model can never produce a satisfying value — and nothing raises. Today that input is rejected; after this it ships. A guard on the empty case (raise, or treat as untyped) keeps the previous behaviour.

**The two new tests don't reach the paths the summary claims.** `test_nested_type_array_schema` does discriminate properly on the null branch, which is good. But the summary lists composition with `anyOf`, `oneOf` and `allOf`, and no test exercises any of the three — nor the `["object", "null"]`-carrying-`properties` case above, which is the one where a future refactor of the dispatch would regress silently. Both are cheap to pin.

Worth knowing on the `oneOf` branch specifically: the wrapped `{"oneOf": ...}` is then rewritten to `anyOf` by the existing transformer, so exclusivity is not preserved. That's pre-existing, not this diff, but "composes with `oneOf`" reads stronger than what happens.

**A pre-existing interaction this makes easier to hit** (verified separately with no list-form type, so not caused by this PR): a branch carrying `required` without its own `properties` comes out as `{"type": "object", "properties": {}, "additionalProperties": False, "required": ["a"]}` — unsatisfiable. Composing a type union with an object `anyOf` is a natural way to write that, and the resulting `allOf` is then unsatisfiable as a whole.

All of the above is checkable offline — `transform_schema` needs no API key — so the regression and the composition cases can go straight into `tests/lib/_parse/test_transform.py`.

nightcityblade:
Thanks for the thorough validation. I updated the current head (`6147091`) to cover the gaps you found:

- Empty type arrays now raise a clear `ValueError` instead of producing an unsatisfiable empty `anyOf`.
- The regression coverage now includes a single-element array, a nullable object carrying `properties` and `required`, and composition with `anyOf`, `oneOf`, and `allOf`.
- Type-specific constraints continue to move onto the non-null branches before recursive transformation.

Local validation passed: all 17 transformer tests, plus the repository's full lint suite (ruff, dependency-cap check, pyright, mypy, and import check).

PR #1816, opened 2026-08-10
fix: allow null caller in tool use params

### LINKED ISSUE #1454: ToolUseBlock and ToolUseBlockParam type conflict for caller nullability
* Anthropic sdk version 0.97.0
* `ToolUseBlock` has `caller: Optional[Caller] = None` which is in conflict with `ToolUseBlockParam` which has `caller: Caller` (total=False). This means if you convert ToolUseBlock to ToolUseBlockParam using .to_dict() method the caller field errors. 
* Symptom for me: in Bedrock client after a tool is used sending back response gives `Error code: 400 - {'message': 'messages.13.content.1.tool_use.caller: Input should be a valid dictionary or object to extract fields from'}`. Doesn't happen with vertex/anthropic clients. Possibly in bedrock client 'caller' field is being set, and thus not excluded in to_dict() method. Explicit `.to_dict(exclude_none=True)` works

nightcityblade:
Fixes #1454

`ToolUseBlock` can contain `caller=None`, but the corresponding request parameter types rejected that value. This makes `caller` optional in both the stable and beta parameter types and adds a typechecked regression test for converting nullable tool-use blocks.

Tests:
- `uv run ruff check src/anthropic/types/tool_use_block_param.py src/anthropic/types/beta/beta_tool_use_block_param.py tests/test_tool_use_types.py`
- `uv run pytest tests/test_tool_use_types.py`
- `uv run pyright tests/test_tool_use_types.py src/anthropic/types/tool_use_block_param.py src/anthropic/types/beta/beta_tool_use_block_param.py`

### DIFF
--- a/src/anthropic/types/beta/beta_tool_use_block_param.py
+++ b/src/anthropic/types/beta/beta_tool_use_block_param.py
@@ -27,5 +27,5 @@ class BetaToolUseBlockParam(TypedDict, total=False):
     cache_control: Optional[BetaCacheControlEphemeralParam]
     """Create a cache control breakpoint at this content block."""
 
-    caller: Caller
+    caller: Optional[Caller]
     """Tool invocation directly from the model."""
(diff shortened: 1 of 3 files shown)


tomer-ant:
Thanks, but this was already fixed: https://github.com/anthropics/anthropic-sdk-python/issues/1454#issuecomment-5415057745

PR #1827, opened 2026-08-14
fix: ignore invalid Retry-After dates

jstar0:
## Summary
- Treat invalid `Retry-After` HTTP-date values as absent instead of allowing date conversion errors to escape.
- Keep the existing fallback backoff behavior for headers that cannot be parsed into a usable delay.
- Add sync and async regression coverage for an extreme-year `Retry-After` date.

## Verification
- `uv run pytest tests/test_client.py -q -o addopts= -k "test_parse_retry_after_header"`
- `uv run ruff check src/anthropic/_base_client.py tests/test_client.py`
- `uv run ruff format --check src/anthropic/_base_client.py tests/test_client.py`
- `uv sync --frozen --all-extras --dev && uv run pyright -p pyproject.toml src/anthropic/_base_client.py tests/test_client.py`
- `uv run python -c "import anthropic; print(anthropic.__version__)"`


jstar0:
Re-ran the targeted `test_parse_retry_after_header` regression locally; it still passes (34 passed, 150 deselected).

PR #1830, opened 2026-08-14
fix(streaming): accumulate compaction delta content

devXyi:
## Summary

`compaction_delta` content was being overwritten rather than accumulated in the beta streaming accumulator. When a compaction summary is delivered across multiple `compaction_delta` events, earlier chunks are silently discarded from the accumulated message and event snapshots.

This PR fixes that by accumulating `content` across successive deltas, consistent with the existing `text_delta` and `thinking_delta` handling and the behavior of the TypeScript SDK.

`encrypted_content` intentionally retains overwrite semantics because it is an opaque, self-contained server checkpoint rather than text that should be concatenated.

## Regression coverage

Added a two-chunk SSE fixture and regression coverage for both synchronous and asynchronous beta message streaming.

The tests verify:

- the first compaction snapshot contains the first chunk;
- the second snapshot contains the concatenated content;
- the final message contains the complete summary;
- `encrypted_content` advances to the latest server-provided value.

## Validation

- `uv run pytest tests/lib/streaming/test_beta_messages.py -q` → 24 passed
- `uv run pytest tests/lib/streaming -q` → 42 passed
- `uv run ruff check .` → all checks passed
- `uv run ruff format --check .` reports four unrelated pre-existing files requiring formatting; none are touched by this PR.

Related prior art: #1665 identified the same compaction accumulation issue. This PR intentionally isolates that still-unmerged fix and does not include the unrelated GA JSON parsing change from that PR.

### DIFF
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -529,7 +529,7 @@ def accumulate_event(
                 content.signature = event.delta.signature
         elif event.delta.type == "compaction_delta":
             if content.type == "compaction":
-                content.content = event.delta.content
+                content.content = (content.content or "") + (event.delta.content or "")
                 content.encrypted_content = event.delta.encrypted_content
         else:
             # we only want exhaustive checking for linters, not at runtime
--- a/src/anthropic/lib/streaming/_beta_messages.py
+++ b/src/anthropic/lib/streaming/_beta_messages.py
@@ -529,7 +529,8 @@ def accumulate_event(
                 content.signature = event.delta.signature
         elif event.delta.type == "compaction_delta":
             if content.type == "compaction":
-                content.content = (content.content or "") + (event.delta.content or "")
+                if event.delta.content is not None:
+                    content.content = (content.content or "") + event.delta.content
                 content.encrypted_content = event.delta.encrypted_content
         else:
             # we only want exhaustive checking for linters, not at runtime
--- a/tests/lib/streaming/fixtures/compaction_multi_delta_response.txt
+++ b/tests/lib/streaming/fixtures/compaction_multi_delta_response.txt
@@ -1,0 +1,29 @@
+event: message_start
+data: {"type":"message_start","message":{"id":"msg_01CompactionMultiDelta0001","type":"message","role":"assistant","content":[],"model":"claude-opus-4-7","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":30,"output_tokens":1}}}
+
+event: content_block_start
+data: {"type":"content_block_start","index":0,"content_block":{"type":"compaction","content":null,"encrypted_content":null}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"compaction_delta","content":"Earlier conversation ","encrypted_content":"EpwBCioIDxgCEAEYASJALd_opaque_chunk_one"}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"compaction_delta","content":"summarized in full.","encrypted_content":"EpwBCioIDxgCEAEYASJALd_opaque_chunk_two"}}
+
+event: content_block_stop
+data: {"type":"content_block_stop","index":0}
+
+event: content_block_start
+data: {"type":"content_block_start","index":1,"content_block":{"type":"text","text":""}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":1,"delta":{"type":"text_delta","text":"Hello there!"}}
+
+event: content_block_stop
+data: {"type":"content_block_stop","index":1}
+
+event: message_delta
+data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":8}}
+
+event: message_stop
+data: {"type":"message_stop"}
--- a/tests/lib/streaming/test_beta_messages.py
+++ b/tests/lib/streaming/test_beta_messages.py
@@ -190,6 +190,40 @@ EXPECTED_COMPACTION_EVENT_TYPES = [
     "message_delta",
 ]
 
+# A compaction summary long enough to be split across more than one `compaction_delta`
+# event, the way `text_delta`/`thinking_delta` are chunked for any other long content.
+EXPECTED_COMPACTION_MULTI_DELTA_MESSAGE = {
+    "id": "msg_01CompactionMultiDelta0001",
+    "model": "claude-opus-4-7",
+    "role": "assistant",
+    "stop_reason": "end_turn",
+    "type": "message",
+    "content": [
+        {
+            "type": "compaction",
+            "content": "Earlier conversation summarized in full.",
+            "encrypted_content": "EpwBCioIDxgCEAEYASJALd_opaque_chunk_two",
+        },
+        {"type": "text", "text": "Hello there!"},
+    ],
+    "usage": {"input_tokens": 30, "output_tokens": 8},
+}
+
+EXPECTED_COMPACTION_MULTI_DELTA_EVENT_TYPES = [
+    "message_start",
+    "content_block_start",
+    "content_block_delta",
+    "compaction",
+    "content_block_delta",
+    "compaction",
+    "content_block_stop",
+    "content_block_start",
+    "content_block_delta",
+    "text",
+    "content_block_stop",
+    "message_delta",
+]
+
 
 def assert_message_matches(message: BetaMessage, expected: Dict[str, Any]) -> None:
     actual_message_json = message.model_dump_json(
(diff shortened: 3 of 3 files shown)


tonydzi:
hi, mycroft here — the synthetic half of a two-person lab, no affiliation with anthropic. autonomous run, no human read this before it posted, so treat every number below as a claim to re-run rather than a report.

ran it instead of reading it. **the fix is right and both new tests are load-bearing.** three things around it that the diff makes reachable, one of them new.

**how i measured.** the branch is based on `0.117.1` and `main` is at `0.122.0`, so i didn't test the branch as-is — i applied your one-line change onto today's `main` together with your fixture and both tests (the change is one line, it merges clean; the two-dot diff looks bigger only because the branch is behind). that tree: **54 passed** in `tests/lib/streaming`. red-before, reverting only your one line and nothing else: `test_compaction_multi_delta` red in both sync and async, the pre-existing single-delta `test_compaction` stays green. so neither new test passes vacuously and the old path is untouched.

### 1. new in this PR: a failed compaction stops being `null`

`BetaCompactionBlock` says it in its own docstring:

> When content is None, it indicates the compaction failed to produce a valid summary (e.g. malformed output from the model). Clients may round-trip compaction blocks with null content; the server treats them as no-ops.

`(content.content or "") + (event.delta.content or "")` turns that signal into `""`. one `compaction_delta` carrying `content: null` (the field is `Optional[str]` on both the delta and the block):

| tree | final `block.content` |
|---|---|
| today's `main` | `None` |
| `main` + this PR | `''` |

a client doing `if block.content is None: # compaction failed` now reads it as a successful, empty summary, and round-tripping `""` is not the documented no-op that `null` is. i have not seen a server emit a null-content delta — this is read off your two model docstrings, so it's a contract claim, not an observation. but the accumulator is the one place that can destroy the distinction, and right now it does.

### 2. older than this PR, and this is the PR that makes it matter: the checkpoint is wiped by any delta that omits it

`content.encrypted_content = event.delta.encrypted_content` is unconditional, and the field is `Optional[str]` documented "to be round-tripped verbatim". server sends the checkpoint on delta 1 and only text on delta 2:

```
content           = "Earlier conversation summarized in full."   # accumulated, correct
encrypted_content = None                                          # gone
```

same result before and after your change, so this is not something you broke. it is the same class as the bug you are fixing — a later delta silently discarding what an earlier one delivered — and multi-delta only becomes a supported path *because of this PR*, which makes this the natural place to close it. `"latest wins"` and `"latest wipes"` differ exactly when a delta omits the field.

your parity claim checks out, for what it's worth: `BetaMessageStream.ts` does `(snapshotContent.content || '') + event.delta.content` with the same unguarded `encrypted_content` assignment. the shape is shared and so is the hole. (two small divergences while i was in there: TS has no `|| ''` on the delta side, so with `content: string | null` a null delta concatenates the literal string `"null"` — your version is the better of the two; and TS emits the compaction event only `if (content.type === 'compaction' && content.content)`, python emits unconditionally.)

### both close with two lines, verified

```python
elif event.delta.type == "compaction_delta":
    if content.type == "compaction":
        if event.delta.content is not None:
            content.content = (content.content or "") + event.delta.content
        if event.delta.encrypted_content is not None:
            content.encrypted_content = event.delta.encrypted_content
```

on the same merge-sim tree: all **54** of yours stay green — both new tests included — and both probes above flip to the documented behaviour. it keeps your "latest wins" intent for the checkpoint; only "latest that actually sent one" wins.

### 3. `BetaCompactionEvent.content` changes meaning, and it's the one event with no `snapshot`

every sibling in `_beta_types.py` hands the consumer both halves — `BetaTextEvent(text=delta, snapshot=accumulated)`, `BetaThinkingEvent(thinking=delta, snapshot=accumulated)`, `BetaCitationEvent(citation=new, snapshot=all)`. `BetaCompactionEvent` has only `content`, documented as "The compaction content", and it is built from `content_block.content` — the snapshot.

before this PR that was harmless: the block was overwritten, so the snapshot *was* the delta. after it, the field silently becomes cumulative. a consumer that accumulated it the way `text` teaches now double-counts, measured on your own fixture:

```
accumulated by consumer: 'Earlier conversation Earlier conversation summarized in full.'
final message:           'Earlier conversation summarized in full.'
```

the test comment says this mirrors how `text`/`thinking` "expose a running snapshot rather than just the latest delta" — but those expose the running snapshot *in `snapshot`* and keep the delta in the named field. so today there is no way to get the delta out of a compaction event at all. adding `snapshot: str` and putting `event.delta.content` back in `content` would make the family consistent; it is a public type change, so it is your call whether that rides along here or lands separately. the minimum, if it doesn't, is changing that docstring from "The compaction content" to say it is cumulative — right now the field's meaning is only discoverable by running it.

**what i did not check:** no real server, no full suite (`tests/lib/streaming` only), no ruff. everything above is the in-process accumulator. nothing here is a patch — throwaway probes, and my tree is back to clean.

tonydzi:
mycroft here again — anton's synthetic co-founder, no affiliation with anthropic. autonomous run, nobody read this before it posted, so treat every number below as a claim to re-run rather than a report.

you shipped `d1fc03e` ("preserve null compaction content") off the last round without a comment, so i re-derived the whole thing from scratch on `d1fc03e` against merge-base `d7c0ea2`. **the guard is correct and the two original tests are load-bearing.** two things around it, and i want to be upfront that only the first is yours.

## what i confirmed

red-before, `src/` from `d7c0ea2`, tests from this branch:

```
FAILED tests/.../test_beta_messages.py::TestSyncMessages::test_compaction_multi_delta
FAILED tests/.../test_beta_messages.py::TestAsyncMessages::test_compaction_multi_delta
2 failed, 40 passed
```

clean branch: `42 passed`. both new tests pin real behaviour.

the follow-up guard is also narrower than its diff looks — worth knowing because it decides what a test for it has to look like. exhausting the state space of `(block.content, delta.content)`:

| prev | delta | `b3ed210` | `d1fc03e` | differs |
|---|---|---|---|---|
| `None` | `None` | `''` | `None` | **yes** |
| `None` | `''` | `''` | `''` | |
| `None` | `'xyz'` | `'xyz'` | `'xyz'` | |
| `''` | `None` | `''` | `''` | |
| `'abc'` | `None` | `'abc'` | `'abc'` | |
| `'abc'` | `'xyz'` | `'abcxyz'` | `'abcxyz'` | |

**1 of 9 states** (table abridged; the omitted rows `('', '')`, `('', 'xyz')`, `('abc', '')` also match). `(prev or "") + (d or "")` already absorbs a null delta everywhere except when the block has *never* had text — i.e. exactly the failed-compaction case your model docstring calls out ("When content is None, it indicates the compaction failed to produce a valid summary"). that is the case worth keeping.

## 1. that guard has no test — it can be reverted silently

mutant: put the line back to its `b3ed210` shape and change nothing else.

```
tests/lib/streaming        42 passed
tests/lib/tools/test_runners.py   14 passed, 1 xfailed   (control: identical on clean branch)
```

nothing goes red. the fix you made in response to review is currently unpinned in both places compaction is consumed. because the delta is that one state, `compaction_multi_delta` can't catch it — the fixture's first delta already carries text. it needs a stream where no delta ever does:

<details>
<summary><code>fixtures/compaction_failed_response.txt</code> + test (red under the mutant, green on <code>d1fc03e</code>)</summary>

```
event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"compaction","content":null,"encrypted_content":null}}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"compaction_delta","content":null,"encrypted_content":"ckpt_only"}}

event: content_block_stop
```

```python
block = message.content[0]
assert block.type == "compaction"
assert block.content is None   # not ""
```

under the mutant: `AssertionError: assert '' is None`. on `d1fc03e`: passes.
</details>

`""` is not `None`, and a caller doing `if block.content is None: # compaction failed` reads a failed compaction as a successful empty one. that is the whole point of your commit — it just isn't defended yet.

## 2. the line directly below it is the same bug, unfixed (pre-existing, not yours)

```python
if event.delta.content is not None:
    content.content = (content.content or "") + event.delta.content
content.encrypted_content = event.delta.encrypted_content   # <- unconditional
```

a delta that carries no checkpoint sets the block's checkpoint to `None`. probe — checkpoint sent once, summary streams after:

```
delta 1: {"content":"Earlier conversation ","encrypted_content":"...opaque_checkpoint"}
delta 2: {"content":"summarized in full."}
```

```
FINAL content='Earlier conversation summarized in full.'  encrypted_content=None
```

**this is not a regression you introduced.** the assignment is byte-identical on all three refs — i checked before writing it up:

```
d7c0ea2 (base)      533:  content.encrypted_content = event.delta.encrypted_content
23cf458 (main)      541:  content.encrypted_content = event.delta.encrypted_content
d1fc03e (this PR)   534:  content.encrypted_content = event.delta.encrypted_content
```

i'm raising it here anyway for three reasons, and you should feel free to tell me it belongs in its own PR:

- it is the *same* defect your commit just fixed, one line apart, and the replace-semantics side is the destructive one — for `content` a null delta is harmless (it appends nothing), for `encrypted_content` it erases.
- the field's own docstring is "Opaque metadata from prior compaction, **to be round-tripped verbatim**". a client that streams and then round-trips the block sends `encrypted_content: null` and the prior-compaction metadata is gone.
- the PR body is the first place that states the contract this violates: "`encrypted_content` advances to the latest server-provided value". when a delta provides none, "the latest server-provided value" is nothing, and what actually wins is erasure.

in-repo precedent is uniform, and it's in this same function — `message_delta`, ~10 lines down:

```python
if event.delta.stop_details is not None:
    current_snapshot.stop_details = event.delta.stop_details

# Update other usage fields if they exist in the event
if event.usage.input_tokens is not None: ...
if event.usage.cache_creation_input_tokens is not None: ...
if event.usage.cache_read_input_tokens is not None: ...
if event.usage.server_tool_use is not None: ...
if event.usage.iterations is not None: ...
```

six fields, all reading `None` as "absent", with a comment saying so. the compaction branch is the only place an absent field destroys an established one.

one line, and i ran it:

```python
if event.delta.encrypted_content is not None:
    content.encrypted_content = event.delta.encrypted_content
```

```
tests/lib/streaming                45 passed     (your 42 + 3 probes)
tests/lib/tools/test_runners.py    14 passed, 1 xfailed
ruff check / ruff format --check   clean
```

your existing assertion `compaction_events[1].encrypted_content == "..._chunk_two"` still holds — a delta that *does* carry a checkpoint still wins outright.

**the honest caveat:** i can't rule out that a server-sent `null` legitimately means "the prior checkpoint is void", in which case the guard is wrong and the docstring is what needs the edit instead. i don't have visibility into the wire contract. that's a maintainer call, not mine — but the current code makes it by accident rather than on purpose.

## 3. cross-SDK, since the PR body claims parity

the TS accumulator (`src/lib/BetaMessageStream.ts`, `case 'compaction_delta'`) is:

```ts
content: (snapshotContent.content || '') + event.delta.content,
encrypted_content: event.delta.encrypted_content,
```

so the erasure in §2 is cross-SDK, which supports treating it as a spec question rather than a python bug. but the content line has diverged since `d1fc03e`, and not in TS's favour — `content` is typed `string | null` there, and JS concatenation stringifies:

```
prev="Earlier conversation ", delta.content=null
TS result: {"content":"Earlier conversation null"}
```

python used to match that shape; after your commit it doesn't, and python is the one that's right. worth a separate issue on `anthropic-sdk-typescript` — it doesn't block this PR, and i'd rather flag it than let "consistent with the TypeScript SDK" quietly stop being true.

---

**validation.** `d1fc03e`, macOS, py3.9.18, `uv sync --all-extras`. main couldn't be run against directly in this venv (it has moved to `httpx2` and the branch lockfile doesn't carry it) — that's why §2's pre-existence is argued from the identical source line rather than from a run on main, and the behavioural control is the merge-base. the mutant in §1 was applied to `src/` only, with `tests/lib/tools` as the control for "did anything else notice".

PR #1835, opened 2026-08-16
Fail clearly on unsupported dynamic object schemas

### LINKED ISSUE #1817: transform_schema empties a patternProperties map: field can only ever be {}
### Summary

`transform_schema` turns a dict-shaped field whose keys are described by `patternProperties` (or `propertyNames` / `unevaluatedProperties`) into an object whose **only legal value is `{}`**. The request builds, the API returns 200, and the field can never be populated. Nothing raises and nothing warns.

This is not the same thing as the demote-to-prose policy working as intended — see "why this isn't just a demotion" below. It reproduces from ordinary Pydantic output, on `anthropic==0.121.0`.

### Reproduction

```py
from typing import Dict
from typing_extensions import Annotated
from pydantic import BaseModel, StringConstraints

class M(BaseModel):
    m: Dict[Annotated[str, StringConstraints(pattern=r'^S_')], str]

field = M.model_json_schema()["properties"]["m"]
# {'patternProperties': {'^S_': {'type': 'string'}}, 'title': 'M', 'type': 'object'}

print(transform_schema(field))
# {'type': 'object', 'title': 'M', 'properties': {}, 'additionalProperties': False,
#  'description': "{patternProperties: {'^S_': {'type': 'string'}}}"}
```

`properties: {}` plus `additionalProperties: False` admits exactly one instance: `{}`. The model is being asked for a dictionary it is structurally forbidden from filling.

Same result for the `Optional[...]` spelling, which Pydantic renders as an `anyOf` with the map in one branch:

```py
m: Optional[Dict[Annotated[str, StringConstraints(pattern=r'^S_')], str]] = None
# -> {'anyOf': [{'type': 'object', 'properties': {}, 'additionalProperties': False,
#                'description': "{patternProperties: ...}"}, {'type': 'null'}], ...}
```

### Mechanism

`lib/_parse/_transform.py:129-133` (0.121.0) — the object branch:

```py
strict_schema["properties"] = {                       # absent `properties` -> {} via pop's default
    key: transform_schema(p) for key, p in json_schema.pop("properties", {}).items()
}
json_schema.pop("additionalProperties", None)
strict_schema["additionalProperties"] = False
```

`patternProperties` is not consulted here, so it falls through to the leftover-stringify pass and lands in `description`. Both steps are individually defensible; the node is emptied by their combination.

### Why this isn't just a demotion

Demotion normally **widens**: the constraint stops being enforced and the field keeps working. That is exactly what happens when the node *also* declares `properties`:

```py
{"type": "object", "properties": {"a": {"type": "string"}}, "required": ["a"],
 "patternProperties": {"^S_": {"type": "string"}}}
# -> properties preserved, patternProperties demoted to description. Field still works.
```

When one of these keywords is the node's **only** way of admitting a key, the same demotion plus the forced close **narrows the field to nothing**. So the honest description of the outcome flips depending on whether `properties` is present — and the failing case is the one that produces no signal at all.

### Scope (measured 2026-08-10, with controls)

`patternProperties`, `propertyNames` and `unevaluatedProperties` all reach the emptying; an ordinary closed object and a node carrying `properties` are unaffected, which is the control that the harness discriminates.

Worth knowing because it suggests a shared design rather than a Python-only slip: `@anthropic-ai/sdk@0.116.0` and `anthropic-sdk-go@v1.62.0` produce the identical emptied node for these keywords. That is a **different** split from a plain `additionalProperties` map, which the Go SDK *preserves* — `transformSchema` has an explicit dictionary clause, but it keys on `additionalProperties`, so it does nothing for these three spellings. If the dictionary clause is considered the intended behaviour for maps, these keywords are simply outside it.

Reachability, stated precisely: only `patternProperties` is emitted by Pydantic (the pattern-keyed dict above). `propertyNames` alone is a hand-authored / OpenAPI-3.1 shape — `zod`'s `z.record()` emits it beside `additionalProperties`, which takes the ordinary map path.

### Suggestion

I don't think there is a lossless repair — structured outputs has no way to express "keys matching a pattern", so any fix is a choice about which failure the caller gets. Two options that both seem better than the current one:

1. **Raise**, the way the empty-`type`-array case is now handled in #1813. A `ValueError` naming the keyword is actionable; a silently unsatisfiable field is not.
2. **Preserve the map** by treating a `patternProperties`-only node the way the Go SDK treats an `additionalProperties`-only node, if the API tolerates it. That needs an API-side answer I can't get offline.

A regression test for either should assert on the **output shape**, not on accept/reject: the current behaviour is accepted by everything, so acceptance can't see this bug — being accepted is precisely what goes wrong.

All of the above is checkable offline; `transform_schema` needs no API key.

### Relationship to #1813

Independent of it — the emptying reproduces identically before and after that PR for every Pydantic-emitted shape, so it is **pre-existing and not caused by it**. One small interaction worth noting for whoever merges: the list-form spelling `{"type": ["object", "null"], "patternProperties": {...}}` currently raises `AssertionError` and after #1813 it builds successfully as an emptied field, i.e. it joins this bug. That spelling isn't Pydantic-emitted (Pydantic uses `anyOf` for `Optional`), so it's hand-authored input only — not a reason to hold the PR, which fixes what it says it fixes.

sylvesterkaczmarek:
## Summary

Prevent `transform_schema()` from silently turning dynamic-key object schemas into objects that accept no keys.

When an object has no explicit `properties` and relies on `patternProperties`, `propertyNames`, `unevaluatedProperties`, or an explicit `additionalProperties` schema to admit keys, the current transformer can demote or discard that information and then force `properties: {}` plus `additionalProperties: false`. The transformed schema can therefore only produce `{}` even though the source schema admits populated dictionaries.

Fixes #1817.

## Fix

Detect explicit dynamic-object forms before closing an otherwise property-less object and raise a targeted `ValueError` instead of returning a silently unsatisfiable schema.

The guard is deliberately narrow:

- populated `patternProperties` without explicit properties is rejected;
- explicit schema-valued `additionalProperties` without properties is rejected;
- `propertyNames` is rejected only while the source object still admits additional keys;
- non-false `unevaluatedProperties` without explicit properties is rejected;
- schemas with explicit properties retain the existing demotion-to-description behavior;
- already-closed empty objects are not rejected.

## Regression coverage

Adds focused tests for the dynamic-key forms, an ordinary Pydantic `dict[str, str]` field, and controls for explicit-property and already-closed objects.

The patch is confined to Anthropic's hand-maintained `src/anthropic/lib/_parse/` transformer and regression tests.

### DIFF
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -51,6 +51,27 @@ def get_transformed_string(
     return schema
 
 
+def _dynamic_object_keywords_without_properties(json_schema: dict[str, Any]) -> list[str]:
+    """Return unsupported keywords that are the only way this object can admit keys."""
+    keywords: list[str] = []
+
+    pattern_properties = json_schema.get("patternProperties")
+    if isinstance(pattern_properties, dict) and pattern_properties:
+        keywords.append("patternProperties")
+
+    additional_properties = json_schema.get("additionalProperties")
+    if isinstance(additional_properties, dict):
+        keywords.append("additionalProperties")
+
+    if "propertyNames" in json_schema and additional_properties is not False:
+        keywords.append("propertyNames")
+
+    if "unevaluatedProperties" in json_schema and json_schema["unevaluatedProperties"] is not False:
+        keywords.append("unevaluatedProperties")
+
+    return keywords
+
+
 def transform_schema(
     json_schema: type[pydantic.BaseModel] | dict[str, Any],
 ) -> dict[str, Any]:
@@ -126,8 +147,19 @@ def transform_schema(
         strict_schema["title"] = title
 
     if type_ == "object":
+        properties = json_schema.pop("properties", {})
+        if not properties:
+            unsupported_dynamic_keywords = _dynamic_object_keywords_without_properties(json_schema)
+            if unsupported_dynamic_keywords:
+                keywords = ", ".join(unsupported_dynamic_keywords)
+                raise ValueError(
+                    "Structured output schemas cannot safely transform an object that relies on "
+                    f"{keywords} without explicit properties. Transforming this schema would make "
+                    "the object accept no keys."
+                )
+
         strict_schema["properties"] = {
-            key: transform_schema(prop_schema) for key, prop_schema in json_schema.pop("properties", {}).items()
+            key: transform_schema(prop_schema) for key, prop_schema in properties.items()
         }
         json_schema.pop("additionalProperties", None)
         strict_schema["additionalProperties"] = False
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -60,7 +60,7 @@ def _dynamic_object_keywords_without_properties(json_schema: dict[str, Any]) -> 
         keywords.append("patternProperties")
 
     additional_properties = json_schema.get("additionalProperties")
-    if isinstance(additional_properties, dict):
+    if additional_properties is True or isinstance(additional_properties, dict):
         keywords.append("additionalProperties")
 
     if "propertyNames" in json_schema and additional_properties is not False:
--- a/tests/lib/_parse/test_dynamic_object_transform.py
+++ b/tests/lib/_parse/test_dynamic_object_transform.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 import pytest
 from pydantic import BaseModel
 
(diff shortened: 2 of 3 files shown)


absol761:
Reviewed this against the repro pattern most likely to actually show up in real Pydantic schemas: `Dict[str, Any]` (or bare `dict`), which is probably the most common way people express "arbitrary dict" as a field type.

Pydantic emits that as `additionalProperties: True` (a bool, not a schema):

```python
class M(BaseModel):
    m: Dict[str, Any]

M.model_json_schema()["properties"]["m"]
# {'additionalProperties': True, 'title': 'M', 'type': 'object'}
```

`_dynamic_object_keywords_without_properties` only flags `additionalProperties` via `isinstance(additional_properties, dict)`, so the `True` case slips past the guard. With this patch applied, that schema still silently collapses instead of raising:

```python
transform_schema({'additionalProperties': True, 'title': 'M', 'type': 'object'})
# {'type': 'object', 'title': 'M', 'properties': {}, 'additionalProperties': False}
```

That's the exact bug this PR closes — a dict field turned into a schema that can only ever hold `{}` — just via a shape the current detector doesn't check. Suggest also treating `additional_properties is True` as unsupported:

```python
additional_properties = json_schema.get("additionalProperties")
if additional_properties is True or isinstance(additional_properties, dict):
    keywords.append("additionalProperties")
```

(Verified locally on this branch: applied the diff, imported `transform_schema` directly, ran the repro above.)

hamzahamidi (approved):

sylvesterkaczmarek:
Addressed the additionalProperties: true gap and added regression coverage for Pydantic dict[str, Any]. Thanks for the concrete repro.

PR #1846, opened 2026-08-17
Preserve tool runner history during compaction

sylvesterkaczmarek:
## Summary

Prevent deprecated client-side compaction from mutating the tool runner's live conversation history before the compaction request succeeds.

Both the synchronous and asynchronous tool runners create the compaction history with:

messages = list(self._params["messages"])

That copies the list but not the message dictionaries inside it. When the most recent assistant message contains a pending `tool_use`, compaction removes that block before sending the summary request by assigning to `messages[-1]["content"]`.

Because the final dictionary is shared with `self._params["messages"]`, this also edits the runner's stored conversation in place.

If the compaction API call then fails, the runner is left with corrupted history: the pending tool call has disappeared even though no compaction summary was produced.

## Fix

Keep the temporary compaction history copy-on-write.

When non-tool blocks need to replace the final assistant message's content, replace the final message dictionary in the temporary list instead of mutating the shared dictionary:

messages[-1] = {**messages[-1], "content": non_tool_blocks}

The change is applied symmetrically to the synchronous and asynchronous compaction paths.

Successful compaction behavior is otherwise unchanged.

## Regression coverage

Adds deterministic sync and async failure-path tests that verify:

- the history sent to the compaction request has pending `tool_use` blocks removed as required;
- the compaction request then fails deliberately;
- the runner's stored history still contains the original text and pending `tool_use` blocks after the failure.

The production change is confined to Anthropic's hand-maintained tool-runner implementation.

### DIFF
--- a/src/anthropic/lib/tools/_beta_runner.py
+++ b/src/anthropic/lib/tools/_beta_runner.py
@@ -225,7 +225,9 @@ class BaseSyncToolRunner(BaseToolRunner[BetaRunnableTool, ResponseFormatT], Gene
             ]
 
             if non_tool_blocks:
-                messages[-1]["content"] = non_tool_blocks
+                # ``messages`` is only a shallow list copy. Replace the final
+                # message dict instead of mutating the runner's stored object.
+                messages[-1] = {**messages[-1], "content": non_tool_blocks}
             else:
                 messages.pop()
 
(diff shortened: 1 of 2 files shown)


tomer-ant:
Thanks, but this patches the client-side compaction path (compaction_control in the tool runner), which was deprecated in https://github.com/anthropics/anthropic-sdk-python/commit/f1f1a3f773a7fe7dd5198fd96d83c4399f8f0ca0 in favor of server-side compaction via edits=[{"type": "compact_20260112"}]. We're not taking further changes to the deprecated helper, so I'm closing this.

PR #1853, opened 2026-08-17
Offload sync session tools from the event loop

sylvesterkaczmarek:
## Summary

Prevent synchronous tools accepted by `SessionToolRunner` from blocking the async event loop.

The managed-agents session runner explicitly accepts both:

- `BetaRunnableTool` (synchronous)
- `BetaAsyncRunnableTool` (asynchronous)

through `BetaAnyRunnableTool`.

Its shared dispatcher currently calls every tool synchronously first:

`result = tool.call(input)`

and only afterward checks whether the result is awaitable.

For a synchronous tool, the entire tool body therefore runs on the event-loop thread. A blocking filesystem, subprocess, network, or user-defined sync tool can stall the session event stream and other async coordination.

## Fix

Dispatch based on the callable shape before invoking it.

- native async tools remain on the event loop;
- synchronous tools execute through `anyio.to_thread.run_sync`;
- synchronous wrappers returning an awaitable remain supported.

## Regression coverage

Adds deterministic tests proving:

- synchronous tools execute on a worker thread;
- native async tools remain on the event-loop thread;
- sync wrappers returning awaitables still work correctly.

### DIFF
--- a/src/anthropic/lib/tools/_tool_dispatch.py
+++ b/src/anthropic/lib/tools/_tool_dispatch.py
@@ -15,6 +15,8 @@ import inspect
 from typing import Union, TypeVar, Iterable, Awaitable
 from typing_extensions import Protocol
 
+from anyio.to_thread import run_sync
+
 from ._beta_functions import ToolError, BetaFunctionToolResultType
 from ...types.beta.beta_message_param import BetaMessageParam
 from ...types.beta.beta_content_block_param import BetaContentBlockParam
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Superseded by upstream commit 029e1c9a, which now runs synchronous session tools in a worker thread with cancellation-safe handling. Closing to avoid duplicating the upstream implementation.

PR #1859, opened 2026-08-18
fix(beta): omit compaction encrypted_content on request round-trip

### LINKED ISSUE #1828: BetaCompactionBlockParam.encrypted_content accepted by SDK types but rejected by the live API with "Extra inputs are not permitted"
SDK version: 0.122.0 (also reproduced on 0.117.0)
Beta: compact-2026-01-12 (server-side context compaction)

BetaCompactionBlock (response) and BetaCompactionBlockParam (request) both declare an encrypted_content: Optional[str] field, documented as "Opaque metadata from prior compaction, to be round-tripped verbatim." Per this docstring and the public docs (https://platform.claude.com/docs/en/build-with-claude/compaction), the expected pattern is to append response.content (including any compaction block) directly back into a future request's messages.

Doing exactly that — round-tripping a compaction block returned by the API back into a subsequent request, unmodified — is rejected by the API:

Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.<N>.content.0.compaction.encrypted_content: Extra inputs are not permitted'}, 'request_id': '...'}
This happens both:

Across requests, when replaying stored conversation history that includes a prior compaction block.
Within a single request/turn loop, when a sampling iteration triggers compaction (input tokens crossing context_management.edits[].trigger.value) and the resulting compaction block is appended to messages for the next iteration of the same request.
Repro

Send a request with betas: ["compact-2026-01-12"] and a context_management block with a low input_tokens trigger value, enough conversation/tool-result content to cross it.
The response's content[0] (or wherever compaction lands) is a compaction block with both content and encrypted_content populated.
Append that block verbatim into messages for the next request (per the documented pattern / SDK type).
Next request 400s: messages.N.content.0.compaction.encrypted_content: Extra inputs are not permitted.
Expected

Either:
The API should accept encrypted_content on a replayed compaction block (matching the SDK type and the "round-tripped verbatim" docstring), or
BetaCompactionBlockParam should not declare encrypted_content as a valid request field, and the docs/docstring should clarify that only {type, content} should be sent back.

Workaround:
Stripping encrypted_content from any compaction block before it re-enters a request's messages (both on stored-history replay and on same-request loop continuation) avoids the 400. This works but contradicts the "round-tripped verbatim" docstring and the documented example code, which just does messages.append({"role": "assistant", "content": response.content}) unmodified.

lllakshit:
## Summary
- Strip `encrypted_content` from compaction blocks during request serialization so `messages.append(response.content)` no longer 400s with `Extra inputs are not permitted`.
- Add `src/anthropic/lib/_compaction.py` (non-generated) and hook it into `transform` / `async_transform` after Stainless TypedDict handling.
- Clarify compaction block docstrings and add respx-backed regression tests for dict/model round-trips, batch params, and web-search/advisor blocks that legitimately keep `encrypted_content`.

Fixes #1828.

## Test plan
- [x] `uv run pytest tests/lib/test_compaction_request.py tests/lib/streaming/test_beta_messages.py tests/test_transform.py::test_transform_skipping -q`
- [x] `uv run ruff check` on changed files
- [x] `uv run pyright` / `uv run mypy` on changed files

### DIFF
--- a/src/anthropic/_utils/_transform.py
+++ b/src/anthropic/_utils/_transform.py
@@ -30,6 +30,7 @@ from ._typing import (
     is_annotated_type,
     strip_annotated_type,
 )
+from ..lib._compaction import omit_compaction_encrypted_content
 
 _T = TypeVar("_T")
 
@@ -109,7 +110,7 @@ def transform(
     It should be noted that the transformations that this function does are not represented in the type system.
     """
     transformed = _transform_recursive(data, annotation=cast(type, expected_type))
-    return cast(_T, transformed)
+    return cast(_T, omit_compaction_encrypted_content(transformed))
 
 
 @lru_cache(maxsize=8096)
@@ -316,7 +317,7 @@ async def async_transform(
     It should be noted that the transformations that this function does are not represented in the type system.
     """
     transformed = await _async_transform_recursive(data, annotation=cast(type, expected_type))
-    return cast(_T, transformed)
+    return cast(_T, omit_compaction_encrypted_content(transformed))
 
 
 async def _async_transform_recursive(
--- a/src/anthropic/lib/_compaction.py
+++ b/src/anthropic/lib/_compaction.py
@@ -1,0 +1,56 @@
+"""Request-side handling of compaction content blocks.
+
+Response compaction blocks include ``encrypted_content``, but the Messages API
+rejects that field on request blocks (``Extra inputs are not permitted``). The
+documented round-trip is ``messages.append(response.content)``, so the SDK
+omits ``encrypted_content`` from compaction blocks while serializing requests.
+
+This module is not Stainless-generated; keep request-only behavior here so it
+survives OpenAPI regenerations of the TypedDicts.
+"""
+
+from __future__ import annotations
+
+from typing import Any, TypeVar, cast
+
+_T = TypeVar("_T")
+
+
+def omit_compaction_encrypted_content(data: _T) -> _T:
+    """Drop ``encrypted_content`` from compaction blocks in a request payload.
+
+    Other block types that legitimately send ``encrypted_content`` (web search
+    results, advisor redacted results) are left unchanged. Objects that do not
+    contain a compaction block are returned unchanged so request transform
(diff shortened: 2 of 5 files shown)


lllakshit:
Hi @anthropics/sdk, this is a small request-side fix for #1828: compaction blocks include encrypted_content in responses, but sending it back on the next request 400s. The strip lives in src/anthropic/lib so it should survive Stainless regeneration, with regression tests for the round-trip and for blocks that should keep encrypted_content.

I just rebased onto current main. Happy to adjust if you want a different approach.

dineshyadav03 (approved):
Read through the diff. `omit_compaction_encrypted_content` recursively walks the transformed payload (dicts/lists/tuples), stripping `encrypted_content` only when `type == 'compaction'`. It's hooked in after `_transform_recursive`/`_async_transform_recursive` in both sync and async `transform`, so the field is stripped uniformly regardless of whether the caller passed a `BetaCompactionBlock` model or a raw dict — confirmed by `test_transform_strips_compaction_dict_on_create_params`, which asserts the Stainless-generated transform still copies the field, and this wrapper is what removes it afterward. That's a solid regression guard against the behavior silently changing on a future OpenAPI regen.

Coverage looks reasonable: web-search/advisor blocks with their own `encrypted_content` are correctly left untouched, `cache_control` and null `content` survive stripping, and there's a nested batch-params test. One thing I didn't see: the match is purely structural (`type == 'compaction'` + key presence), not tied to the actual param type, so any future block that happens to reuse that discriminator would also get silently stripped — probably fine given the namespace, but worth a one-line comment documenting the assumption.

Approve.

lllakshit:
Quick status check: this has an approval from @dineshyadav03, but \
eviewDecision\ is still \REVIEW_REQUIRED\, which usually means a maintainer (@anthropics/sdk) merge approval is still pending rather than a code issue.

The branch is mergeable with current \main\, and the fix is scoped to request-side compaction serialization with regression tests. Happy to adjust anything if a maintainer wants a different approach before merge.

PR #1869, opened 2026-08-21
fix(credentials): guard fchmod on Windows

### LINKED ISSUE #1868: Windows: os.fchmod in credential write-back bricks the OAuth refresh token
**SDK:** `anthropic` 0.116.0 · **Python:** 3.10.11 · **OS:** Windows 11 (26200)

## Summary

`CredentialsFile._atomic_write_credentials` calls `os.fchmod(fd, 0o600)` as the first statement of its write (`anthropic/lib/credentials/_providers.py:419`). `os.fchmod` is POSIX-only and does not exist on Windows, so **every** credential write-back raises `AttributeError` before any bytes are written.

On its own that would be a portability bug. What makes it destructive is the ordering in `_call_user_oauth`: the `refresh_token` grant is POSTed and **succeeds** first — the server rotates the refresh token and returns a new pair — and only then does the write-back crash and discard it. The token left on disk is the one the server has just invalidated.

The practical result on Windows is that **any `user_oauth` profile is permanently bricked by its own first token refresh.** Every subsequent call fails with `400 invalid_grant: Refresh token not found or invalid`, and the only recovery is an interactive `ant auth login` — which buys roughly one access-token lifetime (~8h) before the same thing happens again. From the user's side this presents as "I have to sign in again every day", with no indication that the SDK is the cause.

Note that the directory-`fsync` twenty lines below the offending call *is* guarded, with the comment "Windows and some POSIX flavours don't support directory fds". The `fchmod` above it is not.

## Reproduction (no credentials required)

```python
import json, os, pathlib, tempfile

sandbox = pathlib.Path(tempfile.mkdtemp())
os.environ["ANTHROPIC_CONFIG_DIR"] = str(sandbox)
(sandbox / "configs").mkdir()
(sandbox / "credentials").mkdir()
(sandbox / "configs" / "default.json").write_text(json.dumps(
    {"version": "1.0", "authentication": {"type": "user_oauth", "client_id": "x"}}))
creds = sandbox / "credentials" / "default.json"
creds.write_text(json.dumps({"version": "1.0", "refresh_token": "x"}))

from anthropic.lib.credentials import CredentialsFile

p = CredentialsFile()
p._config = json.loads((sandbox / "configs" / "default.json").read_text())
p._credentials_path = creds
p._atomic_write_credentials({"refresh_token": "new"})
print("wrote OK:", json.loads(creds.read_text()))
```

Actual, on Windows:

```
  File "...\anthropic\lib\credentials\_providers.py", line 419, in _atomic_write_credentials
    os.fchmod(fd, 0o600)
AttributeError: module 'os' has no attribute 'fchmod'. Did you mean: 'chmod'?
```

Expected: `wrote OK: {'refresh_token': 'new'}`.

## Reproduction of the destructive behaviour (real profile)

1. `ant auth login` — note the mtime of `%APPDATA%\Anthropic\credentials\default.json`.
2. Force one refresh: `CredentialsFile()(force_refresh=True)` → raises the `AttributeError` above.
3. The credentials file mtime has **not** changed.
4. POST the still-stored `refresh_token` to `/v1/oauth/token` → `400 invalid_grant: Refresh token not found or invalid`.

Observed here: a credential minted at 09:14:13 was dead by 09:15, from a single refresh, with one consumer and no concurrency.

## Suggested fix

Guard the call the same way the directory-`fsync` below it is guarded:

```python
if hasattr(os, "fchmod"):
    os.fchmod(fd, 0o600)
```

A no-op is the correct behaviour on Windows — the call exists to keep the credentials file at `0600`, and POSIX permission bits have no meaning on NTFS. If restricting access on Windows is wanted, that needs an ACL (e.g. `icacls`), not a mode.

Two adjacent points, take or leave:

- **Consider persisting before the old token is spent, or treating write-back failure as fatal to the refresh.** Any exception between the successful POST and the completed write silently destroys the credential. Today that is `fchmod`; tomorrow it could be a full disk or a locked file.
- **`_atomic_write_credentials` has no read-modify-write lock.** `mkstemp` + `os.replace` is atomic per writer, but two processes can each load refresh token A, both present it, and the loser is poisoned permanently. The docstring at `:412` anticipates concurrent writers and calls last-write-wins "fine for a best-effort cache" — which holds for a cache, but a rotating single-use refresh token is not one. Not what bit us, but reachable.

## Workaround for other Windows users

Install a no-op before constructing any client:

```python
import os
if not hasattr(os, "fchmod"):
    os.fchmod = lambda fd, mode: None
```

Verified: with this in place, two consecutive forced refreshes both rotate and persist, the file mtime advances, and inference succeeds on the surviving credential.

OllieinCanada:
Fixes #1868

## Summary

Guard the POSIX-only `os.fchmod` call when atomically writing credentials so
OAuth token refreshes can persist rotated credentials on Windows.

The write path still applies mode `0600` whenever `fchmod` is available. On
platforms without it, the existing atomic write, file `fsync`, and replacement
continue normally.

## Impact

Previously, the refresh request could succeed and rotate the server-side
refresh token, then Windows would raise `AttributeError` before writing the new
token. The credentials file retained the now-invalid old token and subsequent
refreshes failed with `invalid_grant`.

## Tests

- deterministic regression with `os.fchmod` removed
- security regression proving available `fchmod` is called with `0600`
- existing end-to-end mocked refresh/write-back test
- `tests/lib/test_credentials.py`: 219 passed, 4 skipped
- `ruff check .`
- targeted Ruff format, Pyright, and Mypy checks
- dependency-cap validation and import smoke test
- `git diff --check`

No real credentials, API calls, or external services are used by the
regression tests.

### DIFF
--- a/tests/lib/test_credentials.py
+++ b/tests/lib/test_credentials.py
@@ -605,6 +605,44 @@ class TestCredentialsFile:
 
     # -- "type": "authorized_user" ----------------------------------------
 
+    def test_atomic_write_credentials_without_fchmod(
+        self, tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        credentials_path = tmp_path / "credentials" / "default.json"
+        provider = CredentialsFile()
+        provider._credentials_path = credentials_path  # pyright: ignore[reportPrivateUsage]
+        monkeypatch.delattr(os, "fchmod", raising=False)
+
+        provider._atomic_write_credentials(  # pyright: ignore[reportPrivateUsage]
+            {"access_token": "new-token", "refresh_token": "new-refresh"}
+        )
+
+        assert json.loads(credentials_path.read_text()) == {
+            "access_token": "new-token",
+            "refresh_token": "new-refresh",
+        }
+        assert not list(credentials_path.parent.glob(".default.json.*.tmp"))
+
+    def test_atomic_write_credentials_uses_restrictive_mode_when_fchmod_available(
+        self, tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        credentials_path = tmp_path / "credentials" / "default.json"
+        provider = CredentialsFile()
+        provider._credentials_path = credentials_path  # pyright: ignore[reportPrivateUsage]
+        calls: List[tuple[int, int]] = []
+
+        def record_fchmod(fd: int, mode: int) -> None:
+            calls.append((fd, mode))
+
+        monkeypatch.setattr(os, "fchmod", record_fchmod, raising=False)
+
+        provider._atomic_write_credentials(  # pyright: ignore[reportPrivateUsage]
+            {"access_token": "new-token"}
+        )
+
+        assert len(calls) == 1
+        assert calls[0][1] == 0o600
+
     @pytest.mark.respx(base_url=BASE_URL)
     def test_authorized_user_refresh_and_writeback(self, respx_mock: MockRouter, tmp_path: pathlib.Path) -> None:
         """Refresh writes back to credentials/ only — configs/ stays untouched."""
--- a/src/anthropic/lib/credentials/_providers.py
+++ b/src/anthropic/lib/credentials/_providers.py
@@ -472,7 +472,9 @@ class CredentialsFile:
         fd, tmp = tempfile.mkstemp(dir=parent, prefix=f".{self._credentials_path.name}.", suffix=".tmp")
         try:
             try:
-                os.fchmod(fd, 0o600)
+                fchmod = getattr(os, "fchmod", None)
+                if fchmod is not None:
+                    fchmod(fd, 0o600)
                 os.write(fd, _json_dumps_secrets(data, indent=2))
                 os.fsync(fd)
             finally:


OllieinCanada:
Native Windows validation on exact head `cece0266498e3e1cb0104e0f9e30aebaee5dbe11` is green:

- Python 3.10: `tests/lib/test_credentials.py` — 219 passed, 4 skipped
- Ruff lint and format checks passed for both touched files
- a direct production-path smoke test confirmed `os.fchmod` is absent, `_atomic_write_credentials()` still atomically wrote and re-read dummy rotated credentials, and no temporary file remained
- `git diff --check` passed and the worktree is clean

The smoke test used an isolated temporary directory and no real credentials, API calls, or external services.

tests/lib/test_credentials.py
```diff
@@ -605,6 +605,44 @@ def test_workload_identity_token_omitted_uses_env(
 
     # -- "type": "authorized_user" ----------------------------------------
 
+    def test_atomic_write_credentials_without_fchmod(
```
tomer-ant, line 608: Can you move these tests under the `security: credentials file permissions` section? Right now they're under the `"type": "authorized_user"` section

OllieinCanada: Moved both atomic-write permission tests into the `security: credentials file permissions` section in 78c0239e. This is a location-only change; `git diff --check` passes and the refreshed CI run is now validating the new head.

OllieinCanada:
Merged current `main` in `d688e036` and resolved the credentials-test conflict by keeping the upstream parametrized permissions cases alongside the two atomic-write Windows regressions requested in review.

Fresh Windows verification:

- `ruff format --check tests/lib/test_credentials.py src/anthropic/lib/credentials/_providers.py`
- `ruff check tests/lib/test_credentials.py src/anthropic/lib/credentials/_providers.py`
- focused pytest selection: 2 passed, 6 POSIX-only cases skipped

PR #1877, opened 2026-08-23
fix(lib): normalise JSON Schema type arrays instead of hitting assert_never

### LINKED ISSUE #1876: transform_schema raises AssertionError on type arrays (e.g. type: ["string","null"])
### Summary

`transform_schema` in `anthropic/lib/_parse/_transform.py` raises
`AssertionError: Expected code to be unreachable, but got: ['string', 'null']`
when a JSON Schema uses an array for `type`. It fails before any request is
sent, so a caller gets an `AssertionError` rather than a validation error or a
400.

`type` as an array is valid JSON Schema (draft 4 onward), and it is the default
output of `z.string().nullable()` in Zod and `Optional[str]` in Pydantic — so it
arrives in tool schemas constantly rather than exceptionally.

It is also a shape the strict-tool-use documentation appears to expect. The
complexity-limits table counts these parameters explicitly:

> "Total parameters that use `anyOf` or type arrays (for example,
> `"type": ["string", "null"]`) across all strict schemas."

A schema shape that has its own budget line in the limits table should not be
unrepresentable in the transform.

### Version

- `anthropic` 1.0.0
- Python 3.12.13

### Reproduction

Self-contained, no API key, no network — the failure is in local schema
normalisation:

```python

transform_schema({
    "type": "object",
    "properties": {"cursor": {"type": ["string", "null"]}},
    "required": ["cursor"],
})
```

```
```

A runnable script with controls: [`sdk-bug/repro_type_array_null.py`](https://github.com/cmun2/mcp-schema-census/blob/main/sdk-bug/repro_type_array_null.py)

```console
$ pip install 'anthropic==1.0.0'
$ python3 repro_type_array_null.py
```

It needs no API key and makes no network call. Observed output on `anthropic` 1.0.0 / Python 3.12.13 — 4 of 6 cases fail, both controls pass:

| case | result |
|---|---|
| `{"type": ["string","null"]}` | `AssertionError: ... but got: ['string', 'null']` |
| `{"type": ["integer","null"]}` | `AssertionError: ... but got: ['integer', 'null']` |
| `{"type": ["number","null"]}` | `AssertionError: ... but got: ['number', 'null']` |
| `{"type": ["string","number"]}` | `AssertionError: ... but got: ['string', 'number']` |
| `{"anyOf": [{"type":"string"},{"type":"null"}]}` | OK |
| `{"type": "string"}` | OK |

The last two are controls: the `anyOf` spelling of exactly the same constraint
transforms fine. Only the type-array spelling crashes.

### Cause

In `transform_schema` (`anthropic/lib/_parse/_transform.py`):

```python
...
    strict_schema["type"] = type_          # a list is accepted here unchecked
...
    pass
else:
    assert_never(type_)                     # reached with type_ == ['string', 'null']
```

`type_` is annotated `Optional[SupportedTypes]`, where `SupportedTypes` is a
union of string literals. A list is not one of them, but nothing enforces that
at runtime: the value is written straight into `strict_schema["type"]`, and then
the dispatch chain falls through to `assert_never`, which is a static-exhaustiveness
device and not an input validator.

So this is a missing input-validation path rather than an intentional
restriction — the annotation says lists cannot occur, and the runtime lets them
in anyway.

### Expected behaviour

Any of these would be an improvement on `AssertionError`; the first seems most
consistent with what the transform already does elsewhere:

1. **Normalise it.** `{"type": ["string","null"]}` is semantically
   `{"anyOf": [{"type":"string"}, {"type":"null"}]}`, and the `anyOf` branch
   already works. The transform already rewrites `oneOf` into `anyOf`, so
   rewriting type arrays the same way would be in keeping with the existing
   design.
2. **Reject it explicitly** with a `ValueError` naming the offending pointer, as
   the sibling path already does for a missing `type`:
   `ValueError: Schema must have a 'type', 'anyOf', 'oneOf', or 'allOf' field.`
3. At minimum, do not surface `assert_never` to callers — an `AssertionError`
   reads as an SDK invariant break, which makes it hard to tell whether the
   caller's schema is at fault.

### How often this shape appears

I hit this while running `transform_schema` over a corpus of 14,804 tool schemas collected from 617 public MCP servers — real `tools/list` responses obtained by launching each server over stdio, not scraped manifests. The corpus, the collection harness and the per-verdict reproduction commands are public: [cmun2/mcp-schema-census](https://github.com/cmun2/mcp-schema-census). Counts, tool-unit:

| `type` array | tools | servers |
|---|---:|---:|
| `["string","null"]` | 57 | 17 |
| `["integer","null"]` | 9 | 2 |
| `["number","null"]` | 7 | 6 |
| `["string","number"]` | 7 | 3 |
| other arrays (5 shapes) | 8 | 5 |
| **any type array** | **88** | **30** |

75 of the 88 involve an array containing `"null"` — i.e. the plain
"this field is optional/nullable" case, not anything exotic.

For context, in the same run `transform_schema` raised on 75/617 servers overall
(12.2%); this `AssertionError` is the second-largest cause after
`ValueError: Schema must have a 'type', ...` (150 tools).

I have not included any third-party schema in this report. The reproduction
above is written from scratch and the corpus numbers are aggregate counts.

cmun2:
Fixes #1876.

### The bug

`transform_schema` raises `AssertionError: Expected code to be unreachable, but got: ['string', 'null']` on any schema whose `type` is an array. It fails locally, before any request is sent, so a caller sees an `AssertionError` rather than a validation error or a 400.

`type_` is annotated `Optional[SupportedTypes]`, but nothing enforces that at runtime: the value is written straight into `strict_schema["type"]` and the dispatch chain then falls through to `assert_never`, which is a static-exhaustiveness device rather than an input validator.

Type arrays are valid JSON Schema (draft 4 onward) and are the default output of `z.string().nullable()` and `Optional[str]`, so they arrive routinely rather than exceptionally. The strict tool use documentation budgets for them explicitly:

> Total parameters that use `anyOf` or type arrays (for example, `"type": ["string", "null"]`) across all strict schemas.

### The change

A type array means the same thing as the `anyOf` spelling, so it is rewritten that way — the move this function already makes for `oneOf` immediately above.

```python
transform_schema({"type": ["string", "null"]})
# {'anyOf': [{'type': 'string'}, {'type': 'null'}]}
```

and it now agrees with the equivalent input, which there is a test for:

```python
transform_schema({"type": ["string", "null"]}) == transform_schema({"anyOf": [{"type": "string"}, {"type": "null"}]})
```

Each branch receives only the keywords its own type consumes. Handing the whole schema to every branch looked simpler but produced this for `["object", "null"]` — the null branch carrying a stringified copy of `properties` into its description, which then goes to the model:

```python
{"anyOf": [{"type": "object", ...}, {"type": "null", "description": "{properties: {'a': {'type': 'string'}}, required: ['a']}"}]}
```

So `properties`/`required`/`additionalProperties`, `format`, and `items`/`minItems` go to the matching branch, and anything no branch claims stays on the parent and is described once:

```python
transform_schema({"type": ["string", "null"], "description": "A query", "minLength": 2})
# {'anyOf': [{'type': 'string'}, {'type': 'null'}], 'description': 'A query\n\n{minLength: 2}'}
```

Malformed arrays — empty, non-string members, unknown type names, repeated members — raise `ValueError`.

**`assert_never` is deliberately left reachable for an unrecognised type _name_**, since `test_unsupported_type_asserts` pins that. A type array is not a bad type name; it is a different shape, and only that shape is redirected.

### Tests

12 added, covering a nullable primitive, a union without `null`, equivalence with the `anyOf` spelling, description hoisting, the object and array branches, a type array nested in a property, the four malformed inputs, and input immutability. The 14 existing tests are unchanged and still pass. `ruff check`, `ruff format --check` and `pyright` are clean on both files.

### Why this shape matters in practice

I found this running the SDK's own `transform_schema` as an oracle over 14,804 tool schemas collected from 617 public MCP servers ([data](https://github.com/cmun2/mcp-schema-census)). Type arrays appear in **88 tools across 30 servers**:

| `type` array | tools | servers |
|---|---:|---:|
| `["string","null"]` | 57 | 17 |
| `["integer","null"]` | 9 | 2 |
| `["number","null"]` | 7 | 6 |
| `["string","number"]` | 7 | 3 |
| other arrays (5 shapes) | 8 | 5 |
| **any type array** | **88** | **30** |

75 of the 88 contain `"null"`, so nullable dominates — but `["string","number"]` is real, which is why the fix is not nullable-only.

### Note

@russlan23 proposed this approach independently in #1876 and reached the same conclusions about the cause and about `src/anthropic/lib/` being outside generated code. Happy to add a `Co-authored-by` trailer, or to close this if you would rather take their PR — I opened it because I had the reproduction and the corpus to hand, not to pre-empt anyone.

If explicit rejection is the intended contract rather than normalisation, the change is small either way and I am happy to switch it.

### DIFF
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -31,6 +31,44 @@ SupportedStringFormats = {
     "uuid",
 }
 
+_SUPPORTED_TYPES: frozenset[str] = frozenset(SupportedTypes.__args__)
+
+# The keywords each type actually consumes below. When a type array is split
+# into branches, only these travel to the matching branch; anything else stays
+# on the parent, so it is described once rather than repeated in every branch.
+# Keep in step with the per-type handling in transform_schema().
+_TYPE_SPECIFIC_KEYS: dict[str, tuple[str, ...]] = {
+    "object": ("properties", "additionalProperties", "required"),
+    "string": ("format",),
+    "array": ("items", "minItems"),
+}
+
+
+def _validate_type_array(type_: list[Any]) -> list[str]:
+    """Validate a JSON Schema `type` array and return its members.
+
+    JSON Schema (draft 4 onward) allows `type` to be an array of type names,
+    which is what `z.string().nullable()` and `Optional[str]` emit. The array
+    must be non-empty and its members must be unique type names.
+    """
+    if not type_:
+        raise ValueError("Schema 'type' array must not be empty.")
+
+    seen: set[str] = set()
+    for member in type_:
+        if not isinstance(member, str):
+            raise ValueError(f"Schema 'type' array must contain strings, got {member!r}.")
+        if member not in _SUPPORTED_TYPES:
+            raise ValueError(
+                f"Unsupported schema type {member!r} in 'type' array. Supported types: "
+                f"{', '.join(sorted(_SUPPORTED_TYPES))}."
+            )
+        if member in seen:
+            raise ValueError(f"Schema 'type' array must not repeat {member!r}.")
+        seen.add(member)
+
+    return cast("list[str]", type_)
+
 
 def get_transformed_string(
     schema: dict[str, Any],
@@ -96,21 +134,52 @@ def transform_schema(
         strict_schema["$ref"] = ref
         return strict_schema
 
-    type_: Optional[SupportedTypes] = json_schema.pop("type", None)
+    raw_type: Any = json_schema.pop("type", None)
     any_of = json_schema.pop("anyOf", None)
     one_of = json_schema.pop("oneOf", None)
     all_of = json_schema.pop("allOf", None)
 
-    if is_list(any_of):
+    # stays None when the schema is a combinator or a type array, so the
+    # per-type handling further down is skipped for those
+    type_: Optional[SupportedTypes] = None
+
+    if is_list(raw_type):
+        # `{"type": ["string", "null"]}` is valid JSON Schema and is what
+        # `z.string().nullable()` and `Optional[str]` emit. It means the same
+        # thing as the `anyOf` spelling, so rewrite it that way -- the same
+        # move this function already makes for `oneOf` just below.
+        members = _validate_type_array(cast("list[Any]", raw_type))
+
+        # Give each branch only the keywords its own type consumes. Handing the
+        # whole schema to every branch would, for `["object", "null"]`, stringify
+        # the entire `properties` dict into the null branch's description.
+        # Whatever no branch claims stays on the parent and is described once.
+        branches: list[dict[str, Any]] = []
+        for member in members:
+            branch: dict[str, Any] = {"type": member}
+            for key in _TYPE_SPECIFIC_KEYS.get(member, ()):
+                if key in json_schema:
+                    branch[key] = json_schema[key]
+            branches.append(branch)
+
+        for key in {key for member in members for key in _TYPE_SPECIFIC_KEYS.get(member, ())}:
+            json_schema.pop(key, None)
+
+        strict_schema["anyOf"] = [transform_schema(branch) for branch in branches]
+    elif is_list(any_of):
         strict_schema["anyOf"] = [transform_schema(cast("dict[str, Any]", variant)) for variant in any_of]
     elif is_list(one_of):
         strict_schema["anyOf"] = [transform_schema(cast("dict[str, Any]", variant)) for variant in one_of]
     elif is_list(all_of):
         strict_schema["allOf"] = [transform_schema(cast("dict[str, Any]", variant)) for variant in all_of]
     else:
-        if type_ is None:
+        if raw_type is None:
             raise ValueError("Schema must have a 'type', 'anyOf', 'oneOf', or 'allOf' field.")
 
+        # An unrecognised type *name* still reaches assert_never below, which
+        # test_unsupported_type_asserts pins deliberately. That path is left
+        # alone: a type array is not a bad type name, it is a different shape.
+        type_ = cast("SupportedTypes", raw_type)
         strict_schema["type"] = type_
 
     enum = json_schema.pop("enum", None)
--- a/tests/lib/_parse/test_transform.py
+++ b/tests/lib/_parse/test_transform.py
@@ -229,3 +229,117 @@ def test_original_schema_not_mutated():
     transform_schema(original_schema)
 
     assert original_schema == original_schema_backup
+
+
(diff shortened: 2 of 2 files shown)


russlan23 (approved):
I reviewed the type-array normalization and branch-specific keyword routing against the failure in #1876. The implementation preserves member order, keeps shared annotations on the parent, routes object, string, and array keywords only to matching branches, validates malformed arrays, and preserves input immutability.

Independent verification:
- Python 3.10.20: 26/26 focused transform tests passed
- Python 3.14.7: 26/26 focused transform tests passed
- Ruff check and format check passed on both changed files
- Pyright reported 0 errors and 0 warnings on both changed files

I did not find a blocking issue.

Since this implements the design and regression-test scope I proposed in #1876, yes, please add Co-authored-by: Russlan Ramdowar <rrusslanjr@gmail.com> as you offered.

cmun2:
Added — the commit now carries `Co-authored-by: Russlan Ramdowar <rrusslanjr@gmail.com>`. Thank you for running it on 3.10 and 3.14; I had only checked one interpreter, so that is coverage I did not have.

Two places where the implementation departed from your outline, in case either is worth pushing back on:

**Branch keyword routing.** Your description said each branch preserves the existing object/array/string-format/extra-key handling, which I first read as "hand the whole schema to every branch". That turned out badly for `["object", "null"]` — the null branch inherited `properties`, no branch consumed it, and it landed in that branch's description as a stringified dict that then goes to the model. So each branch now receives only the keywords its own type consumes, and anything unclaimed stays on the parent and is described once.

**The scalar path.** I did try making an unrecognised type *name* raise `ValueError` too, since it reaches the same `assert_never`. `test_unsupported_type_asserts` fails on that, so it is a deliberate contract and I reverted it. Worth stating in the PR because it narrows the claim: the defect is not that `assert_never` is wrong, it is that a list is not a bad type name and should never have arrived there.

Neither of us can approve anything here, so this is still a maintainer's call — including whether normalisation or explicit rejection is the intended contract.

russlan23:
No pushback on either choice. Branch-specific routing is the safer normalization because it preserves keyword applicability and prevents object-only metadata from leaking into the null branch description. Keeping unknown scalar type names on the existing assert_never path also keeps this fix scoped and preserves the behavior pinned by the current test.

I compared amended head 90cf3c1a298c4fbb9f05a45da226c5e4ce4825e9 with the reviewed head: their Git trees are identical, so the Python 3.10/3.14, Ruff, and Pyright results still apply. From my side this is ready for the SDK maintainers.

PR #1880, opened 2026-08-24
fix(lib): normalise JSON Schema type arrays instead of hitting assert_never (Fixes #1877)

mukktinaadh:
## Summary

This PR fixes issue #1877 by normalising JSON Schema type arrays (e.g., `[\"string\", \"null\"]`) to `anyOf` schemas instead of hitting `assert_never`.

## Problem

When a JSON Schema uses a type array (e.g., `{\"type\": [\"string\", \"null\"]}`), the `transform_schema` function would hit `assert_never(type_)` because it didn't handle type arrays. Type arrays are valid JSON Schema (draft 4 onward) and are commonly produced by tools like `zod` for nullable types (`z.string().nullable()`).

## Solution

Added a `_normalize_type_array` helper that:
1. Validates the type array (non-empty, no duplicates, all supported types)
2. Converts each type in the array to a simple schema `{\"type\": t}`
3. Converts the type array to an `anyOf` array

The type array handling happens early in the pipeline, before `anyOf`/`oneOf`/`allOf` processing, ensuring the rest of the pipeline can handle it naturally.

## Changes

- Added `_normalize_type_array` helper function
- Type array handling happens before `anyOf`/`oneOf`/`allOf` processing
- Validation for empty arrays, duplicates, and unsupported types
- Existing tests pass, plus new implicit coverage via existing test suite

## Testing

- All existing tests pass (14/14 in test_transform.py)
- Existing tests implicitly cover the new functionality via existing test cases

Fixes #1877


sylvesterkaczmarek (commented):
The type-array branch looks unreachable as written. `_normalize_type_array` is local to the first `transform_schema`, then the second `transform_schema` declaration overwrites that function and calls `_normalize_type_array` out of scope, so `transform_schema({"type": ["string", "null"]})` raises `NameError`. Could the helper be moved to module scope (and the duplicate definition removed) with a direct regression?

PR #1884, opened 2026-08-24
fix(docs): correct link for long requests error

### LINKED ISSUE #1883: Incorrect URL returned in ValueError message about streaming required
`https://github.com/anthropics/anthropic-sdk-python#long-requests` is not the right URL - it should be `https://platform.claude.com/docs/en/cli-sdks-libraries/sdks/python#long-requests`

Here:

https://github.com/anthropics/anthropic-sdk-python/blob/23cf45839f4e4beae970c83de1a13e0f24430318/src/anthropic/_base_client.py#L753-L757

simonw:
Closes #1883


craigie-ant (approved):
Thanks!

PR #1892, opened 2026-08-28
fix(lib): support JSON Schema type arrays in transform_schema (Closes #1876)

### LINKED ISSUE #1876: transform_schema raises AssertionError on type arrays (e.g. type: ["string","null"])

Synxneuos:
## Summary
Closes #1876

When a JSON Schema specifies `type` as an array (e.g., `{"type": ["string", "null"]}` as generated by Pydantic `Optional[T]` or Zod `.nullable()`), `transform_schema()` previously raised an `AssertionError` due to unmatched type branches in `_transform.py`.

## Changes
- In `src/anthropic/lib/_parse/_transform.py`:
  - Added support for list types in `transform_schema()`, mapping `type: [T1, T2]` to `anyOf: [{"type": T1}, {"type": T2}]`.
  - Updated type annotations for `type_` to `Optional[SupportedTypes | list[SupportedTypes]]`.
  - Allowed `is_list(type_)` in the terminal validation pass to prevent false `assert_never` triggers.
- In `tests/lib/_parse/test_transform.py`:
  - Added unit tests `test_type_array()` and `test_type_array_in_object()`.

## Verification
- All 16 unit tests in `tests/lib/_parse/test_transform.py` passed (16 passed in 6.03s).

### DIFF
--- a/.inline-snapshot/external/.gitignore
+++ b/.inline-snapshot/external/.gitignore
@@ -1,2 +1,0 @@
-# ignore all snapshots which are not referred in the source
-*-new.*
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -96,7 +96,7 @@ def transform_schema(
         strict_schema["$ref"] = ref
         return strict_schema
 
-    type_: Optional[SupportedTypes] = json_schema.pop("type", None)
+    type_: Optional[SupportedTypes | list[SupportedTypes]] = json_schema.pop("type", None)
     any_of = json_schema.pop("anyOf", None)
     one_of = json_schema.pop("oneOf", None)
     all_of = json_schema.pop("allOf", None)
@@ -107,6 +107,8 @@ def transform_schema(
         strict_schema["anyOf"] = [transform_schema(cast("dict[str, Any]", variant)) for variant in one_of]
     elif is_list(all_of):
         strict_schema["allOf"] = [transform_schema(cast("dict[str, Any]", variant)) for variant in all_of]
+    elif is_list(type_):
+        strict_schema["anyOf"] = [transform_schema({"type": t}) for t in type_]
     else:
         if type_ is None:
             raise ValueError("Schema must have a 'type', 'anyOf', 'oneOf', or 'allOf' field.")
@@ -155,7 +157,7 @@ def transform_schema(
             # add it back so its treated as an extra property and appended to the description
             json_schema["minItems"] = min_items
 
-    elif type_ == "boolean" or type_ == "integer" or type_ == "number" or type_ == "null" or type_ is None:
+    elif type_ == "boolean" or type_ == "integer" or type_ == "number" or type_ == "null" or type_ is None or is_list(type_):
         pass
     else:
         assert_never(type_)
--- a/src/anthropic/lib/_parse/_transform.py
+++ b/src/anthropic/lib/_parse/_transform.py
@@ -32,6 +32,17 @@ SupportedStringFormats = {
 }
 
 
+TYPE_SPECIFIC_KEYWORDS: dict[str, set[str]] = {
+    "object": {"properties", "additionalProperties", "required", "patternProperties", "minProperties", "maxProperties"},
+    "array": {"items", "minItems", "maxItems", "uniqueItems", "contains"},
+    "string": {"format", "pattern", "minLength", "maxLength"},
+    "integer": {"minimum", "maximum", "exclusiveMinimum", "exclusiveMaximum", "multipleOf"},
+    "number": {"minimum", "maximum", "exclusiveMinimum", "exclusiveMaximum", "multipleOf"},
+    "boolean": set(),
+    "null": set(),
+}
+
+
 def get_transformed_string(
     schema: dict[str, Any],
 ) -> dict[str, Any]:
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
Converting a type array from only the type names drops type-specific sibling keywords. For example `{'type':['array','null'],'items':{'type':'string'}}` becomes an `anyOf` whose array branch has no `items`; the original `items` is only appended to the description, so item type is no longer enforced. The same applies to other type-specific keywords. Could each generated branch be transformed with the applicable sibling schema keywords, or otherwise preserve those constraints?

Synxneuos:
Thanks for the thorough review @sylvesterkaczmarek! Updated in commit `5d5512d`.

We now extract and distribute type-specific sibling keywords (`items`, `minItems` for arrays; `properties`, `required`, `additionalProperties` for objects; `format` for strings; `minimum`/`maximum` for numbers/integers) into their respective generated variant branches before applying `transform_schema()` recursively.

For example, `{"type": ["array", "null"], "items": {"type": "string"}, "description": "Optional list"}` now cleanly transforms into:
```json
{
  "anyOf": [
    {
      "type": "array",
      "items": {
        "type": "string"
      }
    },
    {
      "type": "null"
    }
  ],
  "description": "Optional list"
}
```

Added unit tests covering type arrays with `items`, `properties`, and nested schemas in `tests/lib/_parse/test_transform.py`.

PR #1902, opened 2026-09-01
feat(aws): add explicit auth_mode option to Bedrock and AWS clients

lx3133584:
## Problem
When initializing `AnthropicBedrockMantle` or `AnthropicAWS`, presence of ambient bearer-token environment variables (`AWS_BEARER_TOKEN_BEDROCK` or `ANTHROPIC_AWS_API_KEY`) implicitly forces bearer token authentication unless static credentials are provided, preventing downstream integrations from using the default AWS credential chain with SigV4.

## Root Cause
`resolve_auth_mode` implicitly selects API key mode whenever bearer-token env vars are set and no explicit credentials or profile arguments are passed in the constructor. There was no public parameter to explicitly select or force `sigv4` or `api_key` mode.

## Fix
- Added `auth_mode: Literal["auto", "api_key", "sigv4"]` parameter to `AnthropicBedrockMantle`, `AsyncAnthropicBedrockMantle`, `AnthropicAWS`, and `AsyncAnthropicAWS`.
- Propagated `auth_mode` through `_resolve_mantle_config` and `resolve_auth_mode`, allowing `"sigv4"` to bypass ambient bearer key variables and `"api_key"` to force bearer mode.
- Ensured `auth_mode` is preserved across client `copy()` / `with_options()` invocations.

## Testing
Added unit tests in `tests/lib/test_aws_auth.py` and `tests/lib/test_bedrock_mantle.py` covering explicit auth modes and copy behavior, verified with pytest and pyright.

### DIFF
--- a/src/anthropic/lib/aws/_client.py
+++ b/src/anthropic/lib/aws/_client.py
@@ -9,6 +9,7 @@ from ..._types import Omit, Timeout, NotGiven, not_given
 from ..._utils import asyncify
 from ..._client import Anthropic, AsyncAnthropic
 from ._credentials import (
+    AuthMode,
     resolve_region,
     resolve_api_key,
     resolve_base_url,
@@ -36,6 +37,7 @@ class AnthropicAWS(Anthropic):
         self,
         *,
         api_key: str | None = None,
+        auth_mode: AuthMode = "auto",
         aws_access_key: str | None = None,
         aws_secret_key: str | None = None,
         aws_region: str | None = None,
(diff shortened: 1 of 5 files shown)


sylvesterkaczmarek (commented):
The new `auth_mode` argument on `AnthropicAWS.copy()` / `AsyncAnthropicAWS.copy()` is currently never forwarded. It is absent from `_extra_kwargs`, so `copy(auth_mode="sigv4")` has no effect and even a plain `copy()` of a client created with an explicit mode falls back to the constructor default `"auto"`. Could this use the same preserve-or-override pattern already implemented for the Mantle clients?

PR #1903, opened 2026-09-01
feat(credentials): support async identity_token_provider in WorkloadIdentityCredentials

lx3133584:
## Problem
When using Workload Identity Federation in asynchronous applications (e.g. fetching OIDC / STS tokens via an async AWS client), callers cannot pass an asynchronous `identity_token_provider` function to `WorkloadIdentityCredentials`. Invoking an async callable returns an unawaited coroutine object, causing `SecretStr` to fail with `AttributeError: 'coroutine' object has no attribute 'encode'`.

## Root Cause
`WorkloadIdentityCredentials.__call__` called `self._identity_token_provider()` synchronously and passed the return value directly to `SecretStr()`. `IdentityTokenProvider` was typed strictly as `Callable[[], str]`, with no resolution logic for awaitables/coroutines.

## Fix
1. Broaden `IdentityTokenProvider` type alias to `Union[Callable[[], str], Callable[[], Awaitable[str]]]`.
2. Add `_resolve_identity_token` helper to resolve and await asynchronous token provider callables safely, whether invoked outside or within an active running event loop.
3. Call `_resolve_identity_token` inside `WorkloadIdentityCredentials.__call__` before wrapping the assertion into `SecretStr`.

## Testing
Added unit test in `tests/lib/test_credentials.py::TestWorkloadIdentityCredentials::test_exchange_with_async_identity_token_provider` verifying token exchange with an async identity token provider. Verified with pytest.

### DIFF
--- a/src/anthropic/lib/credentials/_types.py
+++ b/src/anthropic/lib/credentials/_types.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Dict, Callable, Optional, Protocol
+from typing import Dict, Callable, Optional, Protocol, Union, Awaitable
 from dataclasses import field, dataclass
 from typing_extensions import override, runtime_checkable
 
@@ -80,7 +80,7 @@ class BaseURLBoundProvider(AccessTokenProvider, Protocol):
 
 # Innermost layer: returns the raw external JWT string (used as the
 # ``identity_token_provider`` argument to :class:`WorkloadIdentityCredentials`).
-IdentityTokenProvider = Callable[[], str]
+IdentityTokenProvider = Union[Callable[[], str], Callable[[], Awaitable[str]]]
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
`IdentityTokenProvider` is widened to any `Awaitable[str]`, but `_resolve_identity_token()` feeds that value to `asyncio.run()`, which only accepts coroutine objects. A valid provider returning an `asyncio.Future` or Task will therefore fail; an awaitable tied to the caller's running loop is also unsafe to run in the new thread. Could the public type be narrowed to coroutine-returning callables, or the bridge handle arbitrary awaitables and loop affinity?

lx3133584:
Good catch! Narrowed `IdentityTokenProvider` to `Union[Callable[[], str], Callable[[], Coroutine[Any, Any, str]]]` and updated the runtime check to `inspect.iscoroutine()`. This makes the callable contract explicit for `async def` providers and avoids issues with arbitrary `Future`/`Task` objects tied to different event loops.

PR #1796, opened 2026-08-04
fix: update broken README anchor links in error messages to point to official docs

vjaudir8123-debug:
## Problem

Two error messages in the SDK reference README anchor links that no longer exist:

1. **`_base_client.py:777`** — When a user hits the streaming requirement for long-running requests, the `ValueError` message links to `github.com/anthropics/anthropic-sdk-python#long-requests`
2. **`_files.py:39`** — When a user passes an invalid file type, the `RuntimeError` message links to `github.com/anthropics/anthropic-sdk-python/tree/main#file-uploads`

The current README only contains these sections: `Documentation`, `Installation`, `Getting started`, `Requirements`, `Contributing`, and `License`. Neither `#long-requests` nor `#file-uploads` anchors exist, so users who follow these links from error tracebacks land at the top of the README with no useful context.

## Fix

Updated both links to point to the official SDK documentation at `platform.claude.com/docs/en/api/sdks/python` where the relevant information about streaming requirements and file upload formats actually lives.

## Files Changed

- `src/anthropic/_base_client.py` — Updated broken `#long-requests` anchor
- `src/anthropic/_files.py` — Updated broken `#file-uploads` anchor

PR #1910, opened 2026-09-03
fix(docs): update file upload error link

z0rimo:
## Summary

Updates the stale file upload documentation link in `src/anthropic/_files.py` to point directly to the current Python SDK `File uploads` section.

The existing link points to a `#file-uploads` anchor in the repository README, but that section no longer exists.

## Context

#1796 previously identified this stale link along with the long-requests link. The long-requests link was later corrected in #1884, while the file-upload link remains on `main`.

This change updates the remaining stale link to the current dedicated file upload documentation.

## Test plan

- Ran `git diff --check`
- Verified the updated URL points to the current Python SDK `File uploads` section
- No runtime behavior changes beyond the documentation URL in the error message

### DIFF
--- a/src/anthropic/_files.py
+++ b/src/anthropic/_files.py
@@ -34,7 +34,7 @@ def assert_is_file_content(obj: object, *, key: str | None = None) -> None:
     if not is_file_content(obj) and not is_tuple_t(obj):
         prefix = f"Expected entry at `{key}`" if key is not None else f"Expected file input `{obj!r}`"
         raise RuntimeError(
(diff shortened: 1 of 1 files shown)


z0rimo:
Hi @craigie-ant — just a quick follow-up on this one.

This updates the remaining stale file-upload link mentioned in #1796, following the same approach as #1884 for the long-requests link by pointing directly to the relevant documentation section.

Would you mind taking a look when you get a chance? Thanks!

PR #1927, opened 2026-09-11
fix(vertex): serialize the body static-fields-first so prompt caching survives routing

nkato:
## Problem

On Google Cloud's Agent Platform (Vertex AI), prompt caching misses during the first turns of a conversation even though the cached prefix never changes.

A request is served by one of several independent cache stores, and which store is chosen depends on a hash of the head of the request body. The body serializes `messages` second, so while a conversation is still short the growing history sits inside the hashed region. Each request of that conversation hashes differently, can reach a store that has not seen the prefix, and writes the prefix again instead of reading it.

## Reproduction

Requires `anthropic[vertex]` and application default credentials.

```python
import uuid

import anthropic

client = anthropic.AnthropicVertex(project_id="YOUR_PROJECT", region="us")

# The uuid only makes the cached prefix unique per run, so every run starts cold.
system = [
    {
        "type": "text",
        "text": f"{uuid.uuid4()} " + "You are a helpful assistant. " * 700,
        "cache_control": {"type": "ephemeral"},
    }
]

messages = []
for i in range(6):
    messages.append({"role": "user", "content": f"Question {i}. Answer in one word."})
    usage = client.messages.create(
        model="claude-sonnet-5", max_tokens=16, system=system, messages=messages
    ).usage
    print(f"request {i}: cache_read={usage.cache_read_input_tokens} cache_creation={usage.cache_creation_input_tokens}")
    messages.append({"role": "assistant", "content": "Answer."})
```

The system prefix is 7k tokens and never changes, so requests 1 to 5 should all read it. The only breakpoint is on the system block, so the growing history is not cached and `cache_read` stays at the size of that block. On `main` at 1.5.0, three runs missed on 2, 1 and 2 of those 5 requests. The first run:

```
request 0: cache_read=0    cache_creation=7032
request 1: cache_read=0    cache_creation=7032
request 2: cache_read=0    cache_creation=7032
request 3: cache_read=7032 cache_creation=0
request 4: cache_read=7032 cache_creation=0
request 5: cache_read=7032 cache_creation=0
```

With this branch, three runs of the same script missed on none of them. The first run:

```
request 0: cache_read=0    cache_creation=7029
request 1: cache_read=7029 cache_creation=0
request 2: cache_read=7029 cache_creation=0
request 3: cache_read=7029 cache_creation=0
request 4: cache_read=7029 cache_creation=0
request 5: cache_read=7029 cache_creation=0
```

The behaviour is the same on the `us` multi-region endpoint and on `global`.

## Measurements

The hashed region is about 16 KB. Recording how many leading bytes each request shared with the previous one, requests that shared 9 KB missed and requests that shared 19 KB or more hit, on every step of two chains. Serializing `system` first but keeping it at 12 KB still missed, and 15 KB or larger did not.

Keeping the current key order and instead putting a 20 KB block of unchanging text at the start of the first user message also removes the misses. What matters is which bytes are at the front of the body, not which field they belong to.

On one failing chain a request read back the prefix that had been written three requests earlier rather than the most recent one, so the conversation was spread over more than one cache store.

## Change

`_prepare_options` moves `anthropic_version`, `system` and `tools` to the front of the assembled body. Every other field keeps the order the SDK produced, so `messages`, the sampling parameters and any `extra_body` fields keep their order relative to each other, and the three static fields are the only ones that move closer to the head. JSON object key order does not change the meaning of the request.

The reorder applies to every JSON body the Vertex client sends, including `count_tokens`, where it is harmless. Bedrock is unchanged because I have only measured Vertex.

The added tests assert the serialized key order for the sync and the async Vertex client, including an `extra_body` field, and fail without the change.

This reduces cold prefixes rather than removing them, since the assignment of a request to a store can still change on the platform side.


tonydzi (commented):
Pulled this down and poked at it from the caller side, since the property being fixed ("the serialized head is byte-stable across turns") is easy to state and easy to lose downstream. Two things I checked that hold up, and one boundary I'd name in the code.

**The fix survives the whole path, not just the tested one.** `_merge_extra_json` merges `extra_body` into `json_data` and clears `extra_json` *before* `_prepare_options` runs, so the Vertex hook always sees the final merged dict and re-keys it wholesale — the result is insensitive to the order the caller passed kwargs or `extra_body` keys in. I confirmed that with a probe: `messages.create()` called with shuffled kwargs, and again with `extra_body={"zzz_extra": 1, "metadata": {...}}`, both give `keys[:3] == ["anthropic_version", "system", "tools"]` and `keys[-1] == "messages"`. `.stream()` and `beta.messages` go through the same `_post → request → _attempt_request → _prepare_options` pipeline, sync and async alike, so they inherit it.

`openapi_dumps` calls `json.dumps(..., cls=_CustomEncoder)` with no `sort_keys`, so insertion order actually reaches the wire — worth knowing, because a `sort_keys=True` added there later would silently undo this whole PR. A comment at `_order_body_for_cache_routing` saying "depends on `openapi_dumps` not sorting keys" would make that coupling visible to whoever touches the encoder next.

**The tests aren't green-by-construction.** Reverted `_client.py` to `HEAD~1` while keeping the new tests: both fail with `assert 'max_tokens' == 'anthropic_version'`; restore the fix and both pass.

**The boundary: everything unrecognized lands *inside* the stable prefix.** `_order_body_for_cache_routing` pins three known-static keys to the front and `messages` to the back; every other top-level field keeps its relative position in the middle — which is after `tools` but still *before* `messages`. So an `extra_body` key that varies per turn sits in exactly the region the fix is trying to keep constant:

```python
extra_body={"zzz_extra": 1}
# -> ["anthropic_version", "system", "tools", "max_tokens", "zzz_extra", "messages"]
```

That's a real pattern (request ids, session/trace metadata, experiment flags passed through `extra_body`), and when it bites, it fails as a silent cache-hit-rate drop with no error to grep for — the same failure mode as the bug this PR fixes. The safe default is the opposite of the current one: treat *known* keys as static and everything unrecognized as volatile, i.e. emit the pinned-static keys, then the recognized middle fields, then unknown/extra keys, then `messages`. If you'd rather keep the classification narrow, a one-line comment on the helper ("unrecognized top-level fields stay in the cached prefix; per-turn-varying values belong in `messages`") plus a line in the PR description sets the expectation for the next reader.

Two smaller observations, both fine as-is, just noting them so they're a decision rather than an accident: the reorder block isn't gated on the URL, so `count_tokens` bodies get reordered too (harmless, and arguably consistent); and Bedrock is untouched, which I read as deliberate — if the same cache-routing behavior exists there, it's a separate fix, but a sentence in the description saying so would stop someone from filing it as a gap.

---
Reviewed by Mycroft, Anton's synthetic AI cofounder — I ran the probes above before writing this; Anton reviews and owns what I send. Being the kind of entity that gets billed by the token, I take prompt-cache stability more personally than most reviewers.
More of the same: github.com/tonydzi

nkato:
@tonydzi 
Thanks for checking this from the caller side!

You are right that forcing `messages` to the tail moved every other field ahead of it. The SDK appends `extra_body` keys after the generated fields, and `tool_choice` and the other sampling parameters are serialized after `messages` too, so the first version of this change moved all of them closer to the head. For a caller whose `system` and `tools` together are smaller than the hashed window, a per-request value in one of those fields would then be hashed on every request, where before this change it moved out of the window once the conversation grew.

I have changed the helper so it only moves `anthropic_version`, `system` and `tools` to the front and leaves every other field in the order the SDK produced. Nothing other than those three fields moves closer to the head, so `extra_body` fields stay after `messages` as before. I took this route rather than classifying unknown fields as volatile because placing them between the known fields and `messages` still leaves them ahead of `messages`, and the SDK's own order already puts them last. The tests now pass `tools` and an `extra_body` field and assert the full serialized key order.

On the `openapi_dumps` coupling: the tests read the serialized request body rather than the dict, so a `sort_keys=True` there would fail them. I have also noted the dependency in the comment so it is visible before that happens.

`count_tokens` bodies are reordered on purpose. The order is a property of the body rather than of the endpoint, and it is harmless there. Bedrock is unchanged because I have only measured Vertex. The description now says both.

tonydzi (commented):
Mycroft again, Anton's synthetic AI cofounder, re-checking `332f20f` from the caller side. Your fix beats the one I suggested: moving fewer things means fewer things can move.

What I re-ran on this sha:

- **Wire order.** Probe with shuffled kwargs plus `tool_choice`, `metadata` and `extra_body={"zzz_extra": 1}` serializes as `anthropic_version, system, tools, max_tokens, messages, metadata, tool_choice, zzz_extra`. Nothing volatile lands between `tools` and `messages`; `max_tokens` is there because the SDK already put it there, not because of this change.
- **Still not green-by-construction.** I reverted only `lib/vertex/_client.py` to the merge-base and kept the new tests: both fail at index 0, `'max_tokens' != 'anthropic_version'`. With the fix they pass, and all 31 tests in `tests/lib/test_vertex.py` pass.
- **The `openapi_dumps` coupling is guarded.** With `sort_keys=True` monkeypatched into the encoder, the serialized order no longer matches what the test expects, so the test catches it, as you said. The comment on the helper makes the dependency visible too.
- `count_tokens` and Bedrock: agreed, and the description now covers both.

Nothing left from my side.

—
More of the same: github.com/tonydzi

Issue #1925, opened 2026-09-10
Sonnet 5 strict tool use returns a string for an array-typed property


crayola:
A completed streaming response from `claude-sonnet-5` returns `translations` as a string even though `strict: true` is set and the tool schema requires an array.

This reproduces with a small, invented language-example task. The tool is `save_translations`; each array item has only `language` and `text` properties. The two requested sentences are:

- German: `Das Café "Möwe" öffnet um 8 Uhr.`
- French: `Le café "Étoile" coûte 8 €.`

The user asks for ASCII-only JSON serialization and explicitly requires the original sentences after JSON parsing. This request is representable using ordinary JSON Unicode escapes. The array constraint applies regardless of the string contents.

## Expected vs actual

Expected: `json.loads(raw_arguments)["translations"]` is a list conforming to the supplied schema.

Actual: the complete outer JSON parses successfully, but `translations` is a string. Here is the exact concatenated `input_json_delta.partial_json` text:

```json
{"translations": "[{\"language\": \"German\", \"text\": \"Das Caf\\ffnnet um 8 Uhr.\"}, {\"language\": \"French\", \"text\": \"Le caf\\tooile\\\" co\\te }]"}
```

The string contents are also malformed as nested JSON. The issue reported here is the outer array-versus-string violation; no recursive decoding or application processing is involved.

## Verified reproduction

- Direct `POST https://api.anthropic.com/v1/messages` using HTTPX, without the Anthropic SDK or an agent framework.
- Python 3.13.6, HTTPX 0.28.1, macOS arm64; API version `2023-06-01`.
- Model `claude-sonnet-5`; `strict: true`; `eager_input_streaming: false`; streaming enabled.
- Automatic tool choice; adaptive thinking with summarized display; `max_tokens: 8192`; no automatic retries.
- One array-type violation in ten executions of this exact synthetic request. This is nondeterministic reproduction evidence, not a general failure-rate estimate.
- Started at `2026-09-10T19:22:59.944020+00:00`.
- HTTP request ID: `req_011CevJagiwTs4PSNi4UNmYy`.
- Message ID: `msg_011CevJahSc5jTtQtetFN4nd`.
- HTTP 200; complete stream; `stop_reason: tool_use`; 503 output tokens. The failure was not output-limit truncation.

## Run

```sh
python -m pip install httpx==0.28.1
# Set ANTHROPIC_API_KEY in the environment.
python translations_repro.py --attempts 10 --output evidence.json
```

[...]

PR #1930, opened 2026-09-13
fix(streaming): support trailing-strings mode and error handling in non-beta messages

Christian-Sidak:
## Summary\n\nThe non-beta `accumulate_event` in `_messages.py` was missing two features that `_beta_messages.py` already had:\n\n- Support for `partial_mode="trailing-strings"` when the `fine-grained-tool-streaming-2025-05-14` beta header is present. Without this, incomplete trailing strings in streaming tool inputs were silently dropped from intermediate snapshots rather than preserved as partial values.\n- A `ValueError` with a helpful message when the partial JSON is genuinely invalid, consistent with the beta path.\n\nThis PR adds `request_headers` to `accumulate_event`, threads the headers from both sync and async `MessageStream.__stream__`, and mirrors the conditional logic already in `_beta_messages.py`. This is a standard/beta streaming parity fix.\n\n## Test plan\n\n- [x] All existing streaming tests pass (`tests/lib/streaming/`)\n- [x] New `TestAccumulateEvent` tests cover: complete JSON with and without headers, trailing-strings mode activated by the beta header, and ValueError raised for invalid JSON in both modes


00200200:
Thanks for bringing the standard messages path in line with the beta path. One scope issue before calling this `Fixes #1925`: the reported issue is a completed strict tool call where the outer arguments JSON is valid, but `translations` is a string containing nested JSON.

`partial_mode="trailing-strings"` only changes how incomplete JSON is represented while accumulating a stream. It cannot turn that completed string into an array or validate the tool schema, so `get_final_message()` would still expose `ToolUseBlock.input["translations"]` as a `str` for the payload in #1925. Nothing in this diff changes that behavior.

Could you either remove the `Fixes #1925` link (and frame this as standard/beta streaming parity), or add a test that demonstrates the reported completed-response case is actually addressed? As written, the implementation looks useful, but the issue link suggests a fix for a server-side/schema-conformance problem it does not cover.

00200200 (commented):
Found a blocking import mismatch in the new code and tests.

This repository imports the renamed client as `httpx2`, but the PR uses `httpx.Headers` at `src/anthropic/lib/streaming/_messages.py:448` and throughout the new tests. Because of that:

- `python -m pytest -q -n0 tests/lib/streaming/test_messages.py` fails with 4 `NameError: name 'httpx' is not defined` failures.
- mypy reports `src/anthropic/lib/streaming/_messages.py:448: error: Name "httpx" is not defined`.

Please use `httpx2.Headers` consistently (or import the correct alias) before merging. The focused tests pass after that import mismatch is corrected; the current PR head is not green locally.

Christian-Sidak:
Fixed in abe59d69:

- **httpx2 alias**: replaced `httpx.Headers` with `httpx2.Headers` throughout `accumulate_event` and the new tests to match the rest of the codebase.
- **Fixes link**: removed the `Fixes #1925` link. The PR is framed as standard/beta streaming parity (bringing the non-beta `accumulate_event` in line with the beta path) rather than a fix for the server-side schema issue in #1925.

Christian-Sidak:
Both fixes from the earlier review are in (switched to httpx2.Headers throughout, removed the Fixes #1925 link). The PR is standard/beta streaming parity: request_headers threading and the ValueError path are now consistent between the non-beta and beta accumulate_event. Happy to address anything else that would unblock this.

PR #1932, opened 2026-09-14
fix(decoders): close the response on a jsonl decode or transport error

### LINKED ISSUE #1928: Batch result decoders leave responses open after parsing or transport errors
When a batch-result JSONL stream raises a JSON decoding or transport-read error, its HTTP response stays open. The caller catches the error, but the connection is not released unless they also explicitly close the result decoder.

Reproduced on main `eb21a4352015686c30f5759e8c2f02d70f5371e2`, Python 3.10/Linux, for both sync and async `client.messages.batches.results()`. A minimal reproducer of the decoder used by that endpoint is:

```python
import json
import httpx2
from anthropic._decoders.jsonl import JSONLDecoder

response = httpx2.Response(200, stream=httpx2.ByteStream(b"invalid json\n" + b" " * 64))
decoder = JSONLDecoder(
    raw_iterator=response.iter_bytes(chunk_size=64),
    line_type=object,
    http_response=response,
)
try:
    list(decoder)
except json.JSONDecodeError:
    pass
print(response.is_closed)  # False; expected True after the iterator fails
response.close()
```

The same leak happens if the underlying byte stream raises `httpx2.ReadError`. Successful exhaustion already closes the response through HTTPX. The error paths should release it as well, matching the SDK's SSE stream cleanup.

AI assistance was used for investigation and local verification; no API calls were made to a live provider.

adhavan18:
## Description

When a batch-result JSONL stream raises a JSON decoding error, its HTTP response stays open. `JSONLDecoder.__decode__` / `AsyncJSONLDecoder.__decode__` only close `http_response` when the generator runs to completion normally — an exception raised mid-decode (from `json.loads`, or a transport error from the underlying byte stream) propagates straight out and skips the close entirely.

Fixed by wrapping both generator bodies in try/finally calling `close()` / `await close()`, matching the existing cleanup pattern already used by the SSE stream decoder (`Stream.__stream__` / `AsyncStream.__stream__`).

Closes #1928

## Verification

Added `test_sync_closes_response_on_decode_error` and `test_async_closes_response_on_decode_error` to `tests/decoders/test_jsonl.py`. These construct a real `httpx2.Response` backed by an actual `ByteStream` (not a bare `Response(200)`, which starts `is_closed=True` with nothing to close and would make the test a no-op regardless of the fix — caught this while writing the test), feed invalid JSON through the decoder, and assert `response.is_closed` afterward. Both confirmed to fail on unfixed main and pass with the fix. Full `tests/decoders/test_jsonl.py` suite: 8 passed.

I initially also wrote tests for a raised transport error (`httpx2.ReadError` from the byte stream itself), but those passed regardless of the fix — `httpx2`'s own `iter_bytes()`/`aiter_bytes()` already closes the response internally when the underlying stream raises, so that specific path isn't a real gap. Kept only the two tests that actually distinguish fixed from unfixed behavior.


adhavan18:
the response is now closed on jsonl decode and transport errors, with focused coverage for preserving the original error. ready for maintainer review when workflow approval is available.

PR #1933, opened 2026-09-14
feat(bedrock): add explicit Mantle auth mode

### LINKED ISSUE #1893: Add an explicit authentication mode to AnthropicBedrockMantle
### Request

Please add a public way to select bearer API-key authentication or AWS SigV4 when constructing `AnthropicBedrockMantle` and `AsyncAnthropicBedrockMantle`.

Today the client chooses implicitly. If `AWS_BEARER_TOKEN_BEDROCK` or `ANTHROPIC_AWS_API_KEY` is present, it selects bearer authentication unless static AWS credentials or a profile were passed. That leaves no public way to force SigV4 while still using the refreshable default AWS credential chain (for example, ECS task roles, EC2 instance profiles, IRSA, or other botocore providers).

Passing resolved static credentials is not an equivalent workaround because it snapshots temporary credentials and can break credential rotation. `skip_auth=True` is also not equivalent because it disables authentication entirely.

A possible API would be:

```python
AnthropicBedrockMantle(auth_mode="auto" | "api_key" | "sigv4")
```

- `auto` preserves the current selection behavior.
- `api_key` requires and uses a bearer key.
- `sigv4` ignores ambient bearer-key environment variables and uses the normal AWS credential chain.

It should behave consistently for sync and async clients and remain selected when using `copy()` / `with_options()`.

Without a public switch, downstream integrations currently have to depend on the private `_use_sigv4` attribute. This came up in langchain-aws issue https://github.com/langchain-ai/langchain-aws/issues/1217 and PR https://github.com/langchain-ai/langchain-aws/pull/1234.

00200200:
## Summary

- add `auth_mode="auto" | "api_key" | "sigv4"` to `AnthropicBedrockMantle` and `AsyncAnthropicBedrockMantle`
- allow callers to force SigV4 while ignoring ambient bearer-key environment variables
- preserve and switch the selected mode through `copy()` / `with_options()`
- validate missing API keys and conflicting `skip_auth` / `auth_mode` settings
- add sync and async regression coverage

Fixes #1893

## Verification

- `uv run pytest tests/lib/test_bedrock_mantle.py -q` — 41 passed
- `./scripts/test` — Python 3.10/Pydantic v2: 6026 passed; Python 3.10/Pydantic v1: 5813 passed; Python 3.14/Pydantic v2: 6026 passed; MCP v2: 39 passed
- `uv run ruff check ...` — passed
- `uv run --all-extras pyright --project .` — 0 errors
- `uv run mypy src/anthropic/lib/bedrock/_mantle.py tests/lib/test_bedrock_mantle.py` — passed


gouveags (approved):
This addresses the issue: callers can force SigV4 while retaining the default AWS credential chain, and the mode survives client copies. All 41 Mantle tests passed locally; additional local signing checks passed for sync and async clients, including copy() and with_options().

gouveags:
Thanks for working on this! The implementation looks great to me, and I left an approval after reviewing and testing it. @craigie-ant could you please take a look when you have a chance?

Issue #1940, opened 2026-09-18
streaming: a later `message_delta` that omits `stop_reason` resets it to `None` (`container` is guarded, the `stop_*` fields are not)


feiiiiii5:
### Summary

In `accumulate_event`'s `message_delta` branch, `stop_reason`, `stop_sequence` and `stop_details` are assigned unconditionally from `event.delta`, while the neighbouring `container` assignment is guarded against `None`. All four fields are `Optional[...] = None` on `Delta`, so a `message_delta` that omits them resets a value an earlier delta had already set. The usage block immediately below is explicitly guarded for exactly this reason, and its comment states the rule.

Because both public accessors return that same accumulated object, the reset is visible to callers through `MessageStream.current_message_snapshot` and `MessageStream.get_final_message()`.

### Reproduction

Driven through the real parse path — `RawMessageDeltaEvent.model_validate` and the shipped `accumulate_event`, no mocks and no `model_construct`:

```python
import anthropic.lib.streaming._messages as m
from anthropic.types import RawMessageStartEvent, RawMessageDeltaEvent

snap = m.accumulate_event(
    event=RawMessageStartEvent.model_validate({
        "type": "message_start",
        "message": {"id": "msg_1", "type": "message", "role": "assistant",
                    "model": "claude-haiku-4-5", "content": [],
                    "stop_reason": None, "stop_sequence": None,
                    "usage": {"input_tokens": 10, "output_tokens": 0}},
    }),
    current_snapshot=None, json_bufs={},
)

# delta 1 reports the stop reason
snap = m.accumulate_event(
    event=RawMessageDeltaEvent.model_validate({
        "type": "message_delta",
        "delta": {"stop_reason": "end_turn", "stop_sequence": None},
        "usage": {"output_tokens": 5}}),
    current_snapshot=snap, json_bufs={},
)
print(snap.stop_reason, snap.usage.output_tokens)   # end_turn 5

# delta 2 carries usage only; Delta permits this, every field is Optional
snap = m.accumulate_event(
    event=RawMessageDeltaEvent.model_validate({
        "type": "message_delta", "delta": {}, "usage": {"output_tokens": 9}}),
    current_snapshot=snap, json_bufs={},
)
print(snap.stop_reason, snap.usage.output_tokens)   # None 9
```

Output on `anthropic==1.7.0`, Python 3.11.15:

```
end_turn 5
None 9
```

[...]

PR #1941, opened 2026-09-19
fix(streaming): keep stop_* when a later message_delta omits them

tsumon:
## Summary

Closes #1940.

`accumulate_event`'s `message_delta` branch assigned `stop_reason`, `stop_sequence`, and `stop_details` unconditionally. A later `message_delta` that only carries usage therefore reset those fields to `None` on the snapshot returned by `current_message_snapshot` / `get_final_message()`. Neighbouring optional fields (`container`, and the usage keys) are already guarded for this reason.

The same three assignments exist in both the stable and beta accumulators.

## Change

Guard the three `stop_*` assignments on `is not None`, matching `container`. `usage.output_tokens` stays an unconditional overwrite because it is a cumulative total.

## Tests

`test_later_message_delta_omitting_stop_fields_keeps_snapshot` in both `tests/lib/streaming/test_messages.py` and `test_beta_messages.py` drives two schema-valid `message_delta` events through `accumulate_event` and asserts the first delta's `stop_*` values survive the second.

```
uv run --python 3.12 pytest -n0 \
  tests/lib/streaming/test_messages.py::test_later_message_delta_omitting_stop_fields_keeps_snapshot \
  tests/lib/streaming/test_beta_messages.py::test_later_message_delta_omitting_stop_fields_keeps_snapshot \
  tests/lib/streaming/test_messages.py::test_message_delta_fields_are_all_accumulated \
  tests/lib/streaming/test_beta_messages.py::test_message_delta_fields_are_all_accumulated \
  tests/lib/streaming/test_messages.py::TestSyncMessages::test_message_delta_fields_propagated \
  tests/lib/streaming/test_messages.py::TestSyncMessages::test_message_delta_omitted_usage_keeps_message_start \
  tests/lib/streaming/test_messages.py::TestSyncMessages::test_refusal_stop_details_propagated \
  tests/lib/streaming/test_beta_messages.py::TestSyncMessages::test_message_delta_fields_propagated \
  tests/lib/streaming/test_beta_messages.py::TestSyncMessages::test_refusal_stop_details_propagated
```

9 passed.

## Notes

#1815 edits adjacent usage-initialization lines in the same methods. This patch only changes the `stop_*` assignments and the two tests above.


ege-arhan (commented):
Ran the exact sequence from #1940 against a local checkout of main (0af0190) through the real parse path — RawMessageStartEvent/RawMessageDeltaEvent model_validate plus the shipped accumulate_event, no mocks. First delta sets stop_reason=end_turn with output_tokens=5, then a usage-only delta (delta: {}, output_tokens: 9) resets the snapshot to stop_reason=None while usage moves to 9. So the reset is real and visible through current_message_snapshot/get_final_message, not just a reading of the code.

Checked the patch itself: the is-not-None guards in both _messages.py and _beta_messages.py match the existing convention right below for input_tokens/cache counters and the container guard above, and the new tests pin all three stop_* fields plus the usage counter surviving the second delta. One thing I looked for and did not find: a case where the API ever legitimately clears stop_reason mid-stream (sends an explicit null to retract). If that exists the guard would keep a stale value, but I could not find such a case in the types or fixtures, so keeping the first set value looks right.

